PR #606, opened 2013-03-05
Update block_frame_reflower.cls.php

grzebzi:
Fixes problem when not showing block elements (like DIVs) when they are styled with 
position: absolute;
and
height: in pixels

### DIFF
--- a/include/block_frame_reflower.cls.php
+++ b/include/block_frame_reflower.cls.php
@@ -780,6 +780,8 @@ class Block_Frame_Reflower extends Frame_Reflower {
       
       $style->left = $orig_style->left;
       $style->right = $orig_style->right;
+      $style->top = $orig_style->top; // added by GB to set proper height of block element
+      $style->bottom = $orig_style->bottom; // added by GB to set proper height of block element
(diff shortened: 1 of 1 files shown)


bsweeney:
Also, we'd like to avoid merging changes directly into the master branch where possible. Target your changes to the develop branch and we'll take a closer look.

bsweeney:
@grzebzi without a minimal test case I'm not sure what problem this fixes. I've tried a few test cases and so far the only way I can get an absolutely positioned DIV to not show up is if it has no content (i.e. no width).

I'll push this to the next release unless you can provide more info.

bsweeney:
FYI here's what I tested with:

```
<html>
  <body>
    <div style="position: absolute; height: 50px; left: 10px; top: 10px; background-color: black;"></div>
    <div style="position: absolute; height: 50px; left: 10px; top: 70px; background-color: black; width: 10px;"></div>
    <div style="position: absolute; height: 50px; left: 10px; top: 160px; background-color: black;">a</div>
  </body>
</html>
```

PR #617, opened 2013-04-09
Added font name save name

TeckniX:
When new fonts are added, they are saved as md5 names - To provide a nicer name for organization, the register_font function can be passed a friendly local_name that will be cleaned of spaces and invalid characters, for local save.

A matching change would be made in the www/controller.php, to have the uploaded file's name passed to the register function: (line 64)

`if (!Font_Metrics::register_font($style_arr, $data["tmp_name"][$name], $data["name"][$name])) {`


bsweeney:
We'll consider some changes related to this for the next release. Since the PR is out of date if you have time it'd be great if you had time to update it.

PR #661, opened 2013-06-11
Update cellmap.cls.php

nathanntg:
Made an update to cellmap to prevent exception when rowspan spans multiple pages. Rather than fully removing spaning frames (cells), it removes rows individually from the "rows" index. Only once the "rows" index is empty does it remove the frame (cell).

### DIFF
--- a/include/cellmap.cls.php
+++ b/include/cellmap.cls.php
@@ -612,7 +612,8 @@ class Cellmap {
       return;  // Presumably this row has alredy been removed
     }
 
-    $this->_row = $this->_num_rows--;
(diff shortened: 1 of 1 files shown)


bsweeney:
Interesting, I'd like to take a closer look but can you first port your changes to the develop branch. We'd like to avoid merging code directly into the master branch.

nathanntg:
Will do. I've been using this fix for a while, and it helps with some of the long tables, which were previously causing fatal errors. It's definitely not a perfect solution, but a good improvement. I will submit a new pull request against the develop branch.

PR #780, opened 2014-02-06
Try class_exists before manually loading Font class

phanzink:
Calling `class_exists` serves two purposes. 

- If the class is already loaded, it skips it 
- More importantly it triggers the autoloader to look before including it manually which fixes a problem using the default config with composer as the `DOMPDF_LIB_DIR . "/php-font-lib/classes/Font.php` path is incorrect when loading with composer.


bsweeney:
Considering where we are with the 0.7.0 code I think it's safe to close this issue.

PR #781, opened 2014-02-10
Autoloader

adrianmacneil:
This is my proposed solution to #780. Basically we remove the autoloader entirely from dompdf_config.inc.php, and let the developer make their own decision whether or not to use it.

Unfortunately I didn't realise you were so close to releasing 0.6 final, it would have been nice to make this change before the release. As you said in 0.7 you are planning to make everything PSR0 perhaps you can drop the autoloader entirely at that point (or just ship a default PSR0 autoloader).

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,17 @@ git submodule init
 git submodule update
 ```
 
+To use DOMPDF in your project, you must include the configuration file
+and class autoloader:
+
+```php
+// include DOMPDF's default configuration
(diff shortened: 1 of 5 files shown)


bsweeney:
We'll try to have PSR-0/PSR-4 support in 0.7.0. Supporting PSR-1/PSR-2 will come as we're able to work it in (unless somebody wants to work on it ... I'll create a branch). Considering where we are with 0.7.0 I'm going to go ahead and close this one.

PR #782, opened 2014-02-11
moving the defination of __DOMPDF_FONT_CACHE_FILE to the config file,

zaherg:
moving the defination of __DOMPDF_FONT_CACHE_FILE to the config file, and check if its not defined in the Font_Metrics class, so if its not defined then we use the default definition, this will help to put the fonts cache any where we want


kidalikevin:
Hi, I don't understand what this statement means, I using laravel.
moving the defination of __DOMPDF_FONT_CACHE_FILE to the config file, and check if its not defined in the Font_Metrics class, so if its not defined then we use the default definition, this will help to put the fonts cache any where we want

bsweeney:
@Sulan27 this does not appear to be related to this pull request. Please [start a new discussion](https://github.com/dompdf/dompdf/discussions/new/choose) to receive help.

PR #788, opened 2014-02-20
Improved performance, started implementing support for PSR-*

gabrielbull:
List of changes:
- Started reformatting code to be PSR-2 compliant
- Created a new PSR-4 compliant autoloader
- Started implementing namespaces
- Added unit tests with PHPUnit
- Fixed some performance issues
- Dropped support for PHP <= 5.2 (People can still download older releases for older PHP versions)
- Stopped using submodules as they can cause problems and are not appropriate for a PHP Library, industry best practice is to use Composer

Things to do:
- [x] Create a new release (proposed: 0.7.0)
- [x] Rename `dev` branch to `develop` to use the best practice and be compliant with the rest of the git community

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,20 @@
 language: php
+
 php:
-    - 5.3
-    - 5.4
-    - 5.5
+  - 5.3
+  - 5.4
+  - 5.5
+  - 5.6
+  - hhvm
 
 before_script:
-    ## Composer
-    - curl -s http://getcomposer.org/installer | php
-    - php composer.phar install
+  - composer self-update
+  - composer install --prefer-source --no-interaction --dev
+
+script: phpunit
+
+matrix:
+  allow_failures:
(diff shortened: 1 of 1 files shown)


barryvdh:
Is there an ETA for the new version?

gabrielbull:
Albeit not an ETA, there has been so much change on the current master that this update would need to be reworked from the ground up. I have been wanting to do it for a while without much time to do it. It appears my schedule might get cleared a bit soon and I might be able to get back into the project and make this update happen.

beporter:
Namespacing (and general PSR support), plus making dompdf fully compatible with being installed via composer and their autoloader, are critical show stoppers for my company. We are forced to exclude dompdf as an option in our projects due to these current shortcomings, so I look forward to progress in those areas! :+1: 

Is there a `help-wanted` section where we might be able to assist with work in those areas?

PR #798, opened 2014-03-12
Encoding validation

inluxc:
Validate the encoding before converting it.


bsweeney:
No follow up for a while and it looks like the commit I referenced does pretty much the same thing. Imma close. If any new and relevant information comes up I can re-open and re-assess.

PR #799, opened 2014-03-17
Only one instance of each Positioner type

olivier-weka:
Instead of giving each frame their own positioner object, they only get a reference to the applicable positioner type. This is handled by \Dompdf\Frame\Factory.

Over the sample test files (www/test/*.html), this results in a 5% speed improvement but surprisingly has no effect on memory usage.


bsweeney:
@olivier-weka have you tried the same with the renderers as well? These two class types are fairly light-weight which may be why you don't see much of a change in resource usage.

I wonder if there might be more affect on a document with a large number of frames.

bsweeney:
But preliminary testing says probably not. The real memory hog is obviously elsewhere. But every little bit counts.

PR #818, opened 2014-04-30
fix custom font family loading

xarem:
custom font loading from own cache file wasn't working because this function tries to read the wrong file (generated by load_font.php).

### DIFF
--- a/include/font_metrics.cls.php
+++ b/include/font_metrics.cls.php
@@ -259,7 +259,11 @@ class Font_Metrics {
     }
     
     // Merge provided fonts
-    self::$_font_lookup += $dist_fonts;
+    if (is_array(self::$_font_lookup)) {
+        self::$_font_lookup = array_merge(self::$_font_lookup, $dist_fonts);
+    } else {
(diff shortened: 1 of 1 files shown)


bsweeney:
It does look like something's wrong here, though I think your fix may not be quite what's needed. Still, I'll merge and tweak as necessary.

bsweeney:
Before I merge you'll first need to port your changes to the develop branch. We'd like to avoid merging code directly into the master branch.

bsweeney:
I see what you were trying to fix. The first block of code you modified was an attempt to migrate from an older version of the font family cache to a newer version (if present). This was done on behalf of the user to avoid any problems with upgrades from 0.5.x to 0.6.x.

The real issue is that the old cache file was not deleted after the migration. This would cause dompdf to re-migrate the font family cache every time it was run. I'm surprised we didn't see more complaints about this issue ... or maybe we did and just didn't realize this was the source of people's font problems.

At this point I think it's probably best to just remove that code block entirely.

The change you made in the latter block has since been addressed.

PR #822, opened 2014-05-12
Updating pdflib_adapter.csl.php stream() to match class.pdf.php 

bonfire-dev:
Updating the stream() function to match class.pdf.php to handle UTF-8 characters in filename.

We had already gone over these changes in https://github.com/dompdf/dompdf/pull/594 so the review aspect should be good to go.


bsweeney:
Looks good but please use the develop branch as the basis for your changes. We'd like to avoid merging code directly into the master branch.

PR #827, opened 2014-05-21
Strip away copyrighted ICC profile

DavidPrevot:
This images embed a non free ICC profile that doesn’t allow
modification.

Please use the proposed version stripping away this copyrighted binary
profile, probably embedded by an image software at one time:

	exiftool -icc_profile"-<=" www/test/images/cmyk_test2.jpg

I’ve checked the resulting image and trust it is identical to the
original.

Some more background on the issue:
https://lists.debian.org/debian-devel/2014/05/msg00339.html


bsweeney:
Thank you for your patch. Since the www content is no longer included with dompdf I'm going to close the issue. If the content is resurrected as part of the [utils project](/dompdf/utils) I'll be sure to include the updated image.

We'll try to keep your note in mind when adding images to the library in the future.

PR #842, opened 2014-07-03
dompdf_tempnam 

vavrecan:
Created custom handler for tempnam, for hosting with safemode on, for example google app engine. Uses custom memory php wrapper.


bsweeney:
Looks interesting. We'd like to avoid merging changes directly into the master branch where possible. Target your changes to the develop branch and we'll take a closer look.

bsweeney:
If you have time to fix up this PR I'd be grateful. Still seems like a potentially good fix and I'll look at it more for the next release.

Issue #830, opened 2014-06-05
Barcodes not displaying correctly


vicmar:
Hi! When using DOMPDF the barcodes and QR codes on my HTML document (generated with Zend Barcode library) display blurry, not sharp like they do on the HTML where they only have just black and white pixels. I've read all I could find on google. Changing the value of DOMPDF_DPI just affects the size of the images but they show blurry in any case. I have the same problem with div borders. A simple 1px black border of a div shows blurry. In this case is just an aesthetic problem but with the bar codes the barcode reader struggles to scan them so I need to solve it.
Any help would be really appreciated. Thanks a lot

keescornelisse:
Thanks @grebsle for the fix!

PR #866, opened 2014-10-07
#830 Fixing barcode issue

excellentcontributor:


bsweeney:
Closing this as it's a bit overkill to fix the specific issue. This should help identify what we can do to mitigate the issue.

PR #878, opened 2014-11-27
Compatibility with PDFLib 9

IceReaper:


bsweeney:
This was probably sufficiently covered by #1743.

PR #889, opened 2015-01-05
Use newer versions of jQuery

fisharebest:
This package includes jQuery 1.4.2, which contains a known XSS vulnerability - and my employer will not allow it to be installed.

This PR simply updates to the latest version of jQuery and also fetches it from the official jQuery CDN.


bsweeney:
Excellent suggestion! When the relevant code is resurrected in the utils project you might want to re-submit your PR there.

Issue #905, opened 2015-02-19
white, unclear images


digibeuk:
We experienced white and unclear logos for our customers when working with imagick.

This had to do with the fact that the alpha channel is extracted when the is_alpha check on line 4087 of CPDF::addPngFromFile returns true, but alpha channel is undefined. We observed this with gifs and pngs. These images have about these chars:

logo.gif GIF 300x253 300x253+0+0 8-bit PseudoClass 256c 14.4KB 0.000u 0:00.000
logo.png PNG 292x184 292x184+0+0 8-bit DirectClass 53.3KB 0.000u 0:00.000

We think these images are saved with cheap software which sets more colors than actually used in the image.

The solution is to check if the alpha channel is really defined before extracting the alpha channel and applying this is a mask to the image. The resulting mask is a strange errornous image which if used as mask for the image gives the errornous result. For all the above images the ALPHACHANNEL constant returned was 0 which is undefined.

digibeuk:
https://github.com/dompdf/dompdf/pull/906

bsweeney:
Excellent, thank you for the PR. I hope to be able to catch up on these soon(ish).

PR #906, opened 2015-02-19
do not try to apply alpha mask if no alpha channel is present

digibeuk:
This is a fix for  https://github.com/dompdf/dompdf/issues/905


bsweeney:
Looks good but can you target your changes to the develop branch? We would like to avoid merging changes directly into the master branch where possible.

digibeuk:
sure made a new one here : https://github.com/dompdf/dompdf/pull/985

Issue #104, opened 2012-12-06
Support of Softhyphens


PhenX:
_Original author: mamek...@gmail.com (November 29, 2009 13:05:32)_

<b>What would you like dompdf to do:</b>
Now Softhypens (&amp;shy;) will show every time and there are no line breaks 
after a soft hyphen character.

<b>Do you have an example:</b>

``` html
<div style="width:20px">Do&shy;nau&shy;dampf&shy;schiff&shy;fahrts&shy;gesell&shy;schaft</div>
```

_Original issue: http://code.google.com/p/dompdf/issues/detail?id=97_

1manfactory:
No hyphenation support? Too bad, that makes dompdf useless for me.

naitsirch:
This is fixed in #2009

PR #909, opened 2015-02-26
Simple fix for issue #104

melanholly:
Needed it today and this is the way I have fixed it. It is not really clean but that is my method.


bsweeney:
I think we'll want to do things a bit differently when we add official support. Still, this may prove useful for others so thanks for taking the time to post it. Anyone looking at the issue should see a reference to this PR.

PR #941, opened 2015-05-25
Add branch alias for dev branch

barryvdh:
So peope can require `0.7.x@dev` for the latest dev master.


barryvdh:
Also simplified instructions to use the simpler CLI interface.

bsweeney:
Anyone know of any situations where a user would be using composer without the CLI (e.g. a PaaS environment)? I think this is fine for README.md, but we might want to update the wiki with manual configuration instructions.

PR #942, opened 2015-05-25
Set default filename

barryvdh:
So stream() can be called without argument. (As it's used in the example)

### DIFF
--- a/src/Dompdf/Dompdf.php
+++ b/src/Dompdf/Dompdf.php
@@ -876,7 +876,7 @@ class Dompdf
      * @param string $filename the name of the streamed file
      * @param array $options header options (see above)
      */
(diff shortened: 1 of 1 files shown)


bsweeney:
Calling the `stream()` method without a file name does work as-is, but it's not well coded. PHP throws a warning if you exclude the file name parameter since a default isn't defined. Good catch.

barryvdh:
Yeah It throws a warning, which messes up the output (probably because of output buffering) and makes the pdf corrupt.) So either accept a default value or fail completely, so it's less confusing.

bsweeney:
Yeah, the output buffering issue is fairly annoying and affects a lot of people. At some point we need to do a better job of dealing with that possibility.

PR #948, opened 2015-05-28
Move functions to Dompdf\Helpers

barryvdh:
This PR cleans up the functions.php file:

 - Remove obsolete functions (not more needed in php5.3): `date_default_timezone_get`, `date_default_timezone_set`, `sys_get_temp_dir`
- Move `imagecreatefrombmp` to Helper class, reference that in AbstractRenderer
- Move debug functions to Helpers class, change `global $memusage` to static property on the class. (`def`, `bt`, `print_memusage`, `enable_mem_profile`, `mark_memusage`)

The mb_string functions are kept for compatibility, but should be recommended only when mb_string is not installed.


vkeranov:
And why keeping def, bt, print_memusage, enable_mem_profile, mark_memusage since they're not used at all?

barryvdh:
Not sure why they were there in the first place, so assuming the author has some benefit of those for testing purposes.

But I'm +1 for removing those. If people want to use them, they should probably define them in their own project.

vkeranov:
My personal opinion is... Keep it as clean as possible. The code has too many deprecated functions already. So, I hope some cleaning will come.

barryvdh:
@bsweeney you okay with removing them?

vkeranov:
Ofcourse @bsweeney has the final word. But I can not understand why the devs are so careful about the upcoming 0.7.0 release. Since those functions are not used at all, keeping them is not needed. We're not braking any backwards compatibility, so +1 for removing.

barryvdh:
Yeah this is a good time to cleanup :)

bsweeney:
I agree, let's remove it. I'll take another look, it's possible that these functions were used in other code that has been removed. Or we had plans to use some of it but never actually did.

barryvdh:
I've removed more the functions that aren't used, see latest commit.
Also removed `Helpers::DOMPDF_memory_usage()` and just used `memory_get_peak_usage(true)` directly in the 1 place it's used.

vkeranov:
@bsweeney, what do you think?

bsweeney:
:+1: though I still want to review the changes first.

bsweeney:
I think `Helpers::DOMPDF_fetch_url()` could have been potentially useful, though it was not actually referenced anywhere. I think it would be good to generalize resource fetching (i.e. all resources are fetched using a singular method) so that we can better manage how that's done in the future.

bsweeney:
This PR leaves just the MBString compatibility functions in functions.inc.php. I'll probably move the file to the forthcoming utils project and reference it in the requirements doc.

PR #949, opened 2015-05-28
Fix in_array call

barryvdh:


bsweeney:
It may be more relevant to just remove the TCPDF adapter for now. It has always been in a *very* experimental state. I've never actually tested it out myself.

barryvdh:
I haven't tested it, but phpstorm triggered a warning. The TCPDF driver would probably fail when you try to use it (depending on error settings). I'm not sure how many people use it, or if it even works with the latest TCPDF anyways.

bsweeney:
Probably not with the latest. I think it was initially developed a few years ago. If we do ever decide to support TCPDF it will have to be against a fixed release. Maybe I'll move that to the "utils" project I'm putting together.

bsweeney:
Merged since the actual code was bad. Regardless I'm going to remove the adapter shortly. Anyone interested in working out TCPDF support can grab a copy of the adapter as it existed at this point: https://github.com/dompdf/dompdf/blob/cca62f817d688ce6ad310bd5d7fbff93bf320bc4/src/Dompdf/Adapter/TCPDF.php

PR #951, opened 2015-05-29
Update version of PDF output

wakeless:
This is related to these PDFs getting caught up in virus scanners. No idea if this is a good idea or not, but it fixed the issue I was having.

### DIFF
--- a/lib/class.pdf.php
+++ b/lib/class.pdf.php
@@ -1868,7 +1868,7 @@ EOT;
     $this->checkAllHere();
 
     $xref = array();
-    $content = '%PDF-1.3';
+    $content = '%PDF-1.7';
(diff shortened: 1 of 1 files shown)


bsweeney:
Probably not a good idea. Before updating the PDF version we'll need to review the spec for structural changes. Also, that line just specifies the PDF spec used so I doubt it's causing the problem.

What you might want to try is changing the ["Creator" field](https://github.com/dompdf/dompdf/blob/v0.6.1/include/cpdf_adapter.cls.php#L190). I imagine that the virus scanners are basing the analysis on the PDF headers. The start of the PDF will be consistent across dompdf installations up to a point. We'll have to think about how to best address the issue.

Thanks for following up about the status. I'm going to create an issue where we can hash this out.

PR #954, opened 2015-06-04
Options pdf_backend / pdflib_license / admin_username aren't returnin…

scotteh:
Options pdf_backend / pdflib_license / admin_username aren't returning the correct value via Options::get(...).


bsweeney:
Thanks for that! Must've been a late night coding session :/

PR #956, opened 2015-06-05
Enforce extension requirements

barryvdh:
Require them in composer, so it will produce an error when you are missing something, instead of failing when the functions are actually called.


bsweeney:
Excellent!

PR #963, opened 2015-06-07
Invalid type check on $style argument, style can be passed in as an e…

scotteh:
Invalid type check on $style argument, style can be passed in as an empty array.

This fixed an issue where passing empty arrays into the 2nd argument of imagesetstyle() can cause subsequent functions such as imagerectangle() to crash PHP (Windows / PHP 5.5.12 / x86 MSVC11).

Haven't attempted to replicate this on any other setup.

### DIFF
--- a/src/Dompdf/Adapter/GD.php
+++ b/src/Dompdf/Adapter/GD.php
@@ -300,7 +300,7 @@ class GD implements Canvas
         $c = $this->_allocate_color($color);
 
         // Convert the style array if required
-        if (!empty($style)) {
+        if (is_array($style) && count($style) > 0) {
             $gd_style = array();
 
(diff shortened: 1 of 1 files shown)


bsweeney:
For completeness should we check that $style is actually an array first (e.g. `is_array($style) && count($style) > 0`)? Or perhaps it's sufficient to check that the element isn't empty and we can address any instances where an invalid type (e.g. int or string) is passed as they come up.

scotteh:
All cases appear to be arrays, but added in the additional type check just to be sure.

bsweeney:
I always go in expecting (hoping) methods are being called correctly, but you never know when something will slip through (and it's harder to diagnose with a loosely typed language).

PR #965, opened 2015-06-08
Update GD adapter to output using the current DPI and implement page s…

scotteh:
Update GD adapter to output using the current DPI and implement page support via new stream() / output() 'page' option.

DPI scaling is being handled internally, exposing the actual size via get_width() / get_height() was causing page margins and element margin/spacing to appear much smaller than they should at higher DPIs.

The ability to select the page to output / stream has been added via the 'page' option was also added.

Example to get the second page:
``` php
$dompdf->stream(array('page' => 2));
```

As you can already, the total number of pages can be obtained using:
``` php
echo $dompdf->getCanvas()->get_page_count();
```

### DIFF
--- a/src/Dompdf/Adapter/GD.php
+++ b/src/Dompdf/Adapter/GD.php
@@ -3,7 +3,7 @@
  * @package dompdf
  * @link    http://dompdf.github.com/
  * @author  Benj Carson <benjcarson@digitaljunkies.ca>
- * @author  Fabien MÃ©nager <fabien.menager@gmail.com>
+ * @author  Fabien Ménager <fabien.menager@gmail.com>
  * @license http://www.gnu.org/copyleft/lesser.html GNU Lesser General Public License
  */
 namespace Dompdf\Adapter;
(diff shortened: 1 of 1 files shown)


bsweeney:
Nice catch on the scaling issue!

I like the idea of being able to select the page to output. That would definitely make the GD adapter more useful.

vkeranov:
+1 for this one. Selecting the page to output is a must have.

scotteh:
A couple more fixes for the GD adapter that builds on the initial pull request.

* Scale font sizes down, they were noticeably larger. GD's internal DPI is 96, font sizes in dompdf are 72.
* Fix scaling issues by switching the DPI while rendering - restoring it when we're done. This enables internal functions - primarily length_in_pt() to return expected values.

vkeranov:
is this updated with the latest develop? I mean this one:
https://github.com/dompdf/dompdf/commit/62d72ad24b546e1482bf22c684857c2267046b5b
Because I really like this feature and I want to use it. Still don't know why this is not merged yet...

bsweeney:
Not yet, but I am planning to take a look at it for inclusion. Subscribe to this PR to be notified when it's been merged.

PR #977, opened 2015-06-24
PHP Fatal error : Class 'DOMPDF' not found

nutii:
This will fix fatal error.


bsweeney:
We're not accepting patches against the master branch. It's difficult to see from your patch what's going on, but if you want to work out the issue further start a new issue or visit the [support forum](http://groups.google.com/group/dompdf).

PR #981, opened 2015-07-07
Made the git installation instructions more explicit

davidsickmiller:


bsweeney:
We're not accepting updates to master. Plus check out the develop branch, things are changing ... soon.

PR #986, opened 2015-07-14
Fix the issue with Twitter Bootstrap Frameworks

reyko99:
//Minor fix (replace .ttf by the extension of remote font file)
Now is posible render pages with Twitter Bootstrap Framework integrated. 
But glyph-icons not are showed yet 
Tested with Twitter Bootstrap 3.3.1

### DIFF
--- a/include/font_metrics.cls.php
+++ b/include/font_metrics.cls.php
@@ -329,7 +329,7 @@ class Font_Metrics {
     
     $local_file = DOMPDF_FONT_DIR . md5($remote_file);
     $cache_entry = $local_file;
-    $local_file .= ".ttf";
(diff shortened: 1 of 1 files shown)


bsweeney:
What happens without this modification? dompdf currently only supports the TrueType font format. But I don't see that as being a reason we should hard code the format extension.

bsweeney:
Also, please target your PR at the develop branch.

reyko99:
Without this modification dompdf throws an exception error. The exception is throwed because try to load a file that not exist (<md5 generated file>.ttf) I 'm tested this using Laravel 5.01 framework.

PR #988, opened 2015-07-17
Fix the issue with Twitter Bootstrap Frameworks

reyko99:
When a page using TWB dompdf will be rendered to pdf, dompdf throws a FileNotFound exception(in my case I'am using Laravel 5.0) because not found a file <md5 generated filename>.ttf. The extension of the font is hardcoded. I know that dompdf only supports TrueType font format.

### DIFF
--- a/src/FontMetrics.php
+++ b/src/FontMetrics.php
@@ -213,7 +213,7 @@ class FontMetrics
 
         $local_file = $fontDir . DIRECTORY_SEPARATOR . md5($remoteFile);
         $cache_entry = $local_file;
-        $local_file .= ".ttf";
+        $local_file .= ".".pathinfo($remote_file,PATHINFO_EXTENSION);
 
         $style_string = $this->getType("{$style['weight']} {$style['style']}");
 


bsweeney:
(re-submit of PR #986)

bsweeney:
Do you know what format the remote file was in? And what, exactly, was the error? I'm just trying to understand exactly what was happening. It looks like dompdf will force a TTF extension, but that it doesn't really matter because php-font-lib will sniff the format based on the file headers. Is the actual error occurring elsewhere?

bsweeney:
Which version of dompdf are you using? I was only able to produce an error when using 0.7.0 beta 1 because of a bad path option. And then only a warning, not an exception, which may mean that turning down your error level in laravel may help.

bsweeney:
The actual issue I'm having is that php-font-lib consumes too much memory while processing the font. Not to mention it appears that maybe the WOFF font is being used instead of the TTF. Maybe, I'd have to look into that more.

Still, your change is important. I'll probably pull it in plus make whatever other changes are necessary to remove hard-coded extensions.

bsweeney:
I think the problem is more widespread than indicated in your patch. There are a number of locations where we refer to the filename extension. However, since that info is not recorded in the font family cache we have no way of knowing the appropriate letter casing.

I still have a bit more research to do before I post a fix but I think we need to do three things:

 1. Force a case when saving files to the system.
 2. Prefer the above-set casing, but check for other variants (maybe just check all uppercase and all lowercase).
 3. Update the documentation to indicate that filesystem-based options may be case sensitive, depending on the system.

Issue #639, opened 2013-05-12
Large number of inline elements reduces performance


francoisdusautoir:
Hello,

I have this error message when trying to get pdf file :  

Fatal error: Out of memory (allocated 57409536) (tried to allocate 1792 bytes) in /homepages/28/d468484513/htdocs/bundles/dompdf/include/text_frame_reflower.cls.php(374) : runtime-created function on line 1

The HTML code I want to generate in PDF :  http://calogerobasse.com/tablatures/2.html

My PHP code to generate it : 

// HTML

```
$tab = Tablature::find($id);
$tab->album = Titre::find($tab->titre->id)->album[0];
$html = View::make('pdf.tablature')->with('tab',$tab);
```

// PDF

```
$pdf = new DOMPDF(); 
$pdf->load_html($html);
$pdf->render();
$pdf->stream("tablature_".Str::slug($tab->titre->titre, '_').".pdf");
```

It's working when I have a few of content but when i have more (+100 words for example)  it's crashing on this error. I tried to edit php.ini on my webserver and changed memory_limit but it didn't change anything. I really need help to fix this.

PR #996, opened 2015-07-28
Build frame tree after unwanted/hidden nodes have been removed.

scotteh:
Build frame tree after unwanted/hidden nodes have been removed. Allows sibling text nodes prior to these to be modified/concatenated and the subsequent values to be reflected in the Frame tree.

### DIFF
--- a/src/Frame/FrameTree.php
+++ b/src/Frame/FrameTree.php
@@ -196,6 +196,8 @@ class FrameTree
             return $frame;
         }
 
+        // Remove unwanted nodes from the tree. Do this before further building
+        // of the tree as it may modify the textContent of previous nodes.
         for ($i = 0; $i < $node->childNodes->length; $i++) {
             $child = $node->childNodes->item($i);
 
@@ -207,7 +209,14 @@ class FrameTree
                 $child->parentNode->removeChild($child);
             } elseif (($child->nodeName === "img" || $child->nodeName === "IMG") && $child->getAttribute("src") == "") {
                 $child->parentNode->removeChild($child);
-            } elseif (is_object($child)) {
+            }
+        }
+
+        // Build the current level of the Frame tree.
+        for ($i = 0; $i < $node->childNodes->length; $i++) {
+            $child = $node->childNodes->item($i);
+
+            if (is_object($child)) {
                 $frame->append_child($this->_build_tree_r($child), false);
             }
         }
--- a/src/Frame/FrameTree.php
+++ b/src/Frame/FrameTree.php
@@ -196,28 +196,43 @@ class FrameTree
             return $frame;
         }
 
+        // Since childNodes->length is expensive to call in a loop store
+        // it into a variable and decrement as necessary.
+        $length = $node->childNodes->length;
+
         // Remove unwanted nodes from the tree. Do this before further building
         // of the tree as it may modify the textContent of previous nodes.
-        for ($i = 0; $i < $node->childNodes->length; $i++) {
+        for ($i = 0; $i < $length; $i++) {
             $child = $node->childNodes->item($i);
+            $nodeName = $child->nodeName;
 
-            if (in_array($child->nodeName, self::$HIDDEN_TAGS)) {
-                if (($child->nodeName !== "head" && $child->nodeName !== "HEAD") && ($child->nodeName !== "style" && $child->nodeName !== "STYLE")) {
-                    $child->parentNode->removeChild($child);
+            if (in_array($nodeName, self::$HIDDEN_TAGS)) {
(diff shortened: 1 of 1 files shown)


scotteh:
Using the following example - once the comment is removed 'Second Text' and 'First Text' are automatically merged into a single text node.

Currently 'Second Text' would be omitted from dompdf's output as 'First Text' has already been added to the frame tree and the merged node is a new object.

``` html
<div>
First Text
<!-- Comment -->
Second Text
</div>
```

This patch uses two passes instead of the current single. First to remove nodes and the second to add them to the frame tree.

bsweeney:
How much of a performance hit do you see with this? I'm guessing not much unless the node has a large number of children (which is already an issue, see #639).

If you have a chance to profile the code I have some thoughts on changes that *could* improve performance. These suggestions all fall in the variable vs. object property access speed question. I haven't been keeping up with how well PHP optimizes the byte-code, does anyone know where PHP falls in relation to this question (e.g. is there a significant different between accessing an object property vs. pushing into a variable)?

 - Push the length of the `childNodes` collection into a variable on the second pass since that value shouldn't be modified.
 - Replace `$child->parentNode` with `$node`.
 - Push the `nodeName` into a variable and set the case so the conditionals are simplified.

Otherwise I like this. I was thinking we might use a similar method to pull out non-flow nodes (e.g. fixed-position elements) and push them to the top of the frame stack.

scotteh:
Not much... or at least I thought!

Did some quick crude profiling based on your suggestions.

As a general rule object access does appear to be slower at somewhere around 1.7 times - at least in a `for ($i = 0; $i < $object->length; $i++) { .. }` vs `for ($i = 0; $i < $length; $i++) { .. }` scenario, though we are talking milliseconds.

However... in the case of `for ($i = 0; $i < $node->childNodes->length; $i++) { .. }` it's significantly more. Its surprisingly somewhere in the vicinity of 1000 times slower than `$length = $node->childNodes->length; for ($i = 0; $i < $length; $i++) { .. }`. Probably to do with libxml more than anything else.

After profiling the length I half expected saving $child->nodeName into a variable to be significantly faster, but it appears to be pretty much the same.

Not sure why I didn't just used `$node` rather than `$child->parentNode`. Based on the nodeName test i wouldn't expect this to be huge, but $node is already stored so it's definitely worth changing.

Profiling script for $obj->childNodes->length:
``` php
<?php

$passes = 5;
$results = array();

$html = '<html><head></head><body>' . str_repeat('<i></i>', 10000) . '</body></html>';
$dom = new \DOMDocument();
$dom->loadHTML($html);

$body = $dom->documentElement->childNodes->item(1);

for ($j = 0; $j < $passes; $j++) {
    $time = microtime(true); 

    // Direct property access
/*
    for ($i = 0; $i < $body->childNodes->length; $i++) {
        // Do nothing
    }
*/

    // Read from variable
    $length = $body->childNodes->length;

    for ($i = 0; $i < $length; $i++) {
        // Do nothing
    }

    $exectime = microtime(true) - $time;

    echo sprintf("%f", $exectime) . "<br>\n";
    $results[] = $exectime;
}

echo 'total: ' . sprintf("%f", array_sum($results)) . "<br>\n";
echo 'avg: ' . sprintf("%f", array_sum($results) / $passes) . "<br>\n";
```

bsweeney:
Awesome, thanks for investigating! I'm surprised as well about that slowdown. Your guess makes sense to me, I'm sure PHP can optimize internal object access but when using external libraries there's probably not a whole lot it can do.

bsweeney:
I'm still looking into it, but there appears to be an issue related to this change. White space is inserted in place of the removed element. Try rendering the following:

```
  <p>lorem ipsum</p>
  <!-- -->
  <p>lorem ipsum</p>
  <p>lorem <!-- --> ipsum</p>
```

bsweeney:
So this commit didn't create a new issue, it exposed an existing one that was introduced in commit 439e6477cae04432a88fd44c637098923a9bd961. In the loop dompdf was moving on to the next index of the childNodes collection on each iteration of the loop regardless of what action was taken. If a node was removed from the childNodes collection (shortening the collection by one element) then the following node would be skipped.

Try rendering the following in 0.7 beta:

```
<p>to be <!-- --> or not to be</p>
```

There is a second issue that has always existed that is relevant to my [previous comment](https://github.com/dompdf/dompdf/pull/996#issuecomment-142009368). Non-rendering nodes (e.g. HTML comments) can affect the layout of the page when they split text nodes. The adjacent text nodes cause a failure in the white-space collapse logic.

bsweeney:
@scotteh I'm not sure that adjacent text nodes are merged. I think your commit worked around the problem I identified because goes back through the childNodes collection, catching any nodes that would have been skipped in the first previous iteration.

PR #997, opened 2015-07-28
Move transform style logic from setter to getter.

scotteh:
Move transform style logic from setter to getter. Percentage values require width/height to be set before they can be calculated.

Example that gets evaluated as '0' due to lack of width/height.

``` css
-webkit-transform: translateX(20%);
```

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -2268,20 +2268,19 @@ class Style
 
         $this->_props["size"] = $computed;
     }
-
     /**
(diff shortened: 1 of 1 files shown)


bsweeney:
Trying to start using `git am` to merge PRs. Merged ok but I forgot to amend the commit to indicate it closes this issue. So...

closed by commit a7e1c657919d91b78fd2db16bc12e2c9a099cc78

PR #1011, opened 2015-09-08
Check if the style isn't empty before setting it

nappa85:
Setting an empty style results in a "Floating point exception" on the next image operation.


bsweeney:
Presumably for performance reasons? Can you create a new PR on the develop branch?

nappa85:
A "Floating point exception" isn't a performance reason...

bsweeney:
Sorry ... should have looked at your comment first :/

PR #1024, opened 2015-09-24
Adjust variable filename, prevent case sensitive config PHP error.

douglasmiguel:
A simple adjustment was made in the name of the variable ($fileName) to prevent an error caused by case sensitive in PHP.


bsweeney:
Good catch! I made a minor tweak and added the fix to the PDFLib and GD adapters.

PR #1030, opened 2015-09-29
PHP 7 Compatibility.

IceReaper:
Wanted to share my local changes to make it run with php 7 - Feel free to merge or comment if i should change something, then i will do so.


PHPGangsta:
Thanks for your work Andre!
The two changes regarding uniform variable syntax I had solved the same way with the curly braces, and the rest looks good I would say.

PR #1053, opened 2015-11-04
Remove donation note from 'Limitations' section

akshayagarwal:


akshayagarwal:
The build fails with 'No Rakefile found'

My change is a trivial text change in README.md It should really have nothing do with the rakefile.

bsweeney:
Closed by a0eb016

PR #1059, opened 2015-11-22
Set MediaBox paramter for each page

samuelvogel:
Sets the `MediaBox` for each page to the same value as for the whole document. This fixes printing 100 x 150 mm PDFs (similar to DIN A6) with Google Cloud Print.

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -1444,6 +1444,15 @@ EOT;
 
             case 'out':
                 $res = "\n$id 0 obj\n<< /Type /Page";
+                if ($this->currentPageSize['width'] > 0 && $this->currentPageSize['height'] > 0) {
+                    $res .= "\n/MediaBox [" . sprintf(
+                            '%.3F %.3F %.3F %.3F',
+                            0,
+                            0,
+                            $this->currentPageSize['width'],
+                            $this->currentPageSize['height']
+                        ) . "]";
+                }
                 $res .= "\n/Parent " . $o['info']['parent'] . " 0 R";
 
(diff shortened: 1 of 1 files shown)


bsweeney:
That's interesting. Can you relate the nature of the problem? I'll have to look into the issue a bit more just to see if the problem is more widespread.

samuelvogel:
This is an illustration of the problem that happens without this change:
din a6 print render problem

bsweeney:
yikes

samuelvogel:
But only with Google Cloud Print, it works fine when printed from OS X/Windows. If saved as new file with the Mac Preview.app, it prints fine with GCP as well.

bsweeney:
Good to know. I don't see any reason not to pull this in, I just want to look at it a bit more first to see if there's anything else we can do as well.

samuelvogel:
:+1:

bsweeney:
So I tried to find what was going on here, but I haven't really come across anything useful. The closest I've found is this: https://code.google.com/p/chromium/issues/detail?id=199191. But that appears to be a fit-to-page issue.

All the samples I've seen place the MediaBox on the page object themselves and not on the page tree, though the PDF reference states that the MediaBox is inheritable from the page tree. GCP does appear to understand the size of the page, so I'm not sure why it's placing it in the wrong location when the MediaBox is inherited.

bsweeney:
I tweaked your commit to use the MediaBox property from the parent page tree. Bonus, I think this brings us a bit closer to being able to provide different page sizes/orientations within the same document.

Issue #1062, opened 2015-12-01
fontDir option not working on 0.7.0-beta2


skotos:
When setting the fontDir option in Dompdf 0.7.0-beta2, the font is ignored.  Here is what I can gather...
- Dompdf->_construct() creates a new instance of FontMetrics based on the default values of Options.  This object has only the supplied fonts in fontLookup.
- Dompdf->render() re-creates a new instance of FontMetrics which includes any changes made to Options (via set_options).  This object has my extra fonts included in fontLookup.
- When Cpdf calls getFont, it must be calling the original instance, as fontLookup doesn't have the extra fonts included.

My current work-around is the following patch:

``` diff
diff --git a/src/Dompdf.php b/src/Dompdf.php
index 364489f..9bddbd7 100644
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -256,9 +256,9 @@ class Dompdf
     /**
      * Class constructor
      */
-    public function __construct()
+    public function __construct($options = null)
     {
-        $this->setOptions(new Options);
+        $this->setOptions(new Options($options));

         $versionFile = realpath(__DIR__ . '/../VERSION');
         if (file_exists($versionFile) && ($version = file_get_contents($versionFile)) !== false && $version !== '$Format:<%h>$') {
```

Which I then pass the fontDir, (as well as other options like subsetting,) to Dompdf on creation rather than using set_options.  This fixes the issue.

If this is the correct way of doing this, should I create a pull request with the above patch?  Otherwise, what should I be doing.

Two reasons I have for using a seperate fontDir over font-face css:
1. I want to leave /vendor/dompdf as read-only so that it doesn't get messy with composer.
2. Save on extra http calls to pull in the fonts.

bsweeney:
I definitely recommend using a different fontDir for exactly the reason of keeping the dompdf installation clean. I've been thinking being able to set the options in the constructor would be a nice addition, so go ahead an submit that.

Your points are spot on. When dompdf is instantiated it instantiates an instance of the Canvas. Another instance is later instantiated, which is problematic for exactly the reason you've noted. There are a few issues that need to be worked out to remove this duplication:

 1. When Canvas is instantiated in the dompdf constructor it doesn't receive any updates to the dompdf configuration.
 2. I'm not entirely certain (without more research) if there are any problems instantiating the Canvas object only during render. There is one drawback in that you have to render the document before you can access the canvas object.

[...]

PR #1063, opened 2015-12-01
Allow for passing of options during Dompdf construction

skotos:
Pass an array of options via the constructor.  Whilst handy long-term, it is also a work-around for issue #1062


garex:
+1

PR #1068, opened 2015-12-16
Helpers::getContent() to get rid of allow_url_fopen off

asterixcapri:
Currently Dompdf is not working with external images and css if allow_url_fopen is off. I created Helpers::getContent() method to wrap file_get_contents() and to use curl when allow_url_fopen is not enabled.


bsweeney:
sweet

asterixcapri:
Could you merge this pull request?

bsweeney:
I plan to pull this soon, but I first want to compare it to previous code we had in the library (that was never used). I was thinking that in order to maintain compatibility with the current context configuration I'll modify it to prefer `file_get_contents` over curl.

PR #1075, opened 2015-12-21
at least 1 example

danielecr:
how to use is not just require staff


bsweeney:
That section is actually intended just to show how to install dompdf. But point taken, we don't have an example of how to use it like we do for the [upcoming release](https://github.com/dompdf/dompdf/tree/v0.7.0-beta2#quick-start).

danielecr:
so, if someone is unable to use the newer upcoming release he must live without example?

bsweeney:
That's not quite what I meant. We'll add an example, I like what you did with #1078 as far as adding example usage.

PR #1096, opened 2016-02-01
Fix "Permission denied" error when file not found on Windows

godrar:


bsweeney:
Is this related to #1096?

bsweeney:
FYI, I may not pull this on the 0.6.x branch since we're about to release 0.7.0, but it looks to be relevant to 0.7.0 as well.

godrar:
File was open and never closed. On unlink function php throw "Permission denied" error.

PR #1097, opened 2016-02-01
Fix "Permission denied" error when file not found on Windows

godrar:


bsweeney:
Is this a warning or error? What is `$remoteFile` in your case?

godrar:
Warning: file_put_contents(C:\Users\PRACOW~1\AppData\Local\Temp/ac7f5fff3847d015f6fec14a7c1e75f9): failed to open stream: Permission denied 

$remoteFile: C:\xampp\htdocs\workspace\project1\web\assets\css/../fonts/glyphicons-halflings-regular.ttf 
Font does not exist in folder.

bsweeney:
A warning shouldn't cause problems for rendering the PDF (depending on your PHP/app configuration). But catching for this is a good idea.

godrar:
It would be cool.. File is lock and it can't be unlinked. 
Warning cause problems. Every warning stops my app :)
Thx

bsweeney:
This PR is problematic in that if `$remoteFile` is a URI with protocol and domain (e.g. `http://example.com/somefont.ttf`) the check will fail even if the file exists. I did come up with a tweak, check out commit b5fb3246deccdeda3b0b28e548da255901ed2bb9.

PR #1108, opened 2016-02-13
Few optimization patches

efbiaiinzinz:
First patch reduces the memory churn a bit for the case where value is already cached.
Second patch avoids calling get_option('debugLayout')+get_option('debugLayoutInline') for each child similar to the approach done in Style.php

### DIFF
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -190,8 +190,8 @@ class CPDF implements Canvas
         $this->_pdf = new \Cpdf(
             $size,
             true,
-            $dompdf->get_option("font_cache"),
-            $dompdf->get_option("temp_dir")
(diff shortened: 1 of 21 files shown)


efbiaiinzinz:
This patchset offers approximately 3% speedup: 1% comes from the first patch, additional ~1.5% comes from caching the option values for usage in loops and final 0.5% from using getters directly instead of running through bigger Options->get() function that does lots of string comparisons and most commonly used debugCss comparison is 38th comparison in there.

bsweeney:
@efbiaiinzinz just FYI the merge didn't pick up your user account because of the email used. If you want your account to show up in the commit history (rather than your name) add the sportlyzer email address to your account.

efbiaiinzinz:
Thanks for the tip, added that e-mail to my github also :)

PR #1114, opened 2016-02-18
Fix an issue in Block reflower resulting in style values like nonept and autopt

efbiaiinzinz:
Block reflower always appended "pt" to style values regardless of the original value, causing Style->__set() to be called with values like "autopt" and "nonept". Verified that pdf output for html with multiple div (block) elements remains the same.


efbiaiinzinz:
Hm, jumped a gun a little bit, "pt" is not needed at all, length_in_pt already accepts plain numbers also, assumes them to be in pt units.

PR #1115, opened 2016-02-18
Fix FrameReflower\Block always appending "pt" causing style values like autopt and nonept

efbiaiinzinz:
When tracing through the Style->__set() calls I found that FrameReflower\Block.php forcibly appends "pt" to margin left/right, width and left/right causing weird style values such as nonept and autopt.
The html sample I was testing remained identical, when comparing bytewise, only the creation/modification block was different.

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -178,6 +178,11 @@ class Style
     private $__font_size_calculated; // Cache flag
 
     /**
+     * The computed bottom spacing
+     */
+    private $_computed_bottom_spacing = null;
+
+    /**
      * The computed border radius
      */
     private $_computed_border_radius = null;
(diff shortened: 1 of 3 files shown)


bsweeney:
Good catch. I wonder what issues this might address. I found a similar action in FrameReflower/Image, but it seems to be a bit more thorough in using a numeric value.

efbiaiinzinz:
Not sure what may have gotten fixed exactly, probably it did not affect much. The appended pt causes the length_in_pt andle it in strpos("pt") block and take the first part and cast it to float. so nonept and autopt were both handled as 0 values. Correct "none" handling places (_set_style_side_width_important() and in set_margin() and in set_padding()) convert it to 0px anyways before storing it for margin/padding/border-width, so passing in nonept and none shoud not change anything in end results according to code.

The length_in_pt() seems to have some special logic for auto, so when at least one element is auto, "auto" is returned regardless of other values. Which should be correct I guess? Are there some more extensive tests that could be run?

I'm interested to get this patch accepted or rejected so I can resume sending in optimization patches, as I do not like to work on parallel branches in different states :)

bsweeney:
This has already been merged into the develop branch (see commit 4ab5ab251ad77191c0539c93806006a50aa0e453). I don't always use the github merge process ... rarely these days. This issue will auto-close once develop has been merged back into master.

efbiaiinzinz:
Ah, indeed, I didn't see the "added a commit that referenced this pull request" notification :)

efbiaiinzinz:
I think I have to close this pull request, otherwise new commits I push to my fork's develop branch keep getting added here.

Issue #623, opened 2013-04-19
Call to a member function split() on a non-object


jamieburchell:
Hi

Using the trunk version of dompdf with a single table spanning 2 pages I receive the fatal error:

Call to a member function split() on a non-object ... dompdf\include\page_frame_decorator.cls.php on line 534

Sample HTML
http://www.jamieburchell.com/dompdf/fatal-split.htm

Platform:
Windows 7 php 5.2.17

This does not seem to happen in 0.6.0b3

Regards,

Jamie

DerDu:
Same here... +1

19 Apr 2013 ?? - ITS TIME TO FIX THIS

bsweeney:
@DerDu sample code is the best way to help us help you. So I think the reason I wasn't able to reproduce the issue is that I typically run with the HTML5 parser enabled. When I turned disabled that library dompdf threw the noted exception.

It looks like the main cause of the issue is the white space between the cells. When I enable the HTML5 parser the space is removed and the document render just fine. So, for now, while not ideal, you can try removing any space from between your table (table, tr, td, etc.) elements. Or maybe just try using the HTML5 parser.

The changes suggested in #1125 will prevent the exception from being thrown so that'll be pulled in the for the 0.7.0 release. We'll try to fully address the issue for 0.7.1.

bsweeney:
I'm also wondering if commit 2137e2b8e26e3ea85535080f48abc017e2dc7b40 could be relevant.

bsweeney:
No recent feedback, and so far as I can tell the issue seems to be addressed in the latest develop branch. Closing for now unless further info is incoming.

PR #1125, opened 2016-02-24
Fix fatal error in Dompdf\FrameDecorator\Page::check_page_break

kirugan:
How to check bug (master branch):
```php
<?php
require __DIR__ . '/vendor/autoload.php';
use Dompdf\Dompdf;

error_reporting(E_ALL);
ini_set('display_errors', true);

$html = "<html><head></head><body><table style='margin-top: 678px'>
<tbody><th><td>Test</td></th><tr><td>Test2</td></tr></tbody></table></body></html>";
$dompdf = new Dompdf();
$dompdf->loadHtml($html);
// (Optional) Setup the paper size and orientation
$dompdf->setPaper('A4', 'landscape');

// Render the HTML as PDF
/* Fatal error: Call to a member function split() on a non-object in 
   /var/www/shared/dompdf/src/FrameDecorator/Page.php on line 594
*/
$dompdf->render();
```
My PHP version: **PHP 5.5.9-1ubuntu4.14 (cli)**

### DIFF
--- a/src/FrameDecorator/Page.php
+++ b/src/FrameDecorator/Page.php
@@ -588,7 +588,7 @@ class Page extends AbstractFrameDecorator
         // If we are in a table, backtrack to the nearest top-level table row
         if ($this->_in_table) {
             $iter = $frame;
-            while ($iter && $iter->get_style()->display !== "table-row")
+            while ($iter && $iter->get_style()->display !== "table-row" && $iter->get_style()->display !== 'table-row-group')
                 $iter = $iter->get_parent();
 
             $iter->split(null, true);


bsweeney:
Odd. So far I'm still unable to reproduce the issue. The output is definitely wrong ("Test" is missing from the output), but I get no PHP errors/notices.

bsweeney:
Actually, the text is in the PDF, just placed out of the bounds of the visible page. When I get a chance I'll see if your commit addresses the unexpected text placement.

bsweeney:
Do you experience the same issue if the first table element is a THEAD instead of a TBODY?

bsweeney:
@kirugan do you have any more feedback?

kirugan:
@bsweeney, sorry for late response.
Here is html that cause the main trouble: [cv.txt](https://github.com/dompdf/dompdf/files/154727/cv.txt).
If I put `<thead>` in `<tbody>` in this file, the error remains the same.

bsweeney:
No worries, I just wanted to get a fuller picture of the issue before I pulled in your changes. Still not able to reproduce the issue ... the styling may be the key to reproducing with the updated. At any rate, with confirmation that thead also causes issues I'll probably tweak your commit a bit to account for other variations.

kirugan:
@bsweeney I can attach Vagrantfile if you want to see this error. Do you need it?

bsweeney:
@kirugan sure ... I think that would help. Thanks :)

kirugan:
Just run php `dompdf/test.php` from vagrant home directory
[pdf.zip](https://github.com/dompdf/dompdf/files/160322/pdf.zip)

bsweeney:
OK, so I'm just now seeing this, but your HTML is structurally invalid. You have a TD inside the initial TH. While browsers can handle this by setting up anonymous table elements dompdf can not (yet). I wasn't running into any issues because I had the HTML5 parser enabled, which was correcting the document structure.

Always good to run your document through a HTML validator if you run into errors like this.

I'm going to push this to 0.7.1 to fix so we can take a closer look at this issue and build in better error handling.

bsweeney:
Issue #623 is relevant here. It looks like overly-tall table cells can also result in this exception. Something like the following:

```html
<table>
<tr><td style="height: 150px;"></td></tr>
<tr><td style="height: 150px;"></td></tr>
<!-- etc, etc -->
</table>
```

The issue actually appears to be the white space between the rows. With the HTML5 parser enabled the white space is removed and the document renders without throwing an exception.

The issue isn't quite resolved, though. This will prevent the exception from being thrown, but for some reason the styling (specifically the height) is lost after a certain number of elements are rendered.

Though more work needs to be done to fully address what's going on I'll go ahead and pull this in for 0.7.0.

bsweeney:
Actually, I think the style loss may be unrelated.

bsweeney:
If you explicitly wrap the table body in a tbody element the styling is maintained.

PR #1137, opened 2016-03-04
Improved PHPdoc of stream()

ThomasLandauer:
Is there a specific reason why you're not setting the 'Content-Type:' header to 'application/pdf'? Can I add this here?

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -887,19 +887,18 @@ class Dompdf
     /**
      * Streams the PDF to the client
      *
-     * The file will open a download dialog by default.  The options
-     * parameter controls the output.  Accepted options are:
+     * The file will open a download dialog by default. The options
+     * parameter controls the output. Accepted options (array keys) are:
      *
-     * 'Accept-Ranges' => 1 or 0 - if this is not set to 1, then this
-     *    header is not included, off by default this header seems to
-     *    have caused some problems despite the fact that it is supposed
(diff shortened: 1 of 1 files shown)


bsweeney:
The content-type header is set by either a) the adapter (in the case of PDFLib and GD) or b) the PDF generating back-end (in the case of CPDF).

ThomasLandauer:
Yeah, thanks, you're right!
Details: When calling `$dompdf->stream('dom.pdf');` I get the correct Content-Type. However, when "wrapping" it into a Symfony Response object like this
`return new Response ($dompdf->stream('dom.pdf'));`
... then Symfony sends Content-Type `text/html`

bsweeney:
Ah, now that's interesting. It's possible that Symfony is buffering the output and setting it's own content type header. You might also have to set the content type manually if it sets a default of "text/html".

Actually, you may want to use the `output()` method and rely on Symfony to handle the response to the client.

```
return new Response(
    $dompdf->output(),
    Response::HTTP_OK,
    array('content-type' => 'application/pdf')
);
```

You would have to set the headers yourself that are typically set by dompdf. The Symfony documentation may be useful here: http://symfony.com/doc/current/components/http_foundation/introduction.html#serving-files.

ThomasLandauer:
Thanks for the link - that's just perfect! :-)

BTW: Please merge the pull request.

Issue #1027, opened 2015-09-25
PHP 5.6.13 for Windows - DOMNode::cloneNode(): ID already defined warning


garethellis36:
I have just updated to PHP 5.6.13 (for Windows - NTS, VC11 x86) and have started getting this warning in my application logs. This issue was not happening prior to updating PHP (was previously on 5.6.6) and I have not changed any application code. I am using DOMPDF v0.6.1 installed via composer.

I am able to reproduce on my development machine. The warning is thrown on calling `DOMPDF::render()`. Below is the HTML I am inserting using `DOMPDF::load_html()`.

I am getting two warnings:

```
Warning (2): DOMNode::cloneNode(): ID header already defined [APP\Vendor\dompdf\dompdf\include\frame_decorator.cls.php, line 128]
Warning (2): DOMNode::cloneNode() [domnode.clonenode]: ID footer already defined [APP\Vendor\dompdf\dompdf\include\frame_decorator.cls.php, line 128]
```

As you can see, there is only one element with ID "header" and one with ID "footer".

```html
<html>
	<head>
	<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
	<style>
		@page { margin: 60px 50px 35px; }
    	#header { position: fixed; left: 0px; top: -45px; right: 0px; height: 50px; text-align: center; }
    	#footer { position: fixed; left: 0px; bottom: -15px; right: 0px; height: 50px; text-align: center; border-top: 1px solid black; }
    	/*#footer .footer-message:after { content: "Page " counter(page); }*/
		body { font: 12px "Calibri", "Helvetica", "Sans-Serif"; }
		th, td, p { font: 12px "Calibri", "Helvetica", "Sans-Serif"; }
		p { line-height: 1.5; page-break-inside:avoid; }
		.title { font-size: 16px; }
		table { margin-bottom: 10px; }
		thead th { font-weight: bold; }
	</style>
	</head>

	<body>
			<div id="header">
			</div>
			<div id="footer">
			</div>
	</body>
</html>

bsweeney:
When updating PHP did libxml also update? Do you know what version is on your system (you can check `phpinfo()`)?

bsweeney:
To work around the issue for the time being you can style using classes rather than IDs.

garethellis36:
@bsweeney - thanks for the workaround, that solves the problem.

[...]

PR #1142, opened 2016-03-07
First step to fix bug #1027

ThomasLandauer:
I'm moving the `id`'s content to `data-dompdf-clone-temp-id`. To be honest, I don't understand why this code then works:
````php
        $dompdf = new Dompdf();
        $dompdfOptions = new Options();
        $dompdfOptions->set('defaultPaperSize', 'legal');
        $dompdf->setOptions($dompdfOptions);
        $html = '<style>#foo {color:green; }</style>';
        $html .= '<div id="foo">';
        for ($i=0; $i<1000; $i++)
        {
            $html .= 'Lorem ipsum ';
        }
        $html .= '</div>';
        $dompdf->loadHtml($html);
        $dompdf->render();
````
=> I get 3 pages in green. But shouldn't the `id` be gone? I'm not setting `data-dompdf-clone-temp-id` back to `id` anywhere!

### DIFF
--- a/src/FrameDecorator/AbstractFrameDecorator.php
+++ b/src/FrameDecorator/AbstractFrameDecorator.php
@@ -637,6 +637,12 @@ abstract class AbstractFrameDecorator extends Frame
         }
 
         $node = $this->_frame->get_node();
+        
+        if ($node->hasAttribute("id"))
+        {
+            $node->setAttribute('data-dompdf-clone-temp-id', $node->getAttribute('id'));
+            $node->removeAttribute('id');
+        }
 
         $split = $this->copy($node->cloneNode());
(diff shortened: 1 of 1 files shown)


bsweeney:
By this point the styles have all been applied. First pass sets up the frame tree, second pass applies styles to all elements, third pass renders the elements. It's during the third pass that nodes are split to accommodate page breaks.

ThomasLandauer:
Well, so you mean, we're already done!?

bsweeney:
`cloneNode` is also used in the inline decorator. I'd also want to look to see if there are any other method that could potentially result in the same issue. But otherwise I think that pretty much should cover the issue.

PR #1158, opened 2016-03-30
Auto-width calculation, and other related bootstrap improvements

scotteh:
Bootstrap Improvements:
- Auto-width calculation for floated and inline-block elements.
- Fix style specificity based on the origin of the stylesheet.
- Add :focus psudo-class to stylesheet parser.
- Modify default UA stylesheet to reflect auto-width changes.
- Fix 'content' css property being set to 'normal' for all :before/:after selectors before determining their specificity.
- Add extra check to prevent Helpers::build_url() from throwing a warning for URIs of 1 character (eg: '#').

Visually improves the usability of bootstrap (and most likely any other inline-block/float heavy CSS) substantially. Included PDF examples of the documentation both pre and post changes (see below).

Auto-width appears to work OK based on my testing so far, any feedback on the implementation is welcome.

Included quite a few changes in this PR, if you would prefer these (or some of) individually I can split them up.

[badges-pre-changes.pdf](https://github.com/dompdf/dompdf/files/195505/badges-pre-changes.pdf)
[buttons-pre-changes.pdf](https://github.com/dompdf/dompdf/files/195503/buttons-pre-changes.pdf)
[buttons-post-changes.pdf](https://github.com/dompdf/dompdf/files/195504/buttons-post-changes.pdf)
[badges-post-changes.pdf](https://github.com/dompdf/dompdf/files/195502/badges-post-changes.pdf)

### DIFF
--- a/lib/res/html.css
+++ b/lib/res/html.css
@@ -373,8 +373,11 @@ input[type=reset],
 input[type=file],
 button {
   background: #CCC;
-  width: 8em;
   text-align: center;
+}
+
+input[type=file] {
+  width: 8em;
 }
 
 input[type=text]:before,
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -777,6 +777,32 @@ class Style
         return $this->_prop_cache[$prop] = $this->_props[$prop];
     }
 
+    /**
+     * Similar to __get() without storing the result. Useful for accessing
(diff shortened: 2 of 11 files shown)


bsweeney:
Very nice!

The changes do not appear to be so complex that it will be a problem to review as a single commit.

PR #1159, opened 2016-03-30
Fix binary space replacement in UTF16 text

merlinus1:
space char in UTF16 is actually \x00\x20.
str_replace(' ', ...) replaces only \x20. So some unicode chars gets corrupted. For example (Р) CYRILLIC CAPITAL LETTER ER \x04\x20


bsweeney:
Good catch! We won't be pushing any more updates to the 0.6.x code base, but 0.7.0 is just about ready to release.

Issue #1129, opened 2016-02-28
error Cellmap


idairaaleman:
hi,i have the next problem;

Fatal error: Call to undefined method DOMText::getAttribute() in /home/tonotour/public_html/dompdf/src/Cellmap.php on line 555

the line 555 is $colspan = $node->getAttribute("colspan");

what's happen?

bsweeney:
Difficult to say without seeing a sample of the HTML. However, as a first shot I'd say run your HTML through a validator. Errors like this typically show up when there are structural issues with the document that dompdf is unable to handle. You also might try enabling the HTML5 parser.

idairaaleman:
hi! this is my html;

<patTemplate:tmpl name="pageoutput" unusedvars="strip">
<style>
he i have my style
</style>

<div class="container">

<table width="100%" class="table table-striped table-bordered table-condensed" sTYLE="table-layout:fixed">
		<thead>
			<tr>
				<td colspan="4"><center><img src="/images/logohead.png" WIDTH=240 HEIGHT=126 ALT="logo empresa"><br>
				<center>TONOTOUR ESTANCIAS VACACIONALES,S.L<br>
				Teléfono: 922xxxxxxx.</center></td>
				<td colspan="5">{CLIENT_DETAILS_TEMPLATE}</td>			
				
			</tr>

			<tr>
				<th colspan="4">{HINVOICENO}: {ID}</th>
				<th colspan="3"><fon size=1>{HRAISED}: {RAISED}</fon></th>
				<th> {_JOMRES_BOOKING_NUMBER}: {BOOKING_NUMBER}</th>
			</tr>
			<tr>
				<th colspan="9">{BUSINESS_DETAILS_TEMPLATE}</th>
			</tr>				

			<tr>
				<th colspan="3" width="10%"><font size=1>{HLI_NAME}</font></th>
				<th><font size=1>{HLI_INIT_PRICE}</font></th>
				<th><font size=1>{HLI_INIT_QTY}</font></th>
				<th><font size=1>{HLI_INIT_DISCOUNT}</font></th>
				<th><font size=1>{HLI_INIT_TOTAL}</font></th>
				<th><font size=1>{HLI_TAX_RATE}</font></th>
				<th ><font size=1>{HLI_TAX_DESCRIPTION}</font></th>
				<th><font size=1>{HLI_TAX_AMOUNT}</font></th>
				<th><font size=1>{HLI_INIT_TOTAL_INCLUSIVE}</font></th>
[...]

PR #1161, opened 2016-03-31
Fix a couple of issues with table-cells

### LINKED ISSUE #1103: First table cell misalignment
If I apply one or more css styles to "*:after" rule, then the first cell in any table on that page will be aligned to the left of any other cells in the first column. I am using v0.6.2 to perform the conversion to PDF.

scotteh:
Fix a couple of issues with table-cells:
 * Table-row will now ignore :before/:after frames and not include them as cells.
 * Text frames between a </tr> closing tag and <td> opening tag will no longer be added as a table-cell. Fixes a potential issue with the getAttribute() method not existing in \DOMText.

Fixes tables in bootstrap, and also most likely fixes #1103.

### DIFF
--- a/src/Cellmap.php
+++ b/src/Cellmap.php
@@ -516,7 +516,10 @@ class Cellmap
 
             $start_row = $this->__row;
             foreach ($frame->get_children() as $child) {
-                $this->add_frame($child);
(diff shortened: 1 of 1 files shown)


scotteh:
Text between closing / opening tags should hopefully fix #1129.

PR #1162, opened 2016-03-31
Update PHP version mentioned in README

ADmad:
Since the README has instructions for using the 0.7@beta version, which uses namespaces, min. PHP version needed would 5.4.


scotteh:
Namespaces are supported in PHP >= 5.3. See the version requirements on the following page: http://php.net/manual/en/language.namespaces.rationale.php

ADmad:
Sorry, I feel so stupid.

bsweeney:
@ADmad no worries, we all have days like that.

PR #1191, opened 2016-05-08
Fixes an issue with loading fonts on Windows when a custom fontDir is specified.

BrunoDeBarros:
With the below code, dompdf was having problems when a user tried to run it on Windows. The fact that "/lib/fonts" didn't match "\\lib\\fonts" was tripping it up and causing errors (it tried to put files in fontDir/lib/fonts/Helvetica.afm.php instead of fontDir/Helvetica.afm.php). I checked out the code and it turned out that this had to do with it reading $distFonts (dompdf_font_family_cache.dist.php), which used forward slashes. All I did was update it to use DIRECTORY_SEPARATOR, which resolved the issue.

```
$dompdf = new Dompdf([
    "tempDir" => $uploadsDir . DIRECTORY_SEPARATOR . "dompdf",
    "fontCache" => $uploadsDir . DIRECTORY_SEPARATOR . "dompdf",
    "fontDir" => $uploadsDir . DIRECTORY_SEPARATOR . "dompdf",
    "logOutputFile" => $uploadsDir . DIRECTORY_SEPARATOR . "dompdf" . DIRECTORY_SEPARATOR . "log.htm",
]);
```

### DIFF
--- a/lib/fonts/dompdf_font_family_cache.dist.php
+++ b/lib/fonts/dompdf_font_family_cache.dist.php
@@ -1,93 +1,93 @@
-<?php return array (
-  'sans-serif' => 
-  array (
-    'normal' => $rootDir . '/lib/fonts/Helvetica',
-    'bold' => $rootDir . '/lib/fonts/Helvetica-Bold',
-    'italic' => $rootDir . '/lib/fonts/Helvetica-Oblique',
-    'bold_italic' => $rootDir . '/lib/fonts/Helvetica-BoldOblique',
-  ),
-  'times' => 
-  array (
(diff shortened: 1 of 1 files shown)


bsweeney:
I wonder if the changes in commit af914c2cdcaea0c4b3e0efed3354f58caecf13ba would address your issue? Still, I don't see any reason not to make the highlighted changes.

BrunoDeBarros:
They might do; I'll have to test it on a Windows machine and see if it works without my changes. I only pulled your commit after having already dealt with this, so I didn't get a chance to test it.

PR #1212, opened 2016-06-07
fix for vertical-align

anton-tiptyuk:
There are known issues with aligning images vertically in blocks. Now it's suggested to align them with margins which might be inconvenient when image size is not known while source html is rendered.

I've tried to fix vertical align for simple case when container block contains single `img` element, works for me.

Expected 'vertical-align' values for image are 'top', 'middle', 'bottom'


bsweeney:
Do you have some sample code to illustrate the improvement?

PR #1227, opened 2016-07-01
Fixed appending pdf extension

core23:
If you pass a filename with a `.pdf` extension to these methods, you got a duplicate `.pdf.pdf` filename.


core23:
Thanks for merging it 👍

PR #1230, opened 2016-07-07
Excluding files from git archive exports

core23:
These files aren't needed in production, so we should remove them from the git archive export.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,4 +11,10 @@
 *.txt   text
 *.svg   text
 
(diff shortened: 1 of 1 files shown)


bsweeney:
Thank you for your addition. This was covered by PR #1215 and added to the develop branch by commit 08b395618a0774f6e8a19272d4d50eb8f9b0e6e0. It will be merged into the master branch when we're ready for the 0.7.1 beta.

PR #1234, opened 2016-07-10
Fix floating elements position

nappa85:
When we have a floating previous sibling, we have to ignore it when calculating the margin, or we'll end up with a double margin.

For example, when using dompdf with the random generated contents of http://gensav.altervista.org/
Before my edit: https://dl.dropboxusercontent.com/u/18969675/dompdf_before.jpg
After my edit: https://dl.dropboxusercontent.com/u/18969675/dompdf_after.jpg

### DIFF
--- a/src/FrameReflower/Block.php
+++ b/src/FrameReflower/Block.php
@@ -143,6 +143,17 @@ class Block extends AbstractFrameReflower
 
         }
 
+        // When previous sibling is floating, we have to ignore it when calculating the margin
+        if($lm > 0) {
+            $prev = $frame->get_prev_sibling();
+            if($prev) {
+                $prev_stype = $prev->get_style();
+                if ($prev_stype->float === "left") {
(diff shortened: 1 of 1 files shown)


nappa85:
This should be expanded for every type of float (maybe only "right" is missing).
I'm not sure the solution is good for consecutive floating elements, it should be relative to parent instead of previous sibling.

bsweeney:
Makes sense. I hadn't yet reviewed it, but I was going to do some testing with various scenarios. Feel free to update with additional commits if you happen to work on this more.

bsweeney:
I looked into this a bit more and decided to try a different adjustment to the code. We were already shifting the line boxes based on float so in commit 81e0416 I modified that code to take into account margins/padding in relation to the floated element.

bsweeney:
Whoops ... had a minor arithmetic error in that last one. Updated in commit 9895c02

PR #1235, opened 2016-07-11
Fix specificity of * and improve origin precedence

insightfuls:
It looks like some theoretically correct code was commented out to avoid precedence problems for "*", which would have been overridden by the UA stylesheet. Now that 542483ae has landed, this code can be uncommented, as it should work properly.

Furthermore, although 542483ae addressed one origin precedence issue, there were still some potential precedence problems which I've endeavoured to address here:
- non-CSS presentational HTML attributes had incorrect precedence, possibly not a problem in practice, but better safe than sorry
- origin precedence offsets were poorly chosen; specificities could potentially overflow into the next origin in theory, though perhaps not in practice, but again, better safe than sorry
- loading CSS from style elements could use the 'last used' origin, rather than the author origin, as it was not set

There don't seem to be any existing tests for this module that I can extend.

### DIFF
--- a/src/Css/Stylesheet.php
+++ b/src/Css/Stylesheet.php
@@ -55,11 +55,28 @@ class Stylesheet
      */
     const ORIG_AUTHOR = 3;
 
+    /*
+     * The highest possible specificity is 0x01000000 (and that is only for author
+     * stylesheets, as it is for inline styles). Origin precedence can be achieved by
+     * adding multiples of 0x10000000 to the actual specificity. Important
+     * declarations are handled in Style; though technically they should be handled
+     * here so that user important declarations can be made to take precedence over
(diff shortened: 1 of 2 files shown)


bsweeney:
Sadly, few tests at all at this point.

insightfuls:
I should have said that I wasn't just poking around the codebase looking for trouble. I investigated as I encountered a problem rendering some HTML, so tracked down the cause, fixed it, and then when rebasing onto the developer branch, found that it was already partially fixed, but still thought it worthwhile to submit the rest of the fix here. Anyway, you know where to find me if you have any questions or need any more work for this. 😄

bsweeney:
No worries, I didn't take any offense. Always (*always*) glad to have community support/input. And thanks for the patch, I've already tagged it for inclusion in the next release.

PR #1243, opened 2016-07-20
Add php-font-lib and php-svg-lib as submodule

easybe:
Run

    git submodule update --init --recursive

or

    git clone --recursive ...

for this to work.


bsweeney:
We used to use submodules, but it causes problems for some installations. We've switched to Composer for dependency management. Or a user can download a packaged release, which includes everything you need.

PR #1247, opened 2016-07-23
Fixes a variable naming issue.

pancakeapp:
The variable is remoteFile, not remote_file.


bsweeney:
I really should be more careful. Thanks!

BrunoDeBarros:
It's all good. :D That was me by the way; wrong GitHub account. :smile: Just picking up on these things little by little. I wish I could help out more, dompdf is the best thing the PHP world has for HTML-to-PDF and well owe you a lot.

PR #1251, opened 2016-08-04
Fixes an issue with Dompdf\Image\Cache::resolve_url().

BrunoDeBarros:
Helpers::getFileContent() returns an array, but during the refactoring this bit of code wasn't updated to handle it.


bsweeney:
Relates to issue #1257. The issue was fixed by commit 74a9b6a.

Apologies, I forgot about this commit when I fixed up the helper usage.

FernandoManoelAlmeida:
The last update ( 74a9b6a ) makes all works fine, thanks

Issue #923, opened 2015-03-25
cannot vertically position an inline-block


md84419:
I'm having problems vertically-aligning an inline-block inside of div.  The inline-block contains several lines of text of various sizes and the height is not known.

The testcase renders correctly in Chrome but does not render correctly as a pdf created by dompdf.

md84419:
Testcase at http://pastebin.com/VPaXZ7ni

md84419:
I am using the version cloned from git on 24th March 2015 14:05 GMT/WET.

Issue #672, opened 2013-06-25
Inline-blocks + vertical-align do not render correctly


mnpenner:
This code:

``` html
<style type="text/css">
code {
    background:black;
    color:white;
    display:inline-block;
    vertical-align:middle
}
</style>
<p>
    Some text <code>A<br />
    B<br />
    C<br />
    D</code> continues afterward.
</p>
```

Does not render correctly:

![selection_022](https://f.cloud.github.com/assets/1112998/703705/0f7c6eca-ddb0-11e2-8a3e-088e7a79318d.png)

I'm having a lot of trouble getting inline-blocks to align nicely in a paragraph of text; DOMPDF doesn't seem to respect the `vertical-align` property. It's particularly bad when you put a normally block-level element in a paragraph, but even when I try with a span, the alignment is off:

![selection_023](https://f.cloud.github.com/assets/1112998/703706/0f8100ac-ddb0-11e2-8d84-dfec2361a401.png)

bsweeney:
No updates, but I think auto-width issues will be addressed as part of #627.

PR #1256, opened 2016-08-18
Avoid applying vertical-align to text within cells

insightfuls:
It seems that currently vertical-align is applied twice to table cells: once, correctly, to align the cell's contents within the table row (in FrameDecorator\TableCell::set_cell_height), and secondly, incorrectly, to align text directly within table cells (in FrameReflower\Block::vertical_align). This change inhibits the second alignment by defaulting it to 'baseline', the same as text elsewhere, so that only inline elements within the table cell are aligned.

You can see the difference with HTML like this:

```
<style> td { vertical-align: top; } </style>
<table><tr><td> different <span>heights</span> </td></tr></table>
```

Looking at the code, I think there are still some deficiencies with the vertical-align implementation (e.g. images are not affected; nested inline elements are not correctly treated). I may look into these issues also at some stage, but this change covers a common case that is worth fixing immediately, in my opinion, and is the high priority change for us.

Thanks again for a great library.

### DIFF
--- a/src/FrameReflower/Block.php
+++ b/src/FrameReflower/Block.php
@@ -12,6 +12,7 @@ use Dompdf\FontMetrics;
 use Dompdf\Frame;
 use Dompdf\FrameDecorator\Block as BlockFrameDecorator;
 use Dompdf\FrameDecorator\Text as TextFrameDecorator;
+use Dompdf\FrameDecorator\TableCell as TableCellFrameDecorator;
 use Dompdf\Exception;
 
 /**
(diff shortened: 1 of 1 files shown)


bsweeney:
Indeed there's still some work to be done related to vertical alignment. Relevant issues that I could find on a quick search were #672 and #923.

bsweeney:
Have you checked out #1212?

insightfuls:
I must admit, I neglected to consult existing issues in this case.

It seems to me the reason #1212 addresses such a specific situation is that it doesn't address the nesting issue I identified. If the nesting issue is addressed, it should generalise well to inline blocks, I think. Some aspects of it may be able to be generalised sooner, too.

Should I perhaps rebase this work on top of #1212, which affects the same part of the codebase, as a first step? Then at least these two fixes can land together, and if I find time to do some further work, the way is open.

bsweeney:
I was mostly just curious

bsweeney:
I was mainly curious what your thoughts on it were since you had taken time to work on a similar issue. I appreciate the feedback (I haven't really looked at either PR in much detail yet). I leave it to you to decide if you want to rebase on top of that PR. If you have the time that would be wonderful.

I actually plan to look at these two PRs in tandem to pull in to the code next, but because of my upcoming schedule I may not be able to give it the kind of time I prefer to give for reviewing PRs for a few weeks.

PR #1275, opened 2016-10-03
fix output generation sample code

racztiborzoltan:


bsweeney:
Not sure why you want to make this change. Though, perhaps, adding a note that it's important to close the connection after streaming the PDF may be worthwhile.

racztiborzoltan:
Sorry! I was a bit hot-headed!
The original sample code can not be executed. I thought the error in the original sample code. But I was wrong! I apologize! 
Have a nice day!

bsweeney:
No worries. Calling `exit()` isn't necessarily a bad idea as it ensures the PHP process ends. There's a relevant issue to that in #1282.

Issue #1292, opened 2016-10-27
misleading error message in html5lib


tmountjr:
Current html5lib (lib/html5lib/InputStream.php:79) throws an error if iconv isn't installed:

`Not implemented, please install mbstring or iconv`

However, looking further up the code, there's a comment section:

```
// We previously had an mbstring implementation here, but that
// implementation is heavily non-conforming, so it's been
// omitted.
```

Seems the library isn't maintained anymore - the last time anyone touched that was 8 years ago (https://github.com/html5lib/html5lib-php/blob/master/library/HTML5/InputStream.php). Is it worth it to update the downstream version of that library to fix the misleading error message?

bsweeney:
I've actually been planning to update to the fork maintained by Masterminds, which receives regular updates.
https://github.com/Masterminds/html5-php/

PR #1293, opened 2016-10-27
fix misleading error message

tmountjr:
Error message references `mbstring` or `iconv` but the library doesn't support `mbstring` anymore, so remove the reference to it.

See #1292.


bsweeney:
Just FYI, in a upcoming release we'll switch to [Masterminds/html5-php](https://github.com/Masterminds/html5-php)

PR #1297, opened 2016-11-01
Allow for Custom PAGE_NUM and PAGE_COUNT

BillTheGoat:
I wanted to be able to offset `PAGE_NUM` and `PAGE_COUNT` to start after cover page, restart on new sections, etc.  Use like:` {PAGE_NUM-1}` or `{PAGE_NUM+12}`. Working for me including with original` {PAGE_NUM}` format.

### DIFF
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -893,8 +893,13 @@ class CPDF implements Canvas
 
                 switch ($_t) {
                     case "text":
-                        $text = str_replace(array("{PAGE_NUM}", "{PAGE_COUNT}"),
-                            array($page_number, $this->_page_count), $text);
+                        $text = preg_replace_callback('/\{PAGE_NUM([\+\-]?\d*)\}/', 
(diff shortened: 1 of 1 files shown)


bsweeney:
I'm not inclined to include this change since the same result can be accomplished using the `page_script` method, albeit with a bit more work in writing the script. The idea would be to support retrieving the page count via CSS.

BillTheGoat:
No problems. I did this because I attempted to use the `page_script` in this way by creating a variable like `$pn = $PAGE_NUM - 1;`, but I wound up with every previous page number written one over the other on each page, even if I reset the variable to null at the end of the `page_script`. It was fast to just change the logic here, but honestly what I really wanted was to embed all this in the footer div using `<script>`, which I never got to work at all.

bsweeney:
If you add text using the `page_text` method the text will appear on all pages. Same if you add text to a PDF object that is then added to all pages. Scripts are only run a single time.

If you want to run a script on each page to generate customized text for that page you can use page-scripts. They're like inline scripts but run on every page instead of a single time.

PR #1310, opened 2016-11-21
Removed added ".pdf" file extension from the Cpdf stream function.

fhp:
I removed this added .pdf extension, as calling `Dompdf::stream("test.pdf")` results in a generated file called "test.pdf.pdf".

The `Dompdf::stream()` function is supposed to be called with file extension, as suggested by the default for the filename parameter:
`Dompsd::stream($filename = 'document.pdf', $options = null)`


fhp:
This seems to be a duplicate of #1227, so I close it.

bsweeney:
Yes, indeed this has been fixed in the develop branch. It will be included in the next release.

PR #1322, opened 2016-12-01
Allow excluding pages for the page_script to be run on

SubodhDahal:
I was looking for a way to run a `page_script` on all pages except the first one but there didn't seem to be any existing way to do it(or maybe I couldn't find it). So this pull request allows to specify the page numbers on which to exclude the `page_script` to be executed.

**Usage example:**
Show the page number on top of the page for every page except the first and third page.

    $pdf->page_script('$font = $fontMetrics->get_font("Arial, Helvetica, sans-serif", "normal");
                $size = 12;

                $page_text = "Page " . $PAGE_NUM . " of " . $PAGE_COUNT;
                $pdf->text(470, 15, $page_text, $font, $size);
        ', 'text/php', array(1, 3));

The page numbers to be excluded are passed in an array as third parameter to the `page_script` function.

### DIFF
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -911,7 +911,7 @@ class CPDF implements Canvas
             foreach ($this->_page_text as $pt) {
                 extract($pt);
 
-                if (in_array($page_number, $except_on)) {
+                if ($except_on && in_array($page_number, $except_on)) {
                     // don't run the script on the page that is to be excluded
                     break;
                 }
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -873,11 +873,13 @@ class CPDF implements Canvas
      *
      * @param string $code the script code
      * @param string $type the language type for script
+     * @param array $except_on the page numbers on which not to run the script
      */
-    function page_script($code, $type = "text/php")
+    function page_script($code, $type = "text/php", $except_on = array())
     {
         $_t = "script";
-        $this->_page_text[] = compact("_t", "code", "type");
+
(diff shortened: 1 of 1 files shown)


bsweeney:
Probably don't need this. You can check the current page inside the page script via `$PAGE_NUM` from within the page script. It might, however, be useful for page text.

SubodhDahal:
I tried adding the condition `$PAGE_NUM > 1` inside the `page_script` function at first but it didn't seem to have any effect. So I had to make changes to the `page_script`.

bsweeney:
Can you post a sample of your script?

SubodhDahal:
It was somewhere along the lines of following, inside the file from where I get the html content:

    <script type="text/php">
        if ( isset($pdf) ) {
            $pdf->page_script('if ($PAGE_NUM > 1) {
                    $font = $fontMetrics->get_font("Arial, Helvetica, sans-serif", "normal");
                    $size = 12;

                    $page_text = "Page " . $PAGE_NUM . " of " . $PAGE_COUNT;
                    $pdf->text(470, 15, $page_text, $font, $size);
                }
            ');
        }
    </script>

bsweeney:
That works fine for me, page 1 was skipped for the header. I should note one problem with that sample: FontMetrics::get_font expects a single, available font name not a CSS font family declaration. So it won't find a matching font. I believe this still works with CPDF, but with PDFLib an error is thrown. You would want that line to be something along the lines of:

```
$font = $fontMetrics->get_font("sans-serif", "normal");
```

SubodhDahal:
Well that worked now. Don't know what was wrong or missing the first time. I guess this pull request is not needed after all.
Thanks for the information regarding the font.

PR #1331, opened 2016-12-16
fix #1330 add intval() to improve PHP 7.1 compatibility

### LINKED ISSUE #1330: PHP 7.1 - arithmetic with invalid strings
Upgrading PHP to 7.1 caused some issues related to "Notices and warnings on arithmetic with invalid strings" (see http://php.net/manual/fr/migration71.other-changes.php )

Some operations are throwing errors, such as :
`$max_y = $frame->get_position("y") + intval($margin_height);`

Adding intval() on both members do the trick, as intval('random_string') returns 0.

I'll submit a PR or my use case.

dsampaolo:


bsweeney:
Also, any opinion on casting to float instead of int (as in #1399)?

steffenweber:
In the method `Style::length_in_pt` there are some more `(float)` casts needed. For some reason, the code for handling the units `pt`, `%`, `rem`, `em`, `in` and `pc` casts to `float` but the code for `px`, `cm`, `mm` and `ex` does not ([example](https://github.com/soumettre/dompdf/blob/v0.7.0/src/Css/Style.php#L515)).

Adding the `(float)` casts fixes for example the following error:
```
A non well formed numeric value encountered in vendor/dompdf/dompdf/src/Css/Style.php:515
```
In my case the root cause for this error was the HTML code `` whose `height` attribute for some reason arrived in `Style::length_in_pt` as `84 px` (notice the whitespace between the number and the unit).

bsweeney:
@steffenweber could be an issue in the attribute translator. I'll be sure to take a look at that in conjunction with this update.

steffenweber:
Yeah, removing the whitespace in [line 36](https://github.com/soumettre/dompdf/blob/v0.7.0/src/Css/AttributeTranslator.php#L36) and [line 39](https://github.com/soumettre/dompdf/blob/v0.7.0/src/Css/AttributeTranslator.php#L39) of `AttributeSelector.php` solves this issue. (But I'm not sure whether that could break something else.)

bsweeney:
I imagine it was added to avoid issues with the sprintf format string. I don't think the change will break anything, but I'll look into it a bit more.

PR #1335, opened 2016-12-21
Fix and a test for removing spaces from the beginning of an inline tag

davidsickmiller:
It looks like a class name didn't get updated when the code was refactored, which results in some whitespace being removed that should not be.


bsweeney:
Good catch! Extra kudos for including a unit test.

bsweeney:
While this fixes a missed comparison, the results are still not quite what we want. The logic can cause the text to render in undesirable ways if there are inline elements.

bsweeney:
I went ahead and made the changes I suggested. No need to do further work. This issue will close once I merge back in to master.

For future reference the preference is to submit PRs on the develop branch.

davidsickmiller:
Thanks!

PR #1353, opened 2017-01-19
Fixed reading file content for e.g. SVGs

frenzelSL:
I have a problem using SVGs in image tags. The verification of files (BM or SVG) uses the getFileContent method in the Helpers class. 
1.) It passes five parameters but the method has only four parameters. The second is wrong and has to be removed. 
2.) If the SVG has more than 26 chars before the svg tag, it couldn't be detected, so the max-length shouldn't be passed to check the whole content. The last two parameters should be removed.

### DIFF
--- a/src/Helpers.php
+++ b/src/Helpers.php
@@ -529,7 +529,7 @@ class Helpers
         $type = isset($types[$type]) ? $types[$type] : null;
 
         if ($width == null || $height == null) {
-            list($data, $headers) = Helpers::getFileContent($filename, null, $context, 0, 26);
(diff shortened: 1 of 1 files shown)


bsweeney:
Reminder to please target pull requests at the develop branch.

bsweeney:
Found one other place where that call was incorrect. I went ahead and fixed it when pulling in this change.

bsweeney:
Pretty inefficient code in there (ours, not yours). Having to get the file twice if it's not parsed by the PHP functions is ... well ... not good. Perhaps we can improve it by using the Image\Cache class instead of direct access.

PR #1356, opened 2017-01-22
Implement support for widows when determining the location of page breaks

davidsickmiller:
I took a shot at implementing the code for CSS widows.  This was previously a "FIX ME" comment in the code.

While testing my new code, I noticed and corrected some issues in how whitespace in the HTML affected the page breaks.  I'm not sure whether or not these problems were introduced by my widows changes.

### DIFF
--- a/src/FrameDecorator/Page.php
+++ b/src/FrameDecorator/Page.php
@@ -357,14 +357,32 @@ class Page extends AbstractFrameDecorator
 
                 // Rule C
                 $block_parent = $frame->find_block_parent();
-                if (count($block_parent->get_line_boxes()) < $frame->get_style()->orphans) {
+                $block_line_boxes = $block_parent->get_line_boxes();
+                $frame_line_number = 0;
+                $i = 1;
+                $first_node_on_line = false;
+                foreach ($block_line_boxes as &$line_box) {
+                    if ($line_box === $frame->get_containing_line()) {
+                        $frame_line_number = $i;
+			if (count($first_node_on_line = $line_box->get_frames()) > 0) {
+			    $first_node_on_line = $line_box->get_frames()[0]->get_node();
+                        }
+                        break;
+                    }
+                    $i++;
+                }
+
+                if ($frame_line_number <= $frame->get_style()->orphans) {
                     Helpers::dompdf_debug("page-break", "orphans");
 
                     return false;
                 }
 
-                // FIXME: Checking widows is tricky without having laid out the
-                // remaining line boxes.  Just ignore it for now...
+                if (count($block_line_boxes) - $frame_line_number + 1 < $frame->get_style()->widows) {
+                    Helpers::dompdf_debug("page-break", "widows");
+
+                    return false;
+                }
 
                 // Rule D
                 $p = $block_parent;
@@ -389,13 +407,6 @@ class Page extends AbstractFrameDecorator
                     // We are the body's first child
                     Helpers::dompdf_debug("page-break", "Body's first child.");
 
-                    return false;
-                }
-
-                // Skip breaks on empty text nodes
-                if ($frame->is_text_node() &&
-                    $frame->get_node()->nodeValue == ""
-                ) {
                     return false;
                 }
 
@@ -546,6 +557,12 @@ class Page extends AbstractFrameDecorator
             }
 
             if ($next = $iter->get_prev_sibling()) {
+                while ($next->is_text_node() && $next->get_node()->nodeValue == "") {
+                    $next = $next->get_prev_sibling();
+                    if (!$next)
+                        break 2;
+                }
+
                 Helpers::dompdf_debug("page-break", "following prev sibling.");
 
                 if ($next->is_table() && !$iter->is_table())
--- a/src/FrameReflower/Block.php
+++ b/src/FrameReflower/Block.php
@@ -792,24 +792,30 @@ class Block extends AbstractFrameReflower
 
         // Set the containing blocks and reflow each child
         foreach ($this->_frame->get_children() as $child) {
-
-            // Bail out if the page is full
-            if ($page->is_full()) {
-                break;
-            }
-
             $child->set_containing_block($cb_x, $cb_y, $w, $cb_h);
 
             $this->process_clear($child);
 
             $child->reflow($this->_frame);
 
-            // Don't add the child to the line if a page break has occurred
-            if ($page->check_page_break($child)) {
-                break;
-            }
-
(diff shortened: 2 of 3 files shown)


bsweeney:
Nice! I'll try to take a look for the next release.

bsweeney:
@davidsickmiller you available to work on this PR. Since we took so long to review it's a bit out of date now. I'm happy to pull it in but if you have time to update that'd be awesome.

Mellthas:
This looks good in general! The whitespace changes alone would be really nice to have: Before, page breaks before empty text nodes were not allowed, but they were still checked; and as they report their line height as the space needed, they could force a page break, which then would continue to search backwards in the tree for an element that allowed a page break (even though it would fit on the current page). With the changes, empty text nodes are not checked for page breaks, circumventing the problem with their non-zero reported height.

src/FrameDecorator/Page.php
```diff
@@ -546,6 +558,12 @@ function check_page_break(Frame $frame)
             }
 
             if ($next = $iter->get_prev_sibling()) {
+                while ($next->is_text_node() && $next->get_node()->nodeValue == "") {
+                    $next = $next->get_prev_sibling();
+                    if (!$next)
+                        break 2;
+                }
+
```
Mellthas, line 566: This breaks with a whitespace node as first node, in which case it does not check for a parent in the following `if` block (because of the `break 2`).

Proposal:
```php
            $next = $iter->get_prev_sibling();
            // Skip empty text nodes
            while ($next && $next->is_text_node() && $next->get_node()->nodeValue === "") {
                $next = $next->get_prev_sibling();
            }

            if ($next) {
```

src/FrameDecorator/Page.php
```diff
@@ -357,14 +357,33 @@ protected function _page_break_allowed(Frame $frame)
 
                 // Rule C
                 $block_parent = $frame->find_block_parent();
-                if (count($block_parent->get_line_boxes()) < $frame->get_style()->orphans) {
+                $block_line_boxes = $block_parent->get_line_boxes();
+                $frame_line_number = 0;
+                $i = 1;
+                $first_node_on_line = false;
```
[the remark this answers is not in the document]
Mellthas: The variable is indeed unused.

src/FrameDecorator/Page.php
```diff
@@ -357,14 +357,33 @@ protected function _page_break_allowed(Frame $frame)
 
                 // Rule C
                 $block_parent = $frame->find_block_parent();
-                if (count($block_parent->get_line_boxes()) < $frame->get_style()->orphans) {
+                $block_line_boxes = $block_parent->get_line_boxes();
+                $frame_line_number = 0;
```
Mellthas, line 361: I think this should be initialized as `$frame_line_number = count($block_line_boxes)`, basically as before.
Reason being that the frame might not have been added to the line boxes yet, because we are currently in the process of reflowing its children. This can definitely happen in the case of inline-block frames (which would become a concern with #2497, when line breaks before inline-block frames are enabled).

tests/Dompdf/Tests/DompdfTest.php
```diff
@@ -98,4 +98,125 @@ public function testSpaceAtStartOfSecondInlineTag()
...
+                        }
+                    }
+                }
+                $callback_count++;
+            }
+        )));
+
+        $html = "<p>";
+        $filler = array();
+        $MAX_FILLER = 47; // Is there a better way to know how many lines fit on a page?
```
Mellthas, line 129: Using a fixed, and smaller page size together with a fixed line height should make this easier to handle.

src/FrameReflower/Block.php
```diff
@@ -792,26 +792,32 @@ function reflow(BlockFrameDecorator $block = null)
...
+                    break;
+                }
+
+                // Bail out if the page is full
+                if ($page->is_full()) {
+                    break;
+                }
+            }
+        }
+
```
Mellthas, line 820: This change is problematic with the current infrastructure: A page break in a child element (when reflowing the child) can now result in siblings before it staying on the original page, even though there is not enough space for them; they are checked after the page break has already occurred and so no page break is possible anymore.

Mellthas:
As the change to the reflow process (see last review comment) is problematic, the PR in the current form is unfortunately less ready than I thought. This requires more thought on the interaction (and order) of reflow and page break.

I have distilled and fixed the changes about skipping empty text nodes into #2506. #2507 is just about fixing the orphans check, which would have been handled by this PR.

src/FrameReflower/Block.php
Mellthas: To illustrate that see the following HTML:

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style type="text/css">
@page {
	size: 400pt 300pt;
	margin: 50pt;
}

body {
	margin: 0;
}

.block {
	height: 200pt;
	background-color: red;
}
</style>
</head>

<body><div class="block"></div><div class="block"></div><div class="block"></div><div class="block"></div><div class="block"><div></div></div></body>

</html>
```

An the before and after rendering:
[swallow-before.pdf](https://github.com/dompdf/dompdf/files/6919548/swallow-before.pdf)
[swallow-after.pdf](https://github.com/dompdf/dompdf/files/6919550/swallow-after.pdf)

src/FrameDecorator/Page.php
bsweeney: I think this makes sense. The line count will always be the current number of rendered lines. And if Dompdf has to walk back through previous siblings it should be able to get a valid line number.

src/FrameDecorator/Page.php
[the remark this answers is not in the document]
bsweeney: Perhaps was meant to be used at some point but never fully developed. I think removing the fist-line logic makes sense since it's not current used. This should be easy enough to re-implement if we need it later.

src/FrameReflower/Block.php
bsweeney: Maybe not quite right but I think something along these lines is what we'll need to do in order to support widows. Otherwise we don't have a valid count of the number of lines.

bsweeney: This also fails if a page break is forced using page-break styling.

For sure to fully support this would require a deep rework of how the document is reflowed, e.g. reflow the content then page it instead of doing both at the same time, maybe followed by additional steps. Changing the process this way might allow us to implement some functionality that we might have trouble with otherwise. This would be a major re-work, though, so best to pursue in a future release.

Issue #1340, opened 2017-01-04
Tables with missing tbody tag gets one added around every tr - Result is much more memory use


simonberger:
I use Dompdf since a long time in many projects. Thanks for it. :)

Since now I never payed attention to Dompdfs options. But now I had a major memory usage problems.
One pdf listed 400 articles in a table (normally it were many less) that turned out to render 19 pages. Memory consumption was between 80 to 140 MB just for the rendering.
It was a simple pdf with no graphics or anything. I tried to debug a little bit, removed some CSS and at the end looked at the options.

After changing the option I could render the PDF with isHtml5ParserEnabled = true on every server without hitting the 128MB limit.
Using memory_get_peak_usage after the rendering showed 50MB - 140MB before and with isHtml5ParserEnabled = true showed 2MB (I know that might not be the actual correct value but the difference is still HUGE).

The result looked completely identical so for me that is quiet nice.
I write this issue because it might be a bug in case of isHtml5ParserEnabled = false or it's just a much better performance rendering in isHtml5ParserEnabled = true and you should push this setting and give a note that might be a try for everyone having problems with the memory consumption.

Every server runs Dompdf Version 0.7.0 on PHP 7.0

bsweeney:
This isn't the effect I would expect. All that setting does is pre-filter the HTML document through the HTML5 parser. Otherwise the document is parsed in the same way as it is if the HTML5 parser is not used.

It could be that the clean-up the HTML5 parser did on your document significantly affected how it was represented when parsed. Do you have a sample of one of the documents?

simonberger:
Ok here you go. It's the output of the 19 page pdf mentioned above and generated from a laravel template.
I took that output and anonymized it here and there. But tags and so on stayed the same.

I hope you can reproduce the different results I got.

[testexport.zip](https://github.com/dompdf/dompdf/files/685287/testexport.zip)

simonberger:
I started to debug this problem and found a big difference with enabling the following options 

```
$options->set('debugKeepTemp', true);
$options->set('debugLayout', true);
```

With a DEACTIVATED php5 parser the debug-settings draw a red box around every Row (tr) of the table. while there was none with an activated php5 parser. I hope that helps finding the reason.

As a site note. With enabled CSS debugging I got a crash.
I'll open a separated issue for that minor problem.

simonberger:
I got the change of the HTML5 Parser that reduces the memory usage by more than 20% in my case (80mb instead of 110mb).
[...]

PR #1357, opened 2017-01-22
#1340 | Fixing the fix_table method creating unnecessary tbody elemen…

simonberger:
…ts around every single tr


simonberger:
One side effect I noted about that fix.
If you have a buggy input like
```
<table>
  <td>bad placed tag 1</td>
  <p>bad placed tag 2</p>
  <tr>
    <td>row 1 column 1</td>
  </tr>
</table>
```

tbody will be placed also around the `td` and `p` tags.
I wrote about this [here](https://github.com/dompdf/dompdf/issues/1363#issuecomment-275194149)

I don't think that has any worse effect to the old behaviour. Dompdf can't handle that bad input very well and maybe shouldn't anyway. But I didn't test this much.

bsweeney:
Do you think that instead of selecting all children we should select all direct descendent TR elements? It seems unlikely, but what if somebody tries to render a document that contains a table similar to the following:

```
<table>
  <thead>
    ...
  </thead>
  <tr>...</tr>
    ..
</table>
```

It seems unlikely, except that we see all kinds of wacky HTML.

simonberger:
You are correct. That's a case not being handled well. Also some TRs included in a tbody element following some plain rows wouldn't work.
These cases should be at least valid html.

First I wanted to just skip those leading elements resulting in a more complex method but then got a better solution. We already start with the leading tr dom-position having those possible previous siblings.
So just continue from this point walking through the siblings and still append the tbody to the table at the end.

simonberger:
Sorry, my commits sometimes follow a learning process ;)
Thought again and realized the situation addressed in the comment before still isn't fixed if the leading tr followed by a tbody.
```
<table>
		<tr><td>1</td></tr>
		<tbody><tr><td>2</td></tr></tbody>
		<p>missplaced tag</p>
		<tr><td>3</td></tr>
</table>
```

Results now in

```
<table>
		<tbody><tr><td>1</td></tr></tbody>
		<tbody><tr><td>2</td></tr></tbody>
		<p>missplaced tag</p>
		<tbody><tr><td>3</td></tr></tbody>
</table>
```

bsweeney:
I think we're ok to ignore obviously wrong table formatting such as a P as a child of the TABLE element. If someone's HTML is going to be that poorly formed they should use the HTML5 parser. Since a TR is appropriately a child of the table element, but for the purposes of parsing our document we're moving inside a TBODY, I think this is the only exception we want to handle.

Whenever we get around to fully supporting anonymous table objects we can deal with exceptions at that point.

simonberger:
Ignoring the P element was just meant as a "bonus" of the last commit.
The idea was instead to support the (repeated) mix of TR and TBODY direct child elements of a table (the tbody before that missplaced p in the example) beside your correct argument a THEAD followed by TRs would produce wrong restults.

You are free to skip the last two commits but I think the code additions and performance losses are too tiny to not support those unlikely but valid html case.

trevorclarke:
Hi,Not sure if it helps the discussion, at present I regularly use domPDF to process html tables that use thead and tbody. I've not ever used the tfooter tag.tablethead    tr        th/th    /tr/theadtbody    tr        td/td    /tr/tbody/tableThese can and do sometimes have associated css with them.Regards,Trevor Clarke BSc (hons), MCMI Director0780 90 623 68Chittak Ltd Office: 01202 611 711  Skype: chittakwebdevelopment  Website: www.chittak.co.ukRaR Recommended


-------- Original Message --------
Subject: Re:_[dompdf/dompdf]_#1340_|_Fixing_the_fix_table
_method_creating_unnecessary_tbody_elemen…_(#1357)

src/Frame/FrameTree.php
```diff
@@ -160,17 +160,35 @@ protected function fix_tables()
...
             $table->parentNode->insertBefore($caption, $table);
         }
 
-        $rows = $xp->query("//table/tr");
-        foreach ($rows as $row) {
-            $tbody = $this->_dom->createElement("tbody");
-            $tbody = $row->parentNode->insertBefore($tbody, $row);
-            $tbody->appendChild($row);
+        $firstRows = $xp->query('//table/tr[1]');
+        $tbodyInsertPosition = null;
```
bsweeney: Leftover variable?

simonberger: Yay correct, missed this from a previous way.

Issue #1359, opened 2017-01-24
Refactoring of phpdoc and deprecated methods


simonberger:
I went through some classes by hand and with an inspection tool. To summarize what I did for now:

- In 4 classes added and fixed phpdoc in all methods
- Fixed all easy to fix uses of deprecated marked methods (why did one start that job and didn't finish it?)
- Small code fixes in context of phpdoc

The commits can be seen [here](https://github.com/simonberger/dompdf/commits/minorfixes)

I am not one for marking deprecated methods or properties without any comment or possible context why this job can't be finished by now.
 I mainly refer to `Dompdf::pdf`. It seems to be replaced by `Dompdf::canvas` but gets parallely written and returned if Dompdf::canvas would be null. On top of that Dompdf::pdf is used directly instead of Dompdf::canvas or Dompdf::getCanvas. We know it does work because they are sharing the same object reference but I still call it a bug.

Would be nice if one could shed some light on this.

Apart from that question please have a short look at the changes and give me a note if you are interested to pull those kind of changes at all and within a narrow time frame (those changes be outdated quiet fast) then I'll continue the work,

bsweeney:
That's all very nice work. Please feel free to add a PR. I'm not sure about timing, but I don't see any reason we wouldn't be able to merge this quickly. Though, you can continue adding to the PR if you have more changes.

As for the question regarding deprecated methods ... it's mostly a matter of time and who did the work. I this it would be preferable for changes like that to propagate all the way through, but it hasn't always been the case. My priority of late has been fixing actual bugs in the rendering process, but I would be happy to see some help with code cleanliness.

simonberger:
Ok. Regarding those deprecated marks in Dompdf class I'll try to do a hard cut and remove all methods I can't find a direct or indirect references to including the pdf property.
We'll then see where that leads to. :)

simonberger:
There is a lot of wrong phpdoc type declaration around the Frame base class that should be the decorated class AbstractFrameDecorator, Block, Page or whatever instead.
I would need some time to understand this inheritance structure good enough to try to fix this phpdoc. Surely at many places methods and properties could be many different types.

Do you guys think that work would be worth it or is it ok right now?
IDEs for example aren't able to resolve many methods that are in AbstractFrameDecorator while phpdoc notes (directly or indirectly through parent call) Frame right now. Or they report the wrong return type.

Edit:
[...]

PR #1360, opened 2017-01-24
Minorfixes

simonberger:
#1359 Refactoring of phpdoc and deprecated methods


simonberger:
There is a lot of wrong phpdoc type declaration around the `Frame` base class that should be the decorated class `AbstractFrameDecorator`, `Block`, `Page` or whatever instead.
I would need some time to understand this inheritance structure good enough to try to fix this phpdoc. Surely at many places methods and properties could be many different types.

Do you guys think that work would be worth it or is it ok right now?
IDEs for example aren't able to resolve many methods that are in `AbstractFrameDecorator` while phpdoc notes (directly or indirectly through parent call) Frame right now. Or they report the wrong return type.

bsweeney:
This is all just amazing. Thank you for putting in so much work. I was planning to look at pulling this tonight.

bsweeney:
Everything looks good. If you plan to add more commits please go ahead. Otherwise I'll pull this in.

simonberger:
I plan to go through the rest of the classes and fix all the easy, obvious problems like I did in the other files.
Should be finished today or the weekend. Depends on the time I can find.

Later I maybe try fixing the more complicated problems in context of Frame inheriting and classes.
But that would be done in a separate branch/pull request.
If you have timings regarding a 0.7.1 release I would be fine if you merge that at any time and I also do the rest work mentioned at start of this post in a new pull request.

MatthiasKuehneEllerhold:
You could use the [PHP-CBF](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Fixing-Errors-Automatically) to automate fixing a lot of coding style errors.

simonberger:
Thanks for your note Matthias !
I did not know that tool, but from what I see on the website it doesn't deal with most sort of problems/bugs I found here - beside missing phpdoc for sure.

I'm doing most of the semi manual stuff intentionally though. I walk through all classes and learn much of the DomPDF structure to do more complex stuff later, while one-press fixing the phpdoc of methods it's missing on.
Most other problems I am fixing are just shown by the IDE (missing return value, phpdoc violation, unknown methods...)

Removing some still existing psr style violations at the other hand is a bit annoying. That has been done some time ago I think. But many still or again exist. PhpStorm has a good automatic fix for that. But I do not like the magic. On some code it just does too much and adds more line breaks you would need.

MatthiasKuehneEllerhold:
What I like to do with CBF is I'll let it fix everything and then review everything its done. There are some special cases where it misbehaves and you can easily revert them (via git) but it takes away a lot of pain.

It was just a suggestion to further your cause, not intended as a replacement ;)

simonberger:
I'll definitely try it out.
I like tools helping me to realize crappy mistakes I do. ;)

simonberger:
I finished this task and covered all files.
As described in the associated issue #1359 I plan to do more complicated fixes in a new pull request later.

bsweeney:
I should have mentioned this before. But as you're digging back in don't bother with the html5lib code. Eventually we'll be tossing that for a newer version that's actively maintained.

Issue #1365, opened 2017-01-27
image and line functions aren't working on first page of a pdf


BramG:
I had to upgrade dompdf due to server changes.
At this moment it 'kinda works' again but the logo and some stylish line isn't appearing anymore on the first page. I'm using the latest version (0.7.0)

To summarise:
- **pageText** is working and printed on each screen
- **line** is working but NOT printed on the first screen
- **image** is working but NOT printed on the first screen

I also tried to use 'odd' as $where parameter but then nothing gets printed.
Any help would be appreciated.

```
  $pdf = new \Dompdf\Dompdf();

  $options = $pdf->getOptions();
  $options->set('isRemoteEnabled', TRUE);
  $options->set('isPhpEnabled', TRUE);
  $context = stream_context_create([
    'ssl' => [
      'verify_peer' => FALSE,
      'verify_peer_name' => FALSE,
      'allow_self_signed'=> TRUE
    ]
  ]);
  $pdf->setHttpContext($context);

  $pdf->set_base_path(realpath(path_to_theme() . '/'));
  $pdf->load_html($html);

  $pdf->render();

  $canvas = $pdf->getCanvas();

  $font_metric = new \Dompdf\FontMetrics($canvas, $options);
  $font = $font_metric->get_font('helvetica');
  $fontBold = $font_metric->get_font('helvetica', 'bold');

  // Header
  $path = drupal_get_path('theme', 'spock');
  $image = file_create_url($path . '/img/logo-pdf.png');

  $header = $canvas->open_object();
  $canvas->image($image, 472, 35, 85, 25);
  $canvas->line(0, 71, 458, 71, array(0, 0, 0), 5);
  $canvas->page_text(473, 66, t('specifications sheet'), $fontBold, 8, array(0, 0, 0));
  $canvas->close_object();
  $canvas->add_object($header, 'all');

  // Footer
  $footer = $canvas->open_object();
  $date = date('d-m-Y H:i');
  $canvas->page_text(35, 750, t($date . '   |   page {PAGE_NUM}/{PAGE_COUNT}'), $font, 7, array(0, 0, 0));
  $canvas->close_object();
  $canvas->add_object($footer, 'all');

  $pdf->stream($kr_number . '.pdf');
```

simonberger:
Does something change if you remove writing the page_text and/or doing it before the line/image call?

BramG:
Same result, if the pdf is only one page high it works perfectly
I simplified the example.

[...]

PR #1374, opened 2017-02-04
Just init canvas once without a different declared papersize on css side

simonberger:


simonberger:
#1365

bsweeney:
What if the user sets the paper size using the setPaper method instead of using CSS?

simonberger:
Indeed. We need to internally track an option change between the instantiation and the render call. There is no arguable reason to do this but right, its possible and thats enough.

simonberger:
Oh wow now I see what you exactly point to.

We can mess with paper-setup in:

1. The options. That just been read at the constructor and any later change would have no effect at the moment anyway. We stay with that?
2. The above setting is overwritable in the Dompdf object (why ??)
3. And this again is overwritable in the CSS.

So instead of compare just the option paper size to the one in the CSS we need to compare those three in different circumstances and also the option orientation to the Dompdf-orientation.

I did this but to this point didn't test all cases. I'll do this if I have some more time.

In this step I fixed two related problems.
In case a paper size has been found in the styles it sets orientation always to portrait through the call of setPaper.
Also getPaperSize would return null if I didn't miss something.

src/Dompdf.php
```diff
@@ -727,24 +728,38 @@ public function render()
...
 
+        $defaultOptionPaperSize = $this->getPaperSize($options->getDefaultPaperSize());
+        // If there is a CSS defined paper size compare to the paper size used to create the canvas to determine a
+        // recreation need
         if (is_array($basePageStyle->size)) {
-            $this->setPaper(array(0, 0, $basePageStyle->size[0], $basePageStyle->size[1]));
+            $basePageStyleSize = $basePageStyle->size;
+            if ($defaultOptionPaperSize[0] !== $basePageStyleSize[0] || $defaultOptionPaperSize[1] !== $basePageStyleSize[1]
+                || $this->paperOrientation !== $options->getDefaultPaperOrientation()) {
+                $this->setPaper(array(0, 0, $basePageStyleSize[0], $basePageStyleSize[1]), $this->paperOrientation);
```
bsweeney: I'm thinking we don't want to use the previously-defined paper orientation here. In the [code that parses the CSS-defined page size](https://github.com/dompdf/dompdf/blob/v0.7.0/src/Css/Style.php#L2238) orientation is already taken into account. Since the purpose of orientation right now is merely to switch the height/width during Canvas construction, and considering that switch has already happened in the style code, we should probably just set it to portrait.

src/Dompdf.php
```diff
@@ -778,7 +793,7 @@ public function render()
                     $canvas->register_string_subset($style->font_family, $chars);
 
                     // the raw text of the content property
-                    $canvas->register_string_subset($style->font_family, $style->content);
+                    $this->getCanvas()->register_string_subset($style->font_family, $style->content);
```
bsweeney: How come you're not using the $canvas variable already populated with the current canvas?

bsweeney:
FYI, one reason we have the setPaper method is because in some earlier versions you couldn't set the default paper orientation during instantiation since configuration was done via constants. It's not an issue anymore, but I don't see the harm in maintaining the method. I think in the next release we'll fix up how we deal with page size.

simonberger:
That was no intentional change. I created that feature before you merged my refactoring feature into develop and I merged develop two times into this branch. In one of these I had a conflict and did no perfect job it seems.

Regarding orientation: I know what you mean but still don't really understand/know the functionality behind. In case of no CSS page size change the second canvas creation would always be done with landscape too (if initially chosen).
But if the orientation value simply doesn't matter at that point feel free to to remove both checks and also the parameter passing to setPaper.

It's also no problem at all if you decide to not merge this pull request directly and do a commit instead. The first commits in my branch went in another direction and just confuse while interpreting the git history.

bsweeney:
No worries. You're code looks great and I appreciate the time you've put into it.

PR #1378, opened 2017-02-08
A non-numeric value encountered - PHP 7.1 

webaction:
I experienced an issue using PHP 7.1 and Bootstrap 3.

Similar issues have been reported issue number [1272](https://github.com/dompdf/dompdf/issues/1272)


bsweeney:
Nice find! I'm surprised I hadn't run into that yet Could you use `(float)$w` instead of `floatval($w)`?

Issue #1423, opened 2017-04-10
Multithreading / race condition when parsing remote fonts


consolibyte:
When pulling in remote fonts, DomPDF right now always uses a temporary file based on the `md5` hash of the remote filename:

* https://github.com/dompdf/dompdf/blob/master/src/FontMetrics.php#L183

I think this leads to a race condition, and the following error: 

`unlink(/tmp//d0914a6f5f9002d1ec6148e4cf9ab7b0): No such file or directory] on line ...`

Consider two HTTP requests that are executing at the same time: 

* Request 1 comes in and gets to this line: https://github.com/dompdf/dompdf/blob/master/src/FontMetrics.php#L183
* Request 2 comes in and recreates the same file, by the same line: 
* Request 1 gets to this line, and DELETES the file: https://github.com/dompdf/dompdf/blob/master/src/FontMetrics.php#L210
* Request 2 tries to delete the file again, BUT now the file DOES NOT EXIST

bsweeney:
Yes, it would make more sense to use a `tempnam` for the temporary file and add some extra checks before copying the file into the font directory.

marrch-caat:
@bsweeney Please consider reviewing the pull request #1466 
We face the problem in our production, and since it's a quite highloaded project - we had to do something to the problem, so I created the patch. It's my first pull request to a vendor library, so I ask you to be patient if I did something wrong :-\

bsweeney:
@marrch-caat you did great. Even remembered to submit to the develop branch, bravo.

PR #1466, opened 2017-06-02
Fixing issue #1423: support concurrent loading of the same font in FontMetrics::registerFont()

marrch-caat:
Small patch that prevents collisions when several processes place a font to the same temporary file and then one of them deletes it.
Also some minor code optimization made to the only affected function.

### DIFF
--- a/src/FontMetrics.php
+++ b/src/FontMetrics.php
@@ -180,7 +180,7 @@ class FontMetrics
         }
 
         $localFile = $fontDir . DIRECTORY_SEPARATOR . md5($remoteFile);
-        $localTempFile = $this->options->get('tempDir') . "/" . md5($remoteFile);
+        $localTempFilePrefix = $this->options->get('tempDir') . "/" . md5($remoteFile);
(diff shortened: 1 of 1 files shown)


bsweeney:
We could probably save some trouble by using `tempnam` instead of the MD5 hash of the remote file when creating the temporary file name. You wouldn't need the `do...while` loop in that case.

Then the main thing we'd want need to try to avoid is writing to the stored font file while another process generating the PDF document. I don't know how much extra logic we want to put in to handle this case.

marrch-caat:
Oh, I didn't think about it. @bsweeney should I do another pull-request with tempnam? Of course using it would be much simpler...

bsweeney:
No worries. I think I just about have it all done. Took a bit longer because of the changes I made that broke your PR.

bsweeney:
I dunno what I did with that last pull request, but your changes are now merged. Sorry you don't show up in the commit history, I prefer to show where things came from.

PR #1476, opened 2017-06-22
Fix PHP 7.1 non-numeric value error

MrSpider:
In encountered ref_size being something like "1234.56pt", when using CSS transforms on an image.
I can provide a sample if you want.

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -544,7 +544,7 @@ class Style
             }
 
             if ($l === "normal") {
-                $ret += $ref_size;
+                $ret += (float)$ref_size;
                 continue;
             }
 
@@ -576,7 +576,7 @@ class Style
             }
 
             if (($i = mb_strpos($l, "%")) !== false) {
-                $ret += (float)mb_substr($l, 0, $i) / 100 * $ref_size;
+                $ret += (float)mb_substr($l, 0, $i) / 100 * (float)$ref_size;
                 continue;
             }
 


bsweeney:
Sample would be good. The ref_size is supposed to only be a float per the phpdoc comments. Might need to make a change elsewhere to fix usage.

MrSpider:
This sample shows the *length_in_pt* method being called with 262.5pt and 112.5pt as *$ref_size*.

php sample https://gist.github.com/MrSpider/b4cede60aa22d03f8660a0ad86113dfc
html only https://gist.github.com/MrSpider/3a6b196192cea7ec8ab52dcc3b239788

ged15:
This fixes the numeric problem indeed. Any way to get this merged and tagged?

bsweeney:
It's one of the last things we need to do for 0.8.1. Just need a bit of time to review internal usage of that method since it expects a float for the second argument.

ged15:
If it helps, I have been using this patch in production for 3 weeks now and it works as expected :) Used to issue non-numeric value errors before.

bsweeney:
I'm not worried about the patch. I just want to make sure we review the Dompdf code `style_in_pt` usage. There is obviously some code that's not calling that method correctly.

I recognize this will address the issue anywhere we are making the call. I just want to make sure we check so that we can catch any potential issues with our own code.

bsweeney:
FYI, in the future please remember to submit PRs to the develop branch.

Issue #1296, opened 2016-10-29
IMagick extension version check does not work as expected


pun-ky:
see https://github.com/avalanche123/Imagine/issues/474

please make some defensive code which will handle that bug, Currently I have to manually hardcode that :/ 

As a workaround, I am just changing:

`$imagickClonable = version_compare(phpversion('imagick'), '3.0.1rc1') > 0;`

to

`$imagickClonable = true; // version_compare(phpversion('imagick'), '3.0.1rc1') > 0;`

bsweeney:
Also relevant is #459.

bsweeney:
The _extension_ removed the clone keyword in version 3.1 (to clarify my statement from #459).

So if I am understanding this right, `phpversion` does not always return the correct value (?). It looks like we should check the extension version using the `IMAGICK_EXTVER` contstant (which is available starting with 2.0.1 so ... should be safe).

Can you try replacing that line with:

```
$imagickClonable = version_compare(IMAGICK_EXTVER, '3.0.1rc1') > 0;
```

pun-ky:
yes, `phpversion('imagick')` is returning incorrect value due to invalid build of that library distributed on some operating systems

bsweeney:
Do you know if that modification will work as expected?

bsweeney:
The addition of the `IMAGICK_EXTVER` constant was noted in the changelog for [v2.0.1](https://pecl.php.net/package-info.php?package=imagick&version=2.0.1).

The deprecation of the `clone()` method was noted in the changelog for [v3.1.0b1](https://pecl.php.net/package-info.php?package=imagick&version=3.1.0b1).

bsweeney:
The updated check is in the code. If anyone still has issues please post back here.

bluefirex:
Using the bleeding edge version of DomPDF (master-branch) I still get `Imagick::clone method is deprecated and it's use should be avoided`.

PHP 7.0.15
Using Imagick (Extension) on macOS (`imagick module version => @PACKAGE_VERSION@`)

bsweeney:
😭

bsweeney:
I think maybe we should just check for the existence of the clone method rather than do a version compare. I believe that how you get the extension may determine whether or not the version constant is populated.

[...]

PR #1504, opened 2017-08-07
Dompdf compatibility on HHVM (no Imagick::IMAGICK_EXTVER constant)

alsotop:
When running Dompdf on HHVM, it errors at the point of trying to use Imagick::IMAGICK_EXTVER in order to work out whether or not Imagick is cloneable. I realise in a previous version (see #1296) the change was made to go from phpversion('imagick') which provides the version of library and not the extension, but in the case where this constant is not available, can we fallback to the previous method anyway?

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -4790,7 +4790,12 @@ EOT;
             // the first version containing it was 3.0.1RC1
             static $imagickClonable = null;
             if ($imagickClonable === null) {
-                $imagickClonable = version_compare(Imagick::IMAGICK_EXTVER, '3.0.1rc1') > 0;
+                if (defined('Imagick::IMAGICK_EXTVER')) {
+                    $imagickVersion = Imagick::IMAGICK_EXTVER;
+                } else {
+                    $imagickVersion = phpversion('imagick');
+                }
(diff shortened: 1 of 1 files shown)


bsweeney:
I don't know if this is what we want to do. I think adding the check for the constant is a good idea, but if the constant doesn't exist we still don't want to check the extension version since that doesn't tell us what we need to know.

Perhaps what we should do if the constant doesn't exist is check the existence of the clone method? Something like:

```
if ($imagickClonable === null) {
  if (defined('Imagick::IMAGICK_EXTVER')) {
    $imagickClonable = version_compare(Imagick::IMAGICK_EXTVER, '3.0.1rc1') > 0;
  } else {
    $imagickClonable = !method_exists('\Imagick', 'clone');
  }
}
```

(I haven't tested, do we need the "\" for the root namespace?)

Thoughts?

naitsirch:
@bsweeny The backslash is not absolutely necessary, but it is more future proof ;-)
Here is an example of a similar case: https://3v4l.org/CZPmA

bsweeney:
Sorry, forgot to merge this. Used similar but slightly different logic in 371a3b3d23e124dc29564d548e3dcc456aa7a7c1.

PR #1505, opened 2017-08-07
Encrypt some missing strings (issue 1502)

simonberger:
Was a bit complicated for me to understand this object hierarchy and all those content blocks.
I think I got it and added two new "dumb" object-functions that more or less just use the output action and get no options passed.
Might be a good start for later parameterization as well.

I use this as a test html for all those unicode font issues
These are mixed samples from other issues.

```
<!DOCTYPE html>
<html>
<head>
<style>
  @font-face {
    src: url("/fonts/DejaVuSans.ttf");
    font-family: "DejaVu Sans";
  }
  body {
    font-family: DejaVu Sans, sans-serif;
  }
</style>
<title>Testpdf 123 אב בג/רמ</title>
</head>
<body>
  <p>
28/07/2017<br />
׳כל<br />
Registered,agency, smith smith<br />
האב האב בג/רמ רובע לוחל תועיסנ חוטיבל יאופר םותיחל עגונב וניתצלמה ןלהל<br />
הווהמ ל״נה יכ עובקל ונירה םיאופרה םיכמסמבו שקבמה תרהצהב ןויע רחאל<br />
םיאבה םימעטהמ הגרחה -ל הניה וניתצלמה ךכיפלו<br />
- הלבגה<br />
General Text -<br />
Sport - -<br />
,הכרבב<br />
Avi 2
</p>
```

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -857,36 +857,8 @@ class Cpdf
                     // transform FPDF to TCPDF (http://tcpdf.sourceforge.net/)
 
                     $toUnicodeId = ++$this->numObj;
-                    $this->o_contents($toUnicodeId, 'new', 'raw');
+                    $this->o_toUnicode($toUnicodeId, 'new');
                     $this->objects[$id]['info']['toUnicode'] = $toUnicodeId;
-
-                    $stream = <<<EOT
-/CIDInit /ProcSet findresource begin
-12 dict begin
-begincmap
-/CIDSystemInfo
-<</Registry (Adobe)
-/Ordering (UCS)
-/Supplement 0
->> def
-/CMapName /Adobe-Identity-UCS def
-/CMapType 2 def
-1 begincodespacerange
-<0000> <FFFF>
-endcodespacerange
-1 beginbfrange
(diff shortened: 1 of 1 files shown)


bsweeney:
I have a question about this change. There's not much change to the PDF itself (just the encrypting of the ordering and registry text). Do we need this change for the next release? Does it improve the ability of PDF viewers to read the PDF at all? Or is this more tightening up of the security profile of the PDF?

simonberger:
Thanks. Those evil forgotten early tests.
I don't know how to handle this review or if you have to approve it.

The change is really just some overhead to manage to encrypt those strings.
On one hand I am pretty sure we can easily live without this PR for a version. It worked for many years.
On the other hand I don't know how to test those two (four) unencrypted features.
The behavior with the unencrypted unicode font blocks was really bad in the majority of the PDF readers and hold hidden for a long time too.

bsweeney:
I'll test later and probably go ahead and pull this in. The only thing I'm not sure about is if we're encrypting the right (or sufficient) parts of these objects. We'll need to review this more in the future, but if nothing breaks I'm OK pulling it in.

PR #1519, opened 2017-08-21
Add decorator infrastructure for plugins

rudiedirkx:
I want to make a few decorators for Dompdf tweaks I often use. To make a decorator, I need a simple interface that Dompdf already implements, so my decorators can implement the same interface. With decorators, you make make very useful Dompdf plugins that interact at any (public) moment in the Dompdf flow.

I've made 2 (no stealing!):

* Pageable - parses multiple `<body>` in input HTML to make multiple PDF and then merge them into one
* Processable - pre-processes input HTML to change simple expressive HTML intro page scripts etc

This PR doesn't actually include the plugins, but it allows for them by creating the infrastructure. All of this is doable without this PR, but not as pretty. I've only included the methods I need, not all public methods, because there are many. Maybe a few more would be more useful for other scenario's.

Demo decorators and current hacky infrastructure at [rudiedirkx/dompdf-demo](https://github.com/rudiedirkx/dompdf-demo/tree/master/src)

Very simple example to make all input documents UTF-8, because typing is such a hassle:

```
// Plugin
class Utf8Dompdf extends DompdfDecorator // implements DompdfInterface
{
  public function loadHtml($html)
  {
    $html = str_replace('<head>', <head><meta http-equiv="Content-Type" content="text/html; charset=utf-8" />', $html);
    $this->pdf->loadHtml($html);
  }
}

// Implementation
$options = new Options();
$dompdf = new Utf8Dompdf(new Dompdf($options));
```

This simple example could of course be much easier by extending `Dompdf`. The benefit from decorators comes from combining them, and adding config and logic that doesn't belong in Dompdf itself..

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -67,7 +67,7 @@ use Dompdf\Css\Stylesheet;
  *
  * @package dompdf
  */
-class Dompdf
+class Dompdf implements DompdfInterface
 {
     /**
      * Version string for dompdf
@@ -1482,6 +1482,14 @@ class Dompdf
     public function getFontMetrics()
     {
         return $this->fontMetrics;
+    }
+
+    /**
+     * @return FontMetrics
+     */
+    public function clone()
+    {
+        return new static($this->getOptions());
     }
 
     /**
--- a/src/DompdfDecorator.php
+++ b/src/DompdfDecorator.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ * @package dompdf
+ * @link    http://dompdf.github.com/
+ * @author  Benj Carson <benjcarson@digitaljunkies.ca>
+ * @license http://www.gnu.org/copyleft/lesser.html GNU Lesser General Public License
+ */
+namespace Dompdf;
+
+abstract class DompdfDecorator implements DompdfInterface
+{
+    /**
+     * @var DompdfInterface
+     */
+    protected $pdf;
+
+    public function __construct( DompdfInterface $pdf ) {
+        $this->pdf = $pdf;
+    }
+
+    public function getOptions() {
+        return $this->pdf->getOptions();
+    }
+
+    public function getDom() {
+        return $this->pdf->getDom();
+    }
+
+    public function render() {
+        $this->pdf->render();
+    }
+
+    public function output( $options = null ) {
+        return $this->pdf->output($options);
+    }
+
+    public function loadHtml( $html, $encoding = 'UTF-8' ) {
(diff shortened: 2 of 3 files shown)


simonberger:
You may not name the method clone

rudiedirkx:
How about `copy()` ?

It might be nice to just use the same Dompdf object twice, but I couldn't get that to work.

rudiedirkx:
(Btw, tests would be much faster with a `composer.lock`.)

rudiedirkx:
Did you see my cool `PagerProcessor`?

```
return implode("\n", [
	'<script type="text/php">$GLOBALS["' . $name . '"] = $pdf->open_object();</script>',
	$content,
	'<script type="text/php">',
	'$pdf->close_object();',
	'$pdf->page_script(\'',
	'	if ((!' . $firstPage . ' || $PAGE_NUM >= ' . $firstPage . ') && (!' . $lastPage . ' || $PAGE_NUM <= ' . $lastPage . ')) {',
	'		$nobj = $pdf->open_object(); $pdf->close_object();',
	'		$pdf->get_cpdf()->objects[$nobj] = $pdf->get_cpdf()->objects[ $GLOBALS["' . $name . '"] ];',
	'		$pdf->get_cpdf()->objects[$nobj]["c"] = strtr($pdf->get_cpdf()->objects[$nobj]["c"], [',
	'			"%n" => $PAGE_NUM,',
	'			"%t" => $PAGE_COUNT,',
	'		]);',
	'		$pdf->add_object($nobj);',
	'	}',
	'\');',
	'</script>',
]);
```

PHP writing HTML writing PHP with a string of PHP with variables. Waaat. It's not pretty, but it makes the input HTML much prettier:

```
<div class="pager">
	<dompdf-pager pages="2-">%n / %t</dompdf-pager>
</div>
```

So that's what this PR can do, indirectly.

bsweeney:
It's an interesting concept. We'll take a closer look as we have time.

rudiedirkx:
Excellent. There are more usages at [rudiedirkx/dompdf-demo](https://github.com/rudiedirkx/dompdf-demo/tree/master/src)

rudiedirkx:
Did ya did ya? I'd very much like to change my patched dompdf to a stable release.

I'm sure many users would be happy to have some template sugar like `<dompdf-pager>`. These `<script>`s are tricky.

bsweeney:
I'm slammed these days so I'll only be working on bug fixes for the next release. Will keep you updated on progress with this issue.

PR #1526, opened 2017-08-30
Update Cpdf.php

inosjack:
Fix: Call to undefined method Cpdf::quadTo() add a function quadTo()


bsweeney:
This looks to be a duplicate #1493, which is already targeted at the develop branch (preferred for pull requests).

Issue #1532, opened 2017-09-03
Filename containing spaces displayed incorrectly when downloading PDF


Mellthas:
When using the `stream()` method with a filename containing spaces and no characters outside ISO-8859-1, the current develop branch version of dompdf will show any spaces as `%20` in the current version of Firefox (Chrome seems to be fine; I did not test any other browsers). This worked with dompdf version 0.7.

**Example:**
`This is a test file.pdf` gets display as `This%20is%20a%20test%20file.pdf`.
`This is a test – file.pdf` works correctly.

Addressed by #1533.

bsweeney:
Also seems to be an issue for Safari (tested with 10.1.2).

PR #1533, opened 2017-09-03
Fix filename with spaces for streamed output

Mellthas:
Follow https://tools.ietf.org/html/rfc6266#appendix-D by not encoding the `filename` parameter. As a consequence, any double quotes need to be stripped, as the RFC also advices against including the backslash character.

Also, use the `iconv` method to get a transliterated fallback filename instead of ?s for characters not in ISO-8859-1.

Addresses #1532.

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -17,7 +17,7 @@
  */
 use FontLib\Font;
 use FontLib\BinaryStream;
-
+use Dompdf\Helpers;
 
 class Cpdf
 {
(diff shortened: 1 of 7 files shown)


Mellthas:
The same code is used in three places, so it might make sense to factor setting the `Content-Disposition` header out. Would it make sense to put a method into the `Helpers` class?

bsweeney:
If we do decide to include the iconv usage we'll also want to update composer.json to reflect that requirement. Transliteration is a nice touch, but is it nice enough to require iconv just to handle the filename?

I agree that filename generation might prove useful as a helper since the code is the same.

Mellthas:
Fair point. As it is only for the fallback filename, which people might never actually see (with missing characters), I think it doesn’t really matter. I have changed iconv back to mb_convert_encoding.

The second commit adds a helper method for building the Content-Disposition header. The Cpdf class (in `lib`) now uses the helper method from the Dompdf namespace. Is that any problem (dompdf depends on CPDF, CPDF depends on dompdf)?

bsweeney:
Ah, I didn't think about that. Ideally we would like to eventually move the Cpdf backend into it's own project. But there's no timeline on that.

If you're up for it I wonder if we should modify the Cpdf adapter stream method to use `Cpdf::output` from the backend and rewrite it to be more like the other adapter methods. And keep the `stream` method code in the Cpdf backend, keeping it a bit more clean in terms of it's dependencies. That does require some code duplication, but its one less dependency to fix up if we decide to move it into it's own project.

Thoughts?

Mellthas:
This sounds like a good idea. I’ll give it a shot in the following days, if I find some time.

bsweeney:
Most excellent. I appreciate your efforts.

Mellthas:
Here we go. `Cpdf::stream` is now basically a copy of `CPDF::stream`, minus the use of the helper method. I also went through all the other Adapter stream and output methods to make them more consistent and more readable (also regarding documentation).

Notes:
* Changed the use of `1` and `0` for the options to `true` resp. `false` in the code; I think this is clearer.
* Fixed two bugs on the way: Compression not being used by default in the `CPDF::stream` method and broken JPEG output/streaming.
* Removed the `Accept-Ranges` option from `Cpdf::stream` (see commit message).
* Factored out the code duplication in the GD adapter.
* Activated the Content-Length header in the PDFLib adapter, as the size is always known.

I have not tested the changes to the PDFLib adapter, as I dont’t have access to PDFLib.

bsweeney:
Very nice. Thank you.

Mellthas:
You’re welcome!

Issue #1546, opened 2017-09-24
Margin collapsing does not take negative margins into account


Mellthas:
See the attached two test cases and their current rendering with dompdf 0.8.1:

* [test-1.zip](https://github.com/dompdf/dompdf/files/1327737/test-1.zip) – [Output 1.pdf](https://github.com/dompdf/dompdf/files/1327746/Output.1.pdf)
* [test-2.zip](https://github.com/dompdf/dompdf/files/1327736/test-2.zip) – [Output 2.pdf](https://github.com/dompdf/dompdf/files/1327747/Output.2.pdf)

If I read the spec (https://www.w3.org/TR/CSS2/box.html#collapsing-margins) concerning negative margins correctly, then the expected rendering would be as follows:

In the first one, the top and bottom margins of the `body` and the contained `p` element should collapse to 0 (`20 - abs(-20)`). There should be no white gaps at the top or bottom of the page.

In the second one, the bottom margin of the first `p` and the top margin of the second `p` should also collapse to 0. Both paragraphs should have no space between them. 

I also verified the expected rendering with Chrome and Firefox (ignoring page breaks).

From the spec:
> When two or more margins collapse, the resulting margin width is the maximum of the collapsing margins' widths. In the case of negative margins, the maximum of the absolute values of the negative adjoining margins is deducted from the maximum of the positive adjoining margins. If there are no positive margins, the maximum of the absolute values of the adjoining margins is deducted from zero.

PR #1547, opened 2017-09-24
Properly take negative margins into account when collapsing margins

Mellthas:
See #1546.

Rendering of the two test cases with the changes applied:
* [Output 1.pdf](https://github.com/dompdf/dompdf/files/1327790/Output.1.pdf)
* [Output 2.pdf](https://github.com/dompdf/dompdf/files/1327789/Output.2.pdf)


bsweeney:
Nicely done.

PR #1558, opened 2017-09-28
Enable autoloading HTML5_Tokenizer

JParkinson1991:
The following allows for autoloading of the HTML5_Tokenizer.

When this package is installed via composer, the composer autoloader leverages a lazy loading technique.
This lazy loading means the class_exists statement never passes.

Setting the autoload flag to **true** (via removing the flag and leveraging default) allows this check to pass and HTML5 HTML templates to be parsed without error.

Without setting the above the following must be done to parse HTML5 tempaltes without exception:

```
<?php

$tokenizer = new HTML5_Tokenizer(array());
$dompdf = new Dompdf();
$dompdf->loadHtmlFile('someHtml5File.html') 
```

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -465,7 +465,7 @@ class Dompdf
         // https://developer.mozilla.org/en/mozilla's_quirks_mode
         $quirksmode = false;
 
-        if ($this->options->isHtml5ParserEnabled() && class_exists("HTML5_Tokenizer", false)) {
+        if ($this->options->isHtml5ParserEnabled() && class_exists("HTML5_Tokenizer")) {
             $tokenizer = new HTML5_Tokenizer($str);
             $tokenizer->parse();
             $doc = $tokenizer->save();


bsweeney:
right 😞

JParkinson1991:
Thanks for merge!

When do we see dompdf-next becoming a release?

bsweeney:
Not sure. I had some fairly large changes I was looking at for the next release but I may put those off to address some smaller issues first.

JParkinson1991:
To be honest i dont think anyone would complain at lots of minor releases that fix issues as they're found, thats not for me to dictate though.

bsweeney:
Oh I'm very sure people would not complain if we produced a new release sooner. If you look at our past release history it's not typically how we've done things. But I'm trying to get to a place where we release more often to address minor issues.

DojoGeekRA:
@JParkinson1991 in the meantime, i'm doing this in my `composer.json`

```json
{
  "autoload": {
    "files": [
      "vendor/dompdf/dompdf/lib/html5lib/Tokenizer.php",
    ]
  }
}
```

This ensures that the statement evaluates to true, therefore, really enabling the HTML5 parser when the option is provided.

Hope this helps more developers in this situation 👍 

/cc @bsweeney

Issue #1560, opened 2017-09-30
Styling body with inline-block causes block-level parent exception


MichaelEugeneYuen:
Applying inline-block to body tag will cause following errors:

```
Fatal error: Uncaught Dompdf\Exception: No block-level parent found. Not good. in D:\server_root\dompdf\dompdf\src\Positioner\Inline.php:45 Stack trace: #0 D:\server_root\dompdf\dompdf\src\FrameDecorator\AbstractFrameDecorator.php(873): Dompdf\Positioner\Inline->position(Object(Dompdf\FrameDecorator\Block)) #1 D:\server_root\dompdf\dompdf\src\FrameReflower\Block.php(806): Dompdf\FrameDecorator\AbstractFrameDecorator->position() #2 D:\server_root\dompdf\dompdf\src\FrameDecorator\AbstractFrameDecorator.php(894): Dompdf\FrameReflower\Block->reflow(NULL) #3 D:\server_root\dompdf\dompdf\src\FrameReflower\Page.php(141): Dompdf\FrameDecorator\AbstractFrameDecorator->reflow() #4 D:\server_root\dompdf\dompdf\src\FrameDecorator\AbstractFrameDecorator.php(894): Dompdf\FrameReflower\Page->reflow(NULL) #5 D:\server_root\dompdf\dompdf\src\Dompdf.php(832): Dompdf\FrameDecorator\AbstractFrameDecorator->reflow() #6 D:\server_root\dompdf\index.php(24): Dompdf\Dompdf->render() #7 {main} thrown in D:\server_root\dompdf\dompdf\src\Positioner\Inline.php on line 45
```

I ended up using a wrapper with inline-block to solve the issue:

```
<style>
.wrapper {
display: inline-block;
width: 100%;
}
</style>

<body>
<div class="wrapper">
<!-- content here-->
</div>
</body>
```

bsweeney:
Just curious, what is the goal of using that kind of styling on the body element?

Because of what dompdf needs to successfully process the document as a temporary work around we'll probably make the default display styling on the body element `!important`. It won't prevent someone from over-riding the styling, but should make it less likely that anyone will encounter this error.

MichaelEugeneYuen:
I was trying to create an html out as close as the pdf output. So, I created a pdf.css like this:

```
* {
	margin: 0;
	box-sizing: border-box;
}

html {
	padding-top: 10px;
	background: #e6e6e6;
	text-align: center;
}

body {
	display: inline-block;
[...]

Issue #1494, opened 2017-07-16
Whitespace inside HTML tag breaks rendering


TobiasKappe:
It looks like adding whitespace between the HTML and HEAD tags breaks the renderer.

To reproduce this, run the following:

    use Dompdf\Dompdf;
    
    $dompdf = new DomPDF();
    
    $dompdf->load_html(<<<HTML
    <html>
    <head>
    </head>
    <body>
    Hello world
    </body>
    </html>
    HTML
    );
    
    $dompdf->render();

On my system (Arch Linux, PHP 7.1.7, DomPDF 0.8.0), I get the following exception:

    Fatal error: Uncaught Dompdf\Exception: No block-level parent found.  Not good. in <path>/dompdf/dompdf/src/Positioner/Inline.php:45
    Stack trace:
    #0 <path>/dompdf/dompdf/src/FrameDecorator/AbstractFrameDecorator.php(872): Dompdf\Positioner\Inline->position(Object(Dompdf\FrameDecorator\Text))
    #1 <path>/dompdf/dompdf/src/FrameReflower/Text.php(368): Dompdf\FrameDecorator\AbstractFrameDecorator->position()
    #2 <path>/dompdf/dompdf/src/FrameDecorator/AbstractFrameDecorator.php(893): Dompdf\FrameReflower\Text->reflow(NULL)
    #3 <path>/dompdf/dompdf/src/FrameReflower/Page.php(141): Dompdf\FrameDecorator\AbstractFrameDecorator->reflow()
    #4 <path>/dompdf/dompdf/src/FrameDecorator/AbstractFrameDecorator.php(893): Dompdf\FrameReflower\Page->reflow(NULL)
    #5 <path>/dompdf/dompdf/src/Dompdf.php(831): Dom in <path>/dompdf/dompdf/src/Positioner/Inline.php on line 45

Removing the whitespace from `<html>` to `<head>`, `</head>` to `<body>` and `</body>` to `</html>` prevents the error.

TobiasKappe:
On a PHP 7.0 installation, this problem does not occur (for me).

simonberger:
That sounds strange.
Did you try that without the heredoc syntax?

Activating the HTML5 parser should fix this either.

TobiasKappe:
Hi Simon,

Using quotes instead of heredoc makes no difference (and I'm not sure why it would).

[...]

Issue #902, opened 2015-02-14
No block-level parent found. Not good.


Steinweber:
Fatal error: Uncaught exception 'DOMPDF_Exception' with message 'No block-level parent found. Not good.' in /var/www/web50/htdocs/system/library/dompdf/include/inline_positioner.cls.php:37

All installaled
PHP 5.0+
MBString extension
DOM extension
GD extension

No Images, just external css/js

bsweeney:
Check your PHP error log. Typically this error is the result of a null value being parsed, which probably means your document content was lost during initial processing.

Steinweber:
new install of dompdf resolved the problem

VerzatileTom:
Hi, dompdf is working on my localhost but im getting this error on live server. what could be the solution for this?

bsweeney:
You should check your inputs to make sure the content variable (if using `$dompdf->load_html()`) isn't being nulled out. Also look in your PHP logs for any errors. Also, perhaps you're missing a PHP extension or an external PHP library (e.g. the HTML 5 parser) on the server?

tomvo:
I had this happen when I tried to call `loadHTML()` twice on the same object.

VerzatileTom:
i'm using $dompdf->loadView()

bsweeney:
`$dompdf->loadView()`? Are you using dompdf with another library? That method isn't part of dompdf.

VerzatileTom:
oh silly me! im so sorry, I thought its the same. Sorry for bothering you. *flies away*

bsweeney:
Ha ha ... we all do it (you *will not* look at my recent github activity). The error is from dompdf, but you might have better luck with the other library. If that doesn't help come on back over and we can hash it out in the support form.

themao:
I'm doubling @tomvo, I also receive this error when using loadHtml() twice on the same object. I need to generate several pdf files from a list of html templates, so there's no way to avoid using this method twice. 

Is there anything that has to be called before loading new html? It looks like it creates new DOMDocument every time, but still it might have some leftovers from the previous 'generations', am I right?

themao:
For googlers like me, it refers to this bug https://github.com/dompdf/dompdf/issues/1056

bsweeney:
I haven't yet spent much time trying to figure out exactly what would need to be changed to support this. I thought maybe the canvas and frame tree would need to be reset. But the canvas is currently set at render time and the frame tree when a document is loaded. So...

[...]

PR #1570, opened 2017-10-09
Fix: Uncaught Dompdf\Exception: No block-level parent found. Not good

epreston:
Addresses issue: Fatal error: Uncaught Dompdf\Exception: No block-level parent found. Not good #1560 

The default DOM processor does not remove #text child nodes in nodes that should not have them.  While the HTML 5 processor does.

This seems to address random issues between versions of `libxml2` and encoding types.  It also explains why some people have seen success `inlining` the HTML sent for rendering.

I might suggest adding `head` elements to the list that are not allowed to have meaningful `#text` child elements ?

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -490,6 +490,16 @@ class Dompdf
             $doc->loadHTML($str);
             $doc->encoding = $encoding;
 
+            // Remove #text children nodes in nodes that shouldn't have
+            $tag_names = array("html", "table", "tbody", "thead", "tfoot", "tr");
+            foreach ($tag_names as $tag_name) {
+                $nodes = $doc->getElementsByTagName($tag_name);
+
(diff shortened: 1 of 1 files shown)


epreston:
This directly addresses #1494 , explains some of the behavoir discussed in #902 .

epreston:
Looks like the `Scrutinizer` setup needs to be looked at.  It failed building the machine before it could review the code.

enumag:
@bsweeney We have also encountered the `No block-level parent found.` error today. This patch helped. Any idea when a new version might be released?

LasseRafn:
Any status on this? Will it be released in a new version? :) 

Good job on fixing it!

defcasts:
I agree can this fix please go into an updated version? I've added it locally, but would like the current version to support this. Great job on the fix as well.

PR #1583, opened 2017-10-27
Images dompdf

VictoriaLa:
Images dompdf


bsweeney:
Thanks for the PR. The size seems a bit excessive.

Issue #1563, opened 2017-10-01
Call to undefined method DOMDocument::hasAttribute


Divxtaman:
**error**
Fatal error: Call to undefined method DOMDocument::hasAttribute() in dompdf/src/Css/Stylesheet.php on line 970
**html template**
```
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/css/bootstrap.css" crossorigin="anonymous">
</head>
<body>
    <div class="container">
      <div class="starter-template">
        <h1>Bootstrap starter template</h1>
      </div>
    </div>
</body>
</html>
```
Does bootstrap works with this library?

Divxtaman:
fixed by removing in bootstrap.css
```
.custom-file-control:lang(en)::before {
  content: "Browse";
}
```

bsweeney:
Dompdf does work with Bootstrap ... conceptually. Something we changed with 0.8.1 seems to have broken rendering with Bootstrap 4. Until 0.8.2 you can use 0.8.0.

Relevant to #1551.

Issue #1551, opened 2017-09-25
Call to a member function createElement() on null in Stylesheet.php


WeeHong:
Previously I managed to create a PDF without any issue.

But after I ran composer update and this issue.

May I know how to rectify this?

bsweeney:
Composer updated to 0.8.1? I haven't run across this in my testing. Do you have a sample of the HTML + CSS?

WeeHong:
@bsweeney 

I managed to compile my HTML + CSS into CSS by removing the hyperlink of bootstrap.

```
<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/css/bootstrap.min.css" integrity="sha384-/Y6pD6FV/Vv2HJnA6t+vslU6fwYXjCFtcEpHbNJ0lyAFsXTsjBbfaDjzALeQsN6M" crossorigin="anonymous">

<script src="https://code.jquery.com/jquery-3.2.1.slim.min.js" integrity="sha384-KJ3o2DKtIkvYIK3UENzmM7KCkRr/rE9/Qpg6aAZGJwFDMVNA/GpGFF93hXpG5KkN" crossorigin="anonymous"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.11.0/umd/popper.min.js" integrity="sha384-b/U6ypiBEHpOf/4+1nzFpr53nxSS+GLCkfwBdFNTxtclqqenISfwAzpKaMNFNmj4" crossorigin="anonymous"></script>
<script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/js/bootstrap.min.js" integrity="sha384-h0AbiXch4ZDo7tp9hKZ4TsHbi047NrKGLO3SEJAg45jXxnGIfYzk4Si90RDIqNm1" crossorigin="anonymous"></script>
```

But I believe this is not the main issue since I managed to convert my HTML to PDF a few days ago.

usmanbarefoot:
i installed this in my project into two different branches and yesterday i just run the composer update command then after that it's will shown an error =>  Call to undefined method DOMDocument::hasAttribute() #393
and on other branch it's still working like charm

WeeHong:
@usmanbarefoot 

Did my solution help you rectify your issue?

bsweeney:
[...]

PR #1584, opened 2017-11-01
Fixed createElement, hasAttribute issues

denisdulici:
Refs https://github.com/dompdf/dompdf/issues/1551, https://github.com/dompdf/dompdf/issues/1563

### DIFF
--- a/src/Css/Stylesheet.php
+++ b/src/Css/Stylesheet.php
@@ -9,6 +9,7 @@
  */
 namespace Dompdf\Css;
 
+use DOMElement;
 use DOMXPath;
 use Dompdf\Dompdf;
 use Dompdf\Helpers;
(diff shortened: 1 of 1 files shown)


src/Css/Stylesheet.php
```diff
@@ -965,6 +966,11 @@ function apply_styles(FrameTree $tree)
 
                 /** @var \DOMElement $node */
                 foreach ($nodes as $node) {
+                    // Expect $node to be an instance of DOMElement class
+                    if (!$node instanceof DOMElement) {
```
bsweeney: Shouldn't we wrap the statement in parenthesis (`if (!($node instanceof DOMElement)) {...}`?

denisdulici: Not really, as far as I know. It works fine for me and @akaunting users.

bsweeney: You're right, in the [operator precedence documentation](http://php.net/manual/en/language.operators.precedence.php) `instanceof` comes before the negation operator. I guess not having the forced precedence makes me a bit uncomfortable. However, as you stated, not required.

bsweeney:
After looking at why this error was happening I noticed we were already performing this check in the non-generated content style logic block, but with a slightly different condition. I modified the code to use that condition so that the two match up (a76c706).

Issue #1559, opened 2017-09-29
External links does not work anymore


jonnyxe:
PHP 7.1
Dompdf 0.8.1

bsweeney:
This does appear to be the case. Probably to do with string encoding.

hultberg:
@bsweeney I'm having the same problem. After digging deeper into whats going on, I noticed that the URL gets a BOM character after (in CPDF) Cpdf::filterText is called with $bom=true, that then calls Cpdf::utf8toUtf16BE() https://github.com/dompdf/dompdf/blob/v0.8.1/lib/Cpdf.php#L4017

Is an acceptable fix just setting $bom=false here: https://github.com/dompdf/dompdf/blob/v0.8.1/lib/Cpdf.php#L1480 ?

bsweeney:
This is the solution I was thinking was appropriate.

Julianoe:
Has this been resolved ? I seem to have the same kind of problem

gemal:
what is missing here? if a pull request is made will it be accepted and merged?

bsweeney:
@Julianoe it has not yet been resolved. The fix is fairly straightforward but I've got other responsibilities that are consuming any time I might spend on it.

@gemal yes if someone submits a PR that addresses the issue I will accept it.

hultberg:
Both @gemal and I have now opend a PR @bsweeney, closing mine as I was too late. 🙇‍

ray-kay:
@bsweeney so this will be fixed in the next version (0.8.2)? Any idea when this release will be available?

dynasource:
+1

luukScheur:
Anyone else still having this issue? I tested it in 0.8.0, 0.8.1 and 0.8.2, all of them are not working. It seems that I can click (a little) below the link and it works fine, but clicking on the text itself does not work. PHP 7.0

apcro:
I'm having this issue, and can confirm I am using 0.8.2 with the change to `$bom=false`. Any text that looks like a URI is linked, any text that has an <a> tag is not linked.

seanmadrid:
> I'm having this issue, and can confirm I am using 0.8.2 with the change to `$bom=false`. Any text that looks like a URI is linked, any text that has an `<a>` tag is not linked.
> 
> On further investigation: this appears to be a new bug, as it affects 0.8.2 and 0.8.4 as well, with variable results depending on the structure of the `<a>` tag being rendered.

[...]

PR #1587, opened 2017-11-02
Fix external links (#1559)

hultberg:
See #1559 and https://github.com/dompdf/dompdf/issues/1559#issuecomment-338162458


hultberg:
Closing in favour of #1586

PR #1616, opened 2017-12-11
Some improvements to page-break logic

### LINKED ISSUE #1603: Page break handling inside tables incorrect
It seems that the new improved page break handling inside tables (since version 0.8.1) has some bugs. Sometimes a row breaks to the next page (let's say page 2) although there is enough space left on the page (page 1). And the rest of the new page (page 2) is blank; the next rows of the table appear on page 3.

Perhaps the available/remaining space of the page is not calulated properly? If I change the code of FrameDecorator/Page.php back to version 0.8.0 the „old“ handling is correct and better here …

I use dompdf 0.8.1 with pdflib adapter.

Mellthas:
This is my take on trying to fix #1603. Several changes here (all in `FrameDecorator/Page.php`):

1. Simplified the beginning of the `check_page_break` function. As far as I can see, there does not need to be an explicit check for table-row height to prevent infinite loops (the check could also prevent a page break before the first row of a table row group). This should now be sufficiently handled by the `already_pushed` flag.
2. While trying to find a valid break point for a frame, the search now stops if it encounters an already pushed frame (only checks descendants of that frame). This fixes a bug where dompdf was trying to insert a page break on a previous page. The `break-on-previous-page-bug` test case (see below) demonstrates the bug.
3. Page breaks before tables were effectively always forbidden (because the `in_table` flag was already 1 on the level of the table element itself).

### Test cases

Some test cases to demonstrate the behavior changes are included (page size 400pt×300pt):
[Test cases.zip](https://github.com/dompdf/dompdf/files/1548305/Test.cases.zip)

#### table-empty-page-bug
I tried to create a reduced test case for bug #1603. With dompdf 0.8.2, the second page contains only one row even though the remaining ones would fit on the page. For comparison, dompdf 0.8.0 fits one row to much onto the first page. The proposed changes fix the page breaks.

#### infinite-loop
Creates an infinite loop with dompdf 0.8.0, works and renders identical for later versions and with the proposed changes.

#### break-on-previous-page-bug
Produces only two pages without the fix (version 0.8.0 to 0.8.2), the two blue boxes are completely missing. Fixed by the proposed changes.

#### no-break-before-table-bug
With the proposed changes, breaks before a table are possible. Without the changes, a page break would always happen before the first row if there was not enough space on the page. This is more clearly visible by checking the dompdf debug log.

### DIFF
--- a/src/FrameDecorator/Page.php
+++ b/src/FrameDecorator/Page.php
@@ -169,7 +169,6 @@ class Page extends AbstractFrameDecorator
      */
     function check_forced_page_break(Frame $frame)
     {
-
         // Skip check if page is already split
         if ($this->_page_full) {
             return null;
@@ -198,6 +197,7 @@ class Page extends AbstractFrameDecorator
             // $frame->style to the frame's orignal style.
             $frame->get_style()->page_break_before = "auto";
             $this->_page_full = true;
+            $frame->_already_pushed = true;
 
             return true;
         }
@@ -207,6 +207,7 @@ class Page extends AbstractFrameDecorator
             $frame->split(null, true);
             $prev->get_style()->page_break_after = "auto";
             $this->_page_full = true;
+            $frame->_already_pushed = true;
 
             return true;
         }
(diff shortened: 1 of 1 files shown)


michaelarnauts:
Sweet. Works fine! Any reason why this PR is still open?

PR #1621, opened 2017-12-14
Remove ext-gd as dependency

wuethrich44:
On my docker image for development I have only installed ImageMagick to keep the image size small.

With `ext-gd` as requirement in `composer.json` I cannot simply run `composer install` because `gd` is missing on my image.

We can simply remove this requirement because a runtime error is thrown anyways.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,3 @@ matrix:
   allow_failures:
     - php: hhvm
   fast_finish: true
-  include:
-    -
-      php: "5.3"
-      dist: precise
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Follow us on [![Twitter](http://twitter-badges.s3.amazonaws.com/twitter-a.png)](
  
 ## Requirements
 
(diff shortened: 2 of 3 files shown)


rocketlaunchr-cto:
Were you using Linux alpine?:
https://github.com/dompdf/dompdf/issues/1730

fabwu:
@bsweeney Any chance to get this merged?

bsweeney:
I'll need to review this. On the surface it seems reasonable. You don't absolutely need GD, for example if you're not processing any images you don't need any image libraries. But we probably want to make this clear because without handling it appropriately some users would be confused.

fabwu:
@bsweeney I moved the image extensions to the `suggest`-part of `composer.json` and added a check for GD in the `CanvasFactory`. If GD is not present it falls back to `CPDF` and text rendering works without any issue.

As soon as I wanted to render an image `CPDF` throws this exception:

image

I think this message is sufficient so the users shouldn't get confused.

Issue #609, opened 2013-03-21
Underline not rendered correctly.


tic2000:
What steps will reproduce the problem?

Save <a href="http://dev.multibrandingshop.com/aachen.ttf">this font</a>.
Install it in your dompdf.
Try to put some underlined text in a PDF using the new font family.

What is the expected output? What do you see instead?

Expected <a href="http://dev.multibrandingshop.com/correct.pdf">output</a>
Actual <a href="http://dev.multibrandingshop.com/after-17446bc.pdf">output</a>

What version of dompdf are you using? What version of PHP? On what
operating system?
Latest. 5.3. Tried both on linux and Mac.

Please provide the HTML source code you want to convert, or any additional
information.

Any underlined text will do.

I was using an older version of dompdf and I tried to underline that font. The result was a to thin underline and way below the text. But the text was at top as it should have been.

I updated to the latest version and now the underline is OK, but the text is where the old underline was.

The commit that fixed the underline and broke the text is 17446bc.
Looking at the 2 ufm files I noticed that the one with the text in the correct position had
Ascender 857
Descender -143
FontBBox -25 -332 1078 1144

while the wrong one had
Ascender 1144
Descender -332
FontBBox -25 -332 1078 1144

The same day this was commited, March 25th of last year, a change for php-font-lib was done too. Reverting the change for php-font-lib was how I actually generated the correct pdf. Now both the underline is in the right position and has the right thickness and the text is in the right position too.

But I don't know what that could brake either.

inbaz:
Hey, I found this issue and I got nearly the same problem. I've inserted Open Sans from the google fonts into my document with this link tag:

ethaizone:
@inbaz I tried edit .ufm.php file in vendor directory and this work for me too.

My font is Garuda and Initial UnderlinePosition value is -32. I need change it to -1100 for make it look good in my pdf. 

Thank you for point out this info.  It help me a lot. :)

ethaizone:
Maybe this is related issue on https://github.com/PhenX/php-font-lib/issues/7

In this line.
https://github.com/dompdf/dompdf/blob/5113accd9ae5d466077cce5208dcf3fb871bf8f6/src/Renderer/Text.php#L107

Value will divide by 1000 for use in lib. Maybe this is fail from php-font-lib?

ethaizone:
[...]

PR #1631, opened 2018-01-07
Add base to deco_y

ethaizone:
Just stupid try and error and found this work for me.

Ref: https://github.com/dompdf/dompdf/issues/609

### DIFF
--- a/src/Renderer/Text.php
+++ b/src/Renderer/Text.php
@@ -132,7 +132,7 @@ class Text extends AbstractRenderer
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


ethaizone:
I don't except this will get accept but I want to show this code help me in underline, line-through and over line. If we have better solution will be good, I stuck on this bug two day ago. /sob

bsweeney:
Simple enough change to be worth a review for the next release.

bsweeney:
This isn't quite what's wrong, but you put me on to something. I think I have a solution, just need to do some more testing.

PR #1645, opened 2018-01-26
Update page.php to convert value to number

JumpLink:
This fixes a "A non-numeric value encountered" error on php7.1 and should also be compatible with php5


bsweeney:
I'm surprised (though not really) that any of these have gotten past us with the effort to improve PHP 7 compatibility.

bsweeney:
We prefer to use the `(float)` cast for these cases. Can you update?

bsweeney:
preferred change submitted in #1765

PR #1650, opened 2018-02-06
Enable concatenation on css counters with basic text

eddturtle:
When using css counters would have expected `counters(item, ".") "."` to add a trailing full stop. It does this in the browser but not in pdf. Worth looking over code, as I'm not 100%. Happy to make edits/tests.

(Thanks for a great project!)

Example before fix:
https://www.dropbox.com/s/ihqjdzs3tzvmlis/dompdf-counter-before.pdf?dl=0

Example after fix:
https://www.dropbox.com/s/w0qgh79orgl47pp/dompdf-counter-after.pdf?dl=0

Code to recreate:

```
$html = '<!doctype html>
<html>
<head>
    <title>Example</title>
    <style>
        ol {
            counter-reset: item;
        }
        ol li {
            display: block;
            position: relative;
        }
        ol li:before {
            content: counters(item, ".") ".";
            counter-increment: item;
            position: absolute;
            margin-right: 100%;
            margin-left: -35px;
        }
    </style>
</head>
<body>

<ol start="1" style="list-style-type: decimal;">
    <li>one</li>
    <li>two
    <ol>
        <li>Boom</li>
        <li>Boom</li>
        <li>Boom</li>
    </ol>
    </li>
    <li>three</li>
</ol>

</body>
</html>';

$dompdf = new Dompdf();
$dompdf->loadHtml($html);
$dompdf->render();
$dompdf->stream();
```

### DIFF
--- a/src/FrameReflower/AbstractFrameReflower.php
+++ b/src/FrameReflower/AbstractFrameReflower.php
@@ -354,8 +354,8 @@ abstract class AbstractFrameReflower
     {
         // Matches generated content
         $re = "/\n" .
-            "\s(counters?\\([^)]*\\))|\n" .
-            "\A(counters?\\([^)]*\\))|\n" .
+            "\s(counters?\\([^;\v}]*)|\n" .
+            "\A(counters?\\([^;\v}]*)|\n" .
             "\s([\"']) ( (?:[^\"']|\\\\[\"'])+ )(?<!\\\\)\\3|\n" .
             "\A([\"']) ( (?:[^\"']|\\\\[\"'])+ )(?<!\\\\)\\5|\n" .
             "\s([^\s\"']+)|\n" .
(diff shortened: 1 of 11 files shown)


src/FrameReflower/AbstractFrameReflower.php
```diff
@@ -354,8 +354,8 @@ protected function _parse_content()
     {
         // Matches generated content
         $re = "/\n" .
-            "\s(counters?\\([^)]*\\))|\n" .
-            "\A(counters?\\([^)]*\\))|\n" .
+            "\s(counters?\\([^;\v}]*)|\n" .
```
bsweeney: What's the deal with this change? Not quite rocking what you're doing here.

eddturtle: As far as I understand the old regex would capture content starting with 'counters' which was wrapped in brackets (stopping at a closing bracket or new line). The new regex should capture everything until it hits the semi-colon, \v for vertical space or closing brace, so we can pickup on additional text to concatenate.

bsweeney: Thanks, I'll go over it again and let you know if I have any feedback.

bsweeney:
I don't think this one is ready for 0.8.3. When I tested using [this test from the W3C](http://test.csswg.org/suites/css21_dev/nightly-unstable/html4/content-counters-017.htm) the results were not great. Looks like something in the change to the regex causes dompdf to miss some counters in the content style.

Which version of dompdf did you test with? When I tried out your sample on the 0.8.2 release dompdf seems to render correctly.

eddturtle:
I was working based on the 0.8.2 code. That W3C link's great, and you're right. I think it might be best leaving this PR (maybe closing it?) and for a future version working on getting that W3C scenario right.

PR #1671, opened 2018-02-26
Update composer.json

andreybolonin:


casperbakker:
Why? Why do you create a PR without any explanation and without even bother looking at the tests.

You are helping nobody...

bsweeney:
Any reason to limit to PHP 5.6+ except for the obvious that older versions are no longer actively developed? If you think this is a good idea please explain and also add a commit to fix the .travis.yml configuration.

Issue #1702, opened 2018-04-01
[feature] add a page_line() like page_text()


Cyrille37:
Hi,

The method `page_text()` is great for footer but for nicer footer a method `page_line()` will be marvellous :-) 

Cheers.

PR #1703, opened 2018-04-02
[feature] add Method page_line() (#1702)

Cyrille37:
Like page_text(), add method page_line() to able nicer page footer.
implements #1702.

### DIFF
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -562,6 +562,25 @@ class CPDF implements Canvas
     }
 
     /**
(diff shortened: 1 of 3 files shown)


bsweeney:
I'd like to move away from these methods, but maybe worth considering. Especially since the functionality pretty much already exists using either an object or CSS.

Cyrille37:
I agree with
> already exists using either an object or CSS

But I could not make it work ... So really easiest  with page_xxx() methods.

bsweeney:
FYI, in the future please make sure to submit your PRs against the develop branch.

PR #1707, opened 2018-04-06
Don't create temp font files before they are needed

kierate:
In `\Dompdf\FontMetrics::registerFont()` a new `$localTempFile` is created before checking if`$remoteFileContent` is `false`.
This causes empty `dompdf-font-*` files to be left in the temp directory for cases where `$remoteFileContent` is in fact `false`.


kierate:
@bsweeney do you have any rough ETA for when 0.8.3 will be tagged/released?

bsweeney:
No promises, but let's say a month ... ish?

kierate:
Cool. I'll keep an eye out for it.

PR #1732, opened 2018-05-09
Update Stylesheet.php

### LINKED ISSUE #1731: Error on line 175
I have experienced an error on line 175 of src/Css/Stylesheet.php which exists because its being run via CLI

`list($this->_protocol, $this->_base_host, $this->_base_path) = Helpers::explode_url($_SERVER["SCRIPT_FILENAME"]);
`

the problem is $_SERVER["SCRIPT_FILENAME"] is not a valid index when running via CLI as the $_SERVER params are not being assigned, these should be checked prior to being used and have a fallback incase they aren't available, and shouldn't used unless it is set.

using php7 in my instance i've adjusted it to be 

> src/Css/Stylesheet.php

` list($this->_protocol, $this->_base_host, $this->_base_path) = Helpers::explode_url($_SERVER["SCRIPT_FILENAME"] ?? null) ?? 'commandline' ;
`

agentmarine:
Added a check around the server variable for the script file name, incase it is being run via command line. 

fixes #1731

### DIFF
--- a/src/Css/Stylesheet.php
+++ b/src/Css/Stylesheet.php
@@ -172,7 +172,11 @@ class Stylesheet
         $this->setFontMetrics($dompdf->getFontMetrics());
         $this->_styles = array();
         $this->_loaded_files = array();
-        list($this->_protocol, $this->_base_host, $this->_base_path) = Helpers::explode_url($_SERVER["SCRIPT_FILENAME"]);
+        $script = 'commandLine';
+        if(isset($_SERVER["SCRIPT_FILENAME"])){
+            $script = $_SERVER["SCRIPT_FILENAME"];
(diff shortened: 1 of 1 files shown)


bsweeney:
I run from the command line all the time and don't recall seeing this error. And according to http://php.net/manual/en/reserved.variables.server.php:

> If a script is executed with the CLI, as a relative path, such as file.php or ../file.php, $_SERVER['SCRIPT_FILENAME'] will contain the relative path specified by the user.

In which case also not great. Could explain some of the CLI-based resource referencing issues we experience.

I wonder if we can provide a better dummy path. Not really thinking through this but what about defaulting to `__FILE__`?

PR #1740, opened 2018-05-22
Fix undefined method Dompdf\Adapter\PDFLib::_set_ fill_transparency() error

noone-silentrunner:
Wrong method used


noone-silentrunner:
closing because there is more to do.

Issue #1745, opened 2018-05-31
A non-numeric value encountered in FrameDecorator\Page with php 7.2


dvlpp:
I do know that a similar issue was created (#1272). I'm using the latest version of the project (0.8.2), and I'm getting this A non-numeric value encountered NOTICE on `FrameDecorator\Page` line 518. Truth is I can't really figure out when this happens: it has to do with page break in a table, I think, and I can avoid it removing some parts of my CSS (related to cell paddings, page margins), but not always the same parts, depending on the actual data.

If I update line 518:

`$max_y += $p->get_style()->computed_bottom_spacing();`

with:

`$max_y += (float) $p->get_style()->computed_bottom_spacing();`

It *seems to work*, but I'm not sure this is the right thing to do, so I didn't make a PR. Please tell me if I should do so?

Thanks a lot for this package.

bsweeney:
That is absolutely what you should do. Thanks for the find.

PR #1765, opened 2018-06-26
Fix "A non-numeric value encountered in FrameDecorator\Page" #1745

dvlpp:
Avoid error casting as float


bsweeney:
Thanks for the fix. One note, in the future please submit PRs to the develop branch.

PR #1777, opened 2018-07-30
Veryfing if can't get containing line position

RGiordano:
This commit verify if $li->get_containing_line() don't return what is expected. If don't, get Y from frame instead line. [This was crashing with ol tags]

### DIFF
--- a/src/Renderer/ListBullet.php
+++ b/src/Renderer/ListBullet.php
@@ -230,7 +230,11 @@ class ListBullet extends AbstractRenderer
                     $font_family = $style->font_family;
 
                     $line = $li->get_containing_line();
-                    list($x, $y) = array($frame->get_position("x"), $line->y);
+                    if ($line) {
+                        list($x, $y) = array($frame->get_position("x"), $line->y);
+                    } else {
+                        list($x, $y) = $frame->get_position();
+                    }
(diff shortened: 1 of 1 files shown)


bsweeney:
Do you have some sample code showing the issue?

Mellthas:
This can happen with out-of-flow list items, e.g. when applying `float` or `position: absolute`. The following sample exhibits the issue:

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
    size: 400pt 300pt;
    margin: 50pt;
}

li {
    float: left;
}
</style>
</head>

<body>
    <ol>
        <li>A</li>
        <li>B</li>
        <li>C</li>
    </ol>
</body>

</html>
```

Mellthas:
The proposed fix seems reasonable to me, although the list markers render too low in the sample with the fix applied. Using `$li->get_position("y")` instead for the initial y position gives better results in the case of the sample, though the markers are placed a tiny bit too high in that case.

Mellthas:
I have pulled up #2574, so we can resolve this for now. I think the list item's y position is the most appropriate approximation right now. This can still be improved later, if anyone comes around and complains about the position.

Issue #1045, opened 2015-10-21
inconsistent list-style page break issue


antuxity:
dompdf sometime breaks the content of the `<li>` into new page.  It acts like list-style-position:inside.  please check the image, the content should go after "b" similar to "a" and "c"

![Image of issue](https://cloud.githubusercontent.com/assets/2243710/10648315/7df93fb6-77f2-11e5-9f67-c4ef66480970.png)

``` html
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
<html>
   <head>
      <meta http-equiv="Content-Type" content="text/html; charset=iso-8859-15">
      <style>
         .l1 {
            list-style-type: upper-roman;
            list-style-position: outside;
         }
         .l2 {
            list-style-type: upper-alpha;
            list-style-position: outside;
         }
         .l3 {
            list-style-type: decimal;
            list-style-position: outside;
         }
          .l4 {
            list-style-type: lower-alpha;
            list-style-position: outside;
         }
         li{
            margin-bottom: 2em;
         }
      </style>
   </head>
   <body>
      <ul class='l1'>
         <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Donec at
            odio vitae libero tempus convallis. Cum sociis natoque penatibus et
            magnis dis parturient montes, nascetur ridiculus mus. Vestibulum purus
            mauris, dapibus eu, sagittis quis, sagittis quis, mi. Morbi fringilla
            massa quis velit. Curabitur metus massa, semper mollis, molestie vel,
            adipiscing nec, massa. Phasellus vitae felis sed lectus dapibus
            facilisis. In ultrices sagittis ipsum. In at est. Integer iaculis
            turpis vel magna. Cras eu est. Integer porttitor ligula a
            tellus. Curabitur accumsan ipsum a velit. Sed laoreet lectus quis
            leo. Nulla pellentesque molestie ante. Quisque vestibulum est id
            justo. Ut pellentesque ante in neque. 1</li>
             <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Donec at
            odio vitae libero tempus convallis. Cum sociis natoque penatibus et
            magnis dis parturient montes, nascetur ridiculus mus. Vestibulum purus
            mauris, dapibus eu, sagittis quis, sagittis quis, mi. Morbi fringilla
[...]

Issue #872, opened 2014-11-06
Page break list issue 


shidhu40:
https://github.com/dompdf/dompdf

![capture](https://cloud.githubusercontent.com/assets/7380722/4931358/58aef20c-6577-11e4-9965-e413dbda2066.PNG)

bsweeney:
closing, subscribe to 403 for updates

Issue #403, opened 2012-12-06
List item bullet handling not optimal during page split


PhenX:
_Original author: eclecticgeek (December 21, 2011 21:12:47)_

When a list item (LI) falls at the bottom of a page such that all the content in that LI is pushed to the next page the number remains on the preceding page. Ideally when no content remains for that LI on the current page it will be pushed to the next page.

Confirmed in trunk r449.

Sample code attached.

_Original issue: http://code.google.com/p/dompdf/issues/detail?id=397_

bsweeney:
Still a problem. Here's the HTML from the original issue:

```
<html>
<style>
    .block {
        height: 200px;
        width: 200px;
        background-color: orange;
    }
</style>

<body>
    <ol>
        <li>
            <div class="block">X</div>
        </li>
        <li>
            <div class="block">X</div>
        </li>
        <li>
            <div class="block">X</div>
        </li>
        <li>
            <div class="block">X</div>
        </li>
        <li>
            <div class="block">X</div>
        </li>
    </ol>
</body>

</html>
```

bsweeney:
Actually. I think this particular issue has been resolved. However, there appears to be another issue around list bullets after the page break. The sample in 1799 is missing a bullet on the following page in the current release (0.8.5).

bsweeney:
smh ... or not. The sample from 1799 has a list item with a line break but the same text.

Never make assumptions about what you're looking at, folks.

bsweeney:
I believe this issue is sufficiently addressed by the current release. Please post back if anyone is able to determine otherwise.

PR #1798, opened 2018-09-17
Fixed issues with list bullets on page break

naitsirch:
There are several issues mentioning problems with the rendering of bullets when the page is breaked within a list (see #403, #872 and #1045).

When dompdf detects that the content of a list item overflows the page, it clones the item to the next page and moves all following content to it. If the first child of the item leads to the overflow, the bullet is still drawn at the end of the first page and no bullet is drawn for the cloned item on the next page.

In fact there are three issues that I address in this PR:
* The origin list item gets removed if it has no content anymore after it was splitted.
* The cloned element is marked as non-splitted if the origin item was removed, so that the bullet of the cloned item on the next page gets rendered.
* Copy the value of the attribute `dompdf-counter` of the origin bullet element. Because the attribute of the new bullet element that is created for the cloned list item was not filled correctly, so that the bullet was not rendered if an ordered list (`<ol>`) is used.

The fix is a bit hacky, but I couldn't find a cleaner way of fixing the issue. The problem is that the bullet is handled by a custom DOM element which makes some checks more complicated.

Maybe this could be changed sometime.

I hope this PR could be merged, soon. If so please create a new release, so that it gets installed when composer is used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ php:
   - 5.6
   - 7.0
   - 7.1
+  - 7.2
   - hhvm
 
 install: travis_retry composer install --no-interaction --prefer-source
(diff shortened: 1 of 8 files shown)


naitsirch:
I have detected a special case where my former patch failed to detect an empty frame. It failed if there are empty text nodes between the item bullet and the node that is splitted.

With the latest commit (#8709a44) this case is handled correctly.

naitsirch:
Sorry, I have blown up this PR. I'll close it in favor of #1799.

PR #1799, opened 2018-09-18
Fixed issues with list bullets on page break

naitsirch:
There are several issues mentioning problems with the rendering of bullets when the page is breaked within a list (see #403, #872 and #1045).

When dompdf detects that the content of a list item overflows the page, it clones the item to the next page and moves all following content to it. If the first child of the item leads to the overflow, the bullet is still drawn at the end of the first page and no bullet is drawn for the cloned item on the next page.

In fact there are three issues that I address in this PR:
* The origin list item gets removed if it has no content anymore after it was splitted.
* The cloned element is marked as non-splitted if the origin item was removed, so that the bullet of the cloned item on the next page gets rendered.
* Copy the value of the attribute `dompdf-counter` of the origin bullet element. Because the attribute of the new bullet element that is created for the cloned list item was not filled correctly, so that the bullet was not rendered if an ordered list (`<ol>`) is used.

The fix is a bit hacky, but I couldn't find a cleaner way of fixing the issue. The problem is that the bullet is handled by a custom DOM element which makes some checks more complicated.

Maybe this could be changed sometime.

I hope this PR could be merged, soon. If so please create a new release, so that it gets installed when composer is used.

### DIFF
--- a/src/FrameDecorator/AbstractFrameDecorator.php
+++ b/src/FrameDecorator/AbstractFrameDecorator.php
@@ -697,6 +697,13 @@ abstract class AbstractFrameDecorator extends Frame
         $split->_splitted = true;
         $split->_already_pushed = true;
 
+        // Workaround for list item bullets, because the counter of the bullet gets lost,
+        // because node is not cloned deeply.
+        if ('list-item' === $split->get_style()->display && $node instanceof DOMElement) {
+            $counter = $node->childNodes->item(0)->getAttribute('dompdf-counter');
+            $split->get_node()->childNodes->item(0)->setAttribute('dompdf-counter', $counter);
+        }
+
         // The body's properties must be kept
         if ($node->nodeName !== "body") {
             // Style reset on the first and second parts
@@ -745,6 +752,15 @@ abstract class AbstractFrameDecorator extends Frame
         if ($style->counter_reset && ($reset = $style->counter_reset) !== "none") {
             $vars = preg_split('/\s+/', trim($reset), 2);
             $split->_counters['__' . $vars[0]] = $this->lookup_counter_frame($vars[0])->_counters[$vars[0]];
+        }
+
+        // If the frame is completely empty after the splitting, it must be removed from the tree.
+        // This is necessary to avoid drawing list bullets although the list item has been emptied during splitting.
+        if ($this->is_empty()) {
+            $this->get_parent()->remove_child($this);
+
+            // Mark the copy as non-splitted, because we want the bullet to be drawn if the origin was removed.
+            $split->_splitted = false;
         }
     }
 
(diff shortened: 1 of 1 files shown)


bsweeney:
@naitsirch do you have some sample HTML that shows the issue? I attempted to reproduce but didn't have any luck.

I thought we had addressed some of these issues through various updates. And if so the relevant issues should be closed.

naitsirch:
Yes, it is a bit tricky. It only happens if there are child items within the list item, that is wrapped to the next page.
Here's the code to reproduce:
```php
<?php
require 'vendor/autoload.php';

use Dompdf\Dompdf;

$html = <<<HTML
<!DOCTYPE html>
<html>
    <head>
        <style>
            li { font-size: 15pt; font-family: Helvetica, sans-serif; }
            p { margin: 0; padding: 0; }
            #last { color: red; line-height: 1.35; }
        </style>
    </head>
    <body>
        <ul>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li id="last">Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
            <li>
                <p>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</p>
                <p>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</p>
            </li>
            <li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</li>
        </ul>
    </body>
</html>
HTML;

$dompdf = new Dompdf();
$dompdf->loadHtml($html);
$dompdf->setPaper('A5', 'landscape');
$dompdf->render();
$dompdf->stream();
```
And this is a screenshot that shows the problem:
screenshot_20181114_211041

I hope this helps.

naitsirch:
I have tested it with master branch now, too. The issue still exists.

bsweeney:
I was testing with the develop branch and wasn't able to reproduce with your sample. Switching to master I was. I'm going to set aside the PR until I have a chance to review it a bit more since we need to get the next release out.

naitsirch:
Can you attach a screenshot how it looks on your side?

bsweeney:
Things are a tad better now but there are still issues after the page break. And the issue with not flowing over the bullet for overly large LI content is still a problem (per issue 403).

So, we will be looking at this a bit more.

bsweeney:
OK, nvm. I was looking at the input incorrectly. I think we render OK now. I am curious to see if there is any affect on this after apply the widow support from #1356.

bsweeney:
I think this is no longer needed with the current. I'm going to go ahead and close for now. If the issue comes back up I'll revisit the commit at that time.

PR #1800, opened 2018-09-20
Fixed calculation of minimum width of text if 'word-wrap' is 'break-word'

naitsirch:
dompdf already supports the CSS option `word-wrap` with `break-word`. It correctly breaks words if they are too long to fit in their parent's block. But the calculation of the minimum width of the text has not respected this setting. The problem is that their parent elements are rendered wider as they have to - even if a width is specified.

HTML code to test this behaviour:

```html
<table>
    <tr>
        <td style="width: 283pt; border:1pt solid black;">
            <ul style="border:1pt solid green;">
                <li style="word-wrap:break-word;">
                    nfioweifjiewpfijiepwfjiewpfjiepwjfipwejfipwejfipwejfiwejfiewjfiwejfewfjiowefjipowefjiowefjiowefjiowefjioewfjiwefjiowefjiwoefjiowefjiowefjiowefjiowefjiowefjiowefjiowefjioewfwifiowejfowiejf
                </li>
            </ul>
        </td>
        <td>Hello World</td>
    </tr>
</table>
```

Before the patch the above HTML produced this output:
dompdf_before

After:
dompdf_after

### DIFF
--- a/src/FrameReflower/Text.php
+++ b/src/FrameReflower/Text.php
@@ -467,9 +467,18 @@ class Text extends AbstractFrameReflower
             $style->border_right_width,
             $style->margin_right), $line_width);
         $min += $delta;
+        $min_word = $min;
         $max += $delta;
 
-        return $this->_min_max_cache = array($min, $max, "min" => $min, "max" => $max);
+        if ($style->word_wrap === 'break-word') {
+            // If it is allowed to break words, the min width is the widest character.
+            // But for performance reasons, we only check the first character.
+            $char = mb_substr($str, 0, 1);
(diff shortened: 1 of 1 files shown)


bsweeney:
This seems reasonable. It would certainly pose a potential performance issue to parse all characters to find the smallest.

Another option, I'd have to think on it more, and it still wouldn't be perfect, would be to get the bounding box for the font and use that as the minimum. This would require exposing some functionality via the canvas for retrieving this information. And though I know it's available via Cpdf I'm not sure if it is for other renderers.

naitsirch:
I think I cannot help you with this idea. My font and PDF knowledge is to small for that :-)

bsweeney:
heh ... no worries. I'll take whatever help comes my way.

PR #1818, opened 2018-10-17
Dev float cast

Jawelo:
Add (float) cast in order to prevent the error "A non-numeric value encountered" from the FrameDecorator/Page.php file.

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -1390,24 +1390,24 @@ class Style
     {
         return array(
             "top" => array(
-                "width" => $this->__get("border_top_width"),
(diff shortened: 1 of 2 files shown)


bsweeney:
`get_border_properties` should return the CSS property values and not necessarily a numeric value. Particularly for things like "color" or "style" which aren't defined as potentially numeric values.

If you're running into issues because of a missing case it's probably somewhere else in the code.

bsweeney:
Also, if you want to submit a PR please do so against the develop branch. There may be changes there that address issues you encounter and it helps keep the master branch more "stable" (from a release perspective).

PR #1821, opened 2018-10-26
Fix rendering of remote images on PHP 7.1+

### LINKED ISSUE #1781: tempnam(): file created in the system's temporary directory
Hi

when i run this in laravel i get this error 
tempnam(): file created in the system's temporary directory

```
public function generatePDF()
 {
        $students = Student::all();

        $pdf = PDF::loadView('admin.index', compact('students'));

        return $pdf->download('page.pdf');

    }
```

benface:
Fixes #1781 

Following [a change with the `tempnam()` function in PHP 7.1](https://bugs.php.net/bug.php?id=69489), rendering a PDF with remote images (when they are enabled) is broken. This PR prevents PHP from throwing the notice that breaks it.

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -4801,12 +4801,12 @@ EOT;
         imagesavealpha($img, false);
 
         // create temp alpha file
-        $tempfile_alpha = tempnam($this->tmp, "cpdf_img_");
+        $tempfile_alpha = @tempnam($this->tmp, "cpdf_img_");
         @unlink($tempfile_alpha);
         $tempfile_alpha = "$tempfile_alpha.png";
 
         // create temp plain file
-        $tempfile_plain = tempnam($this->tmp, "cpdf_img_");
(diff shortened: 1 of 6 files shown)


bsweeney:
I'm fine with this change but notices really shouldn't break execution of the code. They're informational in nature and only potentially problematic. From the PHP documentation:

> Run-time notices. Indicate that the script encountered something that could indicate an error, but could also happen in the normal course of running a script.

bsweeney:
Also please note that pull requests should be submitted against the develop branch.

benface:
Oh! Sorry about that, I'll be careful next time, and thank you for the merge. As for why this breaks execution of the code (at least for me in the [context of Craft Commerce](https://github.com/craftcms/commerce/issues/537)), I have no clue but I confirm that I tested this change and it fixes it.

bsweeney:
Some frameworks are a bit more finicky by default in their handling of exceptions, warnings, and notices. I know laravel can be. That's likely the case here as well.

Thanks for the confirmation of the fix.

PR #1830, opened 2018-10-29
Add output_buffering requirement to README

oscar-latorre:


barryvdh:
Why is this required?

bsweeney:
We use it mainly as a means of capturing file contents. And also for logging (bad form, old chap). We should be able to rewrite the lib so that OB is not required, but not for the upcoming release.

PR #1849, opened 2018-11-29
Update VERSION

pmayd:


bsweeney:
The version file was originally intended to be updated programmatically during project export, I'll need to rethink how we're doing this, but the PR is now a version behind so I'm going to close it.

PR #1850, opened 2018-12-02
save as PDF /A

uomopalese:
Wold be nice to save documents as PDF/A for long term storage.


bsweeney:
This doesn't look like a pull request (I don't see any modifications from you). PDF/A support has been requested as an enhancement, see #1106.

Issue #1668, opened 2018-02-25
Strange internal problem


shawe:
It was happening to me using PHP 7.1.14 and 7.2.2, and I change dompdf from 0.8.0 to 0.8.2 to try if something happens different, but seems to be the same problem.

This error occurs to me when call new Dompdf object:
$dompdf = new Dompdf($options);
$dompdf = new Dompdf();

vendor/dompdf/dompdf/src/Adapter/CPDF.php
Line: 216
Message: Uncaught Error: Call to protected method Cpdf::getFirstPageId() from context 'Dompdf\Adapter\CPDF' in vendor/dompdf/dompdf/src/Adapter/CPDF.php:216 Stack trace: #0 vendor/dompdf/dompdf/src/CanvasFactory.php(57): Dompdf\Adapter\CPDF->__construct('letter', 'portrait', Object(Dompdf\Dompdf)) #1 vendor/dompdf/dompdf/src/Dompdf.php(294): Dompdf\CanvasFactory::get_instance(Object(Dompdf\Dompdf), 'letter', 'portrait')...

I change this protected method to public, and the error was change to another line, but still not working. 

vendor/dompdf/dompdf/src/Adapter/CPDF.php
Line: 477
Message: Uncaught Error: Call to undefined method Cpdf::setLineTransparency() in vendor/dompdf/dompdf/src/Adapter/CPDF.php:477 Stack trace: #0 vendor/dompdf/dompdf/src/Adapter/CPDF.php(521): Dompdf\Adapter\CPDF->_set_line_transparency('Normal', 1) #1 vendor/dompdf/dompdf/src/Renderer/AbstractRenderer.php(910): Dompdf\Adapter\CPDF->set_opacity(1) #2 vendor/dompdf/dompdf/src/Renderer/Block.php(32): Dompdf\Renderer\AbstractRenderer->_set_opacity(1) #3 vendor/dompdf/dompdf/src/Renderer.php(293): Dompdf\Renderer\Block->render(Object(Dompdf\FrameDecorator\Block)) #4 vendor/dompdf/dompdf/src/Renderer.php(101): Dompdf\Renderer->_render_frame('block', Object(Dompdf\FrameDecorator\Block))...

Any idea that what can be happening and how to solve it?

Some time ago, I test to generate a PDF in same way, and no problem was happening to me, simply works. And this time was based with same PHP version, and dompdf 0.8.0, I look to same previous code to re-use it now.

hiboutik:
[...]

PR #1851, opened 2018-12-04
added namespace for class Cpdf in lib folder

gozeller:
s. issue #1668: https://github.com/dompdf/dompdf/issues/1668

Added a namespace to lib/Cpdf.php to avoid namespace conflicts, e.g. with rospdf/php-pdf 

Not sure, if it's "allowed" to add "\Dompdf" namespace to Cpdf - is it an old/modified version of rospdf/php-pdf?

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -15,6 +15,9 @@
  * @license Public Domain http://creativecommons.org/licenses/publicdomain/
  * @package Cpdf
  */
+
+namespace Dompdf;
+
 use FontLib\Font;
 use FontLib\BinaryStream;
 
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -4859,7 +4859,7 @@ EOT;
             // the first version containing it was 3.0.1RC1
             static $imagickClonable = null;
             if ($imagickClonable === null) {
-                $imagickClonable = version_compare(Imagick::IMAGICK_EXTVER, '3.0.1rc1') > 0;
+                $imagickClonable = version_compare(\Imagick::IMAGICK_EXTVER, '3.0.1rc1') > 0;
             }
 
             $imagick = new \Imagick($file);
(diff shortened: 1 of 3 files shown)


bsweeney:
It is a modified version of the original R&OS PDF (back when it was being developed by Wayne Monroe, not the iteration currently hosted on github). I was not involved back when it was introduced with the project so I can't make any statements about its provenance beyond that.

shital7mahajan:
I have added that code from and run following command to check php Combability
Command=> phpcs --standard=PHPCompatibility --extensions=php,module,inc,install,test,profile,theme --runtime-set testVersion 7.2 sites/default/libraries/dompdf

Error: Extension 'cpdf' is removed since PHP 5.1; Use pecl/pdflib instead is still there

any update on it?

bsweeney:
@shital7mahajan you changed the three files indicated and are still receiving the error?

shital7mahajan:
@bsweeney yes.
If we run this command phpcs --standard=PHPCompatibility --extensions=php,module,inc,install,test,profile,theme --runtime-set testVersion 7.2 sites/default/libraries/dompdf
It's showing following error

1] 3444|Warning|Since PHP 7.0, functions inspecting arguments, like debug_print_backtrace(), no longer report the original value as passed to a parameter, but will instead
provide the current value. The parameter "$node" was used, and possibly changed (by reference), on line 3444.
2]198 | ERROR | Extension 'cpdf' is removed since PHP 5.1; Use pecl/pdflib instead
3]480|Warning|Since PHP 7.0, functions inspecting arguments, like debug_backtrace(), no longer report the original value as passed to a parameter, but will instead provide thecurrent value. The parameter "$type" was used, and possibly changed (by reference), on line 480.

So we have resolved that error by https://github.com/dompdf/dompdf/pull/1934
and warning is still there can you please help me to resolved this.

PR #1868, opened 2018-12-29
fix undefined index N

lenguyentrang:
Some fonts not have type "N"

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -2584,6 +2584,9 @@ EOT;
                             }
 
                             $c = (int)$dtmp['U'];
+                            
+                            if(!isset($dtmp['N']))
+                                break;
                             $n = $dtmp['N'];
                             $glyph = $dtmp['G'];
                             $width = floatval($dtmp['WX']);


bsweeney:
I haven't run into this issue yet. Do you have a sample font where that property is missing from one of the glyphs?

lib/Cpdf.php
```diff
@@ -2584,6 +2584,9 @@ private function openFont($font)
                             }
 
                             $c = (int)$dtmp['U'];
+                            
+                            if(!isset($dtmp['N']))
```
bsweeney: I think we want to keep processing in this case. Probably a better way to deal with the lack of a value is to pre-define the array since we expect a standardized format. So initialize $dtmp as `array('G' => null, 'N' => null, 'U' => null, 'WX' => null)` above (line 2566).

Then, move the break to the else below (line 2605 in your repo) since the name is only required if the character index is less than zero.

We'll also want to set the codeToName key only if a value exists in $n, so line 2601 should be something like `if ($n != null && $c != hexed($n)) {`

bsweeney: ha ha thanks auto-correct for making `hexdec` into `hexed`

bsweeney: also note that my comment relating to line 2601 was addressed related to #2003

PR #1825, opened 2018-10-28
fixes #1822 - PHP 7.3 Warnings

PHPGangsta:

PR #1889, opened 2019-02-05
Added a missing break in the @page stylesheet parsing

mdriessen:
Since updating to version 0.8.3 my HTML wasn't rendering properly. I traced it back to pull-request #1825 Only the default should break 2, the other cases should only break once.


bsweeney:
This addresses issue #1898, which I worked on before fully reviewing this PR (see cccb56d5150ff1e4b921ef0b087dee645d6866e8). If I had realized this was the fix I would have merged this PR a long time ago (I prefer to give credit where due). On the plus side we came to the same conclusion about the source of the problem.

PR #1892, opened 2019-02-08
Move phenx/php-svg-lib to suggested, not required, packages

fitztrev:
`phenx/php-svg-lib` is only required if you are rendering SVG.


bsweeney:
I'm not sure we want to make it optional. As of 0.8.3 graphics libraries are no longer required. To accommodate this we're using an SVG-based "not found" image.

To pull this requirement out we'd want to remove this SVG usage. I can think of a few ways:

 - generate the "not found" image using PDF methods directly
 - use a pre-defined PDF object
 - switch back to a bitmap image, specifically a JPEG since it requires no processing.

fitztrev:
Hi @bsweeney - Thanks for letting me know. I wasn't aware there was a broken image placeholder. I took the 3rd option and switched it back to a JPEG.

PowerKiKi:
JPEG instead of SVG file in a PDF sounds like a regression to me. Shouldn't we keep things as much scalable as possible ?

Why remove the dependency in the first place if it has a use in core anyway ? I'd leave it as required.

fitztrev:
The `phenx/php-svg-lib` package required by Dompdf requires an outdated version of `sabberworm/php-css-parser`. There have been a couple releases to `php-css-parser` that fix a number of issues. I've had an open pull request to update that since November (see PhenX/php-svg-lib#45). As it currently stands, if someone wants to use both Dompdf + the latest `php-css-parser`, they can't.

bsweeney:
If the main concern is the version of php-svg-lib I'll see if I can get that PR accepted. I am keen on limiting the number of dependencies where possible, but it's also important to keep in mind what functionality we want to support out of the box. Making things suggested will lead to people needing help understanding why certain functionality does not work.

PhenX:
Hello, I just merge the @PowerKiKi 's PR (PhenX/php-svg-lib#45), and invited @bsweeney and @PowerKiKi as contributors on the project, as I may not push new code anymore on these projects.

composer.json
```diff
@@ -28,8 +28,7 @@
         "php": ">=5.4.0",
         "ext-dom": "*",
         "ext-mbstring": "*",
-        "phenx/php-font-lib": "0.5.*",
-        "phenx/php-svg-lib": "0.3.*"
+        "phenx/php-font-lib": "0.5.*"
```
PowerKiKi: Use newly released version instead:

```suggestion
        "phenx/php-font-lib": "0.5.*",
        "phenx/php-svg-lib": "^0.3"
```

composer.json
```diff
@@ -38,7 +37,8 @@
     "suggest": {
         "ext-gd": "Needed to process images",
         "ext-imagick": "Improves image processing performance",
-        "ext-gmagick": "Improves image processing performance"
+        "ext-gmagick": "Improves image processing performance",
+        "phenx/php-svg-lib": "Required if rendering SVG (^0.3)"
```
PowerKiKi:
```suggestion
        "ext-gmagick": "Improves image processing performance"
```

fitztrev:
I'll close now that `phenx/php-svg-lib` has been released with an updated constraint of `sabberworm/php-css-parser`.

But I do still think it could be moved to be a suggested rather than required dependency.

bsweeney:
We'll keep the suggestion in mind.

PR #1893, opened 2019-02-10
Update Version file to 0.8.3

dustinsilva:
Using this library in Drupal 7.x causes the reports/status page to incorrectly display an error saying the version was out of date...I apologize for not knowing how to correctly solve this issue, so I am creating this "change" to alert the owner of the project so he/she can fix this issue properly.


bsweeney:
ref #1849, but I appreciate the additional info why this change is being requested. We'll need to think through this a bit more.

PR #1903, opened 2019-03-11
Fixing Phar Deserialization vulnerabilities

jurysosnovsky:
Hello, thank you for this great library, we use it in our projects. At last code security check with RIPS, I found some vulnerabilities with phar files deserialization.  I'm not sure that your library are vulnerable, but I thout that better to fix any possible apportunite for the vulnerabilities.
I added check for the "phar://" file, it will throw an exception if someone try to use "phar" files.
For more information about vulnerabilities you could check https://blog.ripstech.com/2018/new-php-exploitation-technique/.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,9 @@ require 'vendor/autoload.php';
 
 ### Download and install
 
-Download an archive of dompdf and extract it into the directory where dompdf
-will reside
+Download a packaged archive of dompdf and extract it into the 
+directory where dompdf will reside
+
  * You can download stable copies of dompdf from
    https://github.com/dompdf/dompdf/releases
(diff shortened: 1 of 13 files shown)


bsweeney:
There's a lot of other things mixed in here unrelated to your change. Also, we prefer you submit PRs against the develop branch. I'm going to close. Can you pull a recent copy of develop and resubmit. We'll review then.

bsweeney:
Confusing to read through the discussion because of the unrelated commit comments. I think because of this we didn't appreciate the severity of the issue. At any rate, I am working on the PHAR issue now so no need to work on a separate PR. The vulnerability will be addressed in the 2.0.0 release.

PR #1904, opened 2019-03-11
Arabic Support

hebatollah:


0888:
thank you for this

bsweeney:
ref #712

simonberger:
What is the status of this? Are there unresolved problems? In the referenced issue I see some potential problems but the solution there was a bit different.
Aside of this I prefer the class as a composer dependency. Dompdf already has some external packages inside the code which should be outsourced instead of getting new friends. The page (ar-php.org) teases a composer package coming soon. But I think that this ad is there since quiet some time.

bsweeney:
I'm thinking this may not be the best solution since it's fairly limited in scope, but I wanted to review it a bit more before I decided one way or the other. I seem to recall there may have been an issue with the implementation that would make this something we wouldn't want to add to the core project.

simonberger:
This pull request runs all text through this package which can't be good. Additionally it newly initialized the class on every single text entry.
Yet from the short search I did the php-ar package seems the best available. It could be better implemented into dompdf.
It is difficulty to implement and test with zero knowledge of the languages though.

bsweeney:
I also think it would probably be more appropriate to implement this in the canvas adapter + PDF backend. For example, [PDFLib already has support](https://www.pdflib.com/pdflib-cookbook/complex_scripts/arabic-formatting/) we just don't have any logic to handle it into the adapter.

Implementation using the ar-php library does simplify things a bit by abstracting the character formatting to an individual call independent of the PDF library. But a user can just as easily call the script when constructing their HTML and feeding it to Dompdf.

So, in addition to other concerns I do have a question on where the appropriate location is for this logic.

src/Renderer/Text.php
```diff
@@ -80,7 +81,22 @@ function render(Frame $frame)
...
+            // treat also numbers with dot (decimal) and email
+            $text = preg_replace_callback('/\d+-\d+|\d+|\d+\.\d+|\S+@\S+/', function (array $m) { return strrev($m[0]); }, $text);
+        }*/
+        if ( ! class_exists( 'I18N_Arabic' ) ){
+
+            require_once( DOMPDF_DIR . "/src/I18N/Arabic/Glyphs.php" );
+            $Arabic = new I18N_Arabic_Glyphs('Glyphs');
+            $text = $Arabic->utf8Glyphs($text);
 
+        }
```
bsweeney: Does this logic mean the ar-php class is only called one time in the entire document? The condition around this code means it only runs if the class does not exist.

PR #1934, opened 2019-05-14
Resolved error Extension 'cpdf' is removed since PHP 5.1

shital7mahajan:
After running phpcs following command for library getting error. 

**Command**: phpcs --standard=PHPCompatibility --extensions=php,module,inc,install,test,profile,theme,css --runtime-set testVersion 7.2 

**File**:sites\default\libraries\dompdf\src\Adapter\CPDF.php
----------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------------------------------
 198 | ERROR | Extension 'cpdf' is removed since PHP 5.1; Use pecl/pdflib instead
----------------------------------------------------------------------------------------------

### DIFF
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -194,7 +194,7 @@ class CPDF implements Canvas
         }
 
         $this->_dompdf = $dompdf;
-        $cpdf_class_var = \Cpdf;
+        $cpdf_class_var = "\Cpdf";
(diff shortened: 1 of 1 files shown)


bsweeney:
We'll be going with the solution provided in #1851.

shital7mahajan:
Thanks for the update.

PR #1948, opened 2019-06-12
Do not allow failures on PHP 7.3

gocanto:
We should not be allowing failures on PHP 7.3 by now since it has been out for quite a bit of time. 

Thanks.


bsweeney:
Reminder, pull requests should be made against the develop branch.

gocanto:
cool; let me come back to this tonight then.

PR #1954, opened 2019-06-20
do not allow failure on php 7.3

gocanto:
Do not allow any CI failure for PHP 7.3

Replaces https://github.com/dompdf/dompdf/pull/1948


bsweeney:
I typically merge to master anyway (no need to hold up progress any more than it already is) but I appreciate that you took the time.

PR #1979, opened 2019-08-14
Add new fonts using dompdf/utils

milos1302:
Added 3 new fonts. Roboto supports many languages like Greek, Russian, German etc. Taviraj supports Thai language and ZCOOLXiaoWei supports Chinese language. These fonts have been tested in drupal project using drupal/entity_print module.


milos1302:
Last commit (3a3f600) was created because Roboto font supports many languages (their characters), much more than Halvetica font which is currently being used for 'sans-serif' font family.

bsweeney:
I've been considering adding new fonts to the library but adding a font with support for CJK glyphs significantly increases the size of the download (relative to the base package). I'd like to investigate another way to handle expanded glyph support via variations of the standard package (e.g. Dompdf-0.8.1-cjk) so that the user can download what's relevant to them. Additionally, I'd like to use a single font if possible to simplify implementation (e.g. Noto) along with unicode range support so that we can tell people to just specify, for example, "Noto Sans" in their font-family tag.

Additionally, I don't know how I feel about the change to the "sans-serif" generic font family. First, Helvetica is a core PDF font (included with all installed) and that's why we use it. If people need something more they should use a supporting font. Second, this change only impacts the sans-serif font family. If we're going to modify the generic font mappings we should do it for all of them (serif, monospace).

milos1302:
Thanks for the response. My knowledge about the fonts and how they work is minimal. This was just a workaround for an issue that I had with this library on a project that I'm working on. To be honest, I knew this was not going to be merged, and I just wanted to "hear" what you think about it. It would be great if you implement some (if not all) of the upgrades that you've mentioned. Thanks again.

casperbakker:
Maybe there is some way to create separate packages like `dompdf/dompdf-font-roboto` which you can install next to dompdf itself to add extra font support without any configuration. That way you can support as many fonts as you like, while keeping the main package as small as possible.

PR #1991, opened 2019-09-11
Allow install of newest sabberworm/php-css-parser

PowerKiKi:
This supersedes #1892


bsweeney:
Guess it's time to plan a release, then.

PowerKiKi:
If I were you I'd drop [non-supported PHP versions](https://www.php.net/eol.php) before releasing a 1.0.0. If you agree I'll make a PR for that too.

bsweeney:
Yeah, I'm fine with that.

Issue #1977, opened 2019-08-12
loadHtml() - $encoding argument has no effect


MarkMaldaba:
The `Dompdf::loadHTML()` function takes a second argument, called `$encoding`, but this currently has no effect.

On line 457 (DomPDF 0.8.3), before the argument is used for anything, we have the following, meaning that the argument is always forced to 'UTF-8' regardless of what you specify:
```php
        //FIXME: since we're not using this just yet
        $encoding = 'UTF-8';
```

The function seems to have some code that attempts to detect the correct encoding, but in my tests we are finding that we get invalid characters (boxes) when using iso-8859-1 content, even when we set this argument correctly, so this is not working effectively (if at all).

Is the intention to allow for non-UTF8 HTML documents to be renderable by DomPDF, or should we always be passing in UTF-8 and thus perform any relevant conversions before calling `loadHtml()`?

bsweeney:
Encoding is a tricky thing to handle gracefully since the encoding of the actual file may differ from the encoding indicated in the HTML headers, which may further differ from the encoding of the transport stream. I believe this is ultimately why we went this route at the time. I personally have looked at this code since that change was implemented.

To be safe you should just pass in UTF8-encoded text. We're happy to look into the issue further if you can provide additional details. Can you provide some more information on the text you're feeding dompdf? Are you loading a remote file or calling loadHtml with the encoding argument?

MarkMaldaba:
> Encoding is a tricky thing to handle gracefully since the encoding of the actual file may differ from the encoding indicated in the HTML headers, which may further differ from the encoding of the transport stream.

By my understanding, the transport encoding should be handled transparently at the transport layer 
before the file contents are returned to the PHP code.

However, more generally, the character encoding can be specified in one of two places - either external to the document, or as markup within the document itself.  In the context of `loadHtml()` any externally-specified encoding is passed in as `$encoding`, therefore for the purposes of this ticket we don't really need to worry about where this comes from, we just need it to be respected.  I have separately logged ticket #1987 to cover the correct handling of externally-defined encodings in `loadHtmlFile()`, but that is a separate issue.

In terms of in-document encoding, there is already some code which is handling this to a certain extent, but it's hard to see whether it is behaving correctly.  For the purposes of this ticket, we can assume it is and if that turns out not to be the case it should be logged as a separate ticket.

[...]

Issue #1987, opened 2019-09-09
loadHtmlFile() should use the Content-Type HTTP header, if present


MarkMaldaba:
Currently this function always pulls the encoding from the `<meta>` tag, if present, falling-back on `UTF-8` if not.

In addition, it should be checking the returned `Content-Type` header, if present (which would only apply to remote files, obviously) and if this specifies a character set, then this should take precedence over the value in the document (as per [the HTML specification](https://www.w3.org/International/questions/qa-html-encoding-declarations)).

Note that I am logging this for completeness, as it came up in the discussion for ticket #1977.  I don't think this is a particularly urgent issue.

lucasMesquitaBorges:
#2012 Will solve this issue when encoding is informed explicitly.. From Content-type header we still need a better detection encoding.

lucasMesquitaBorges:
#2012 Will solve this issue, as $encoding will be used and we detect it from content-type. It won't be overrided by UTF-8 before convertion.

PR #2012, opened 2019-10-28
Fix load html encoding

lucasMesquitaBorges:
This fixes loadHtml() - $encoding argument has no effect  #1977

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -343,10 +343,11 @@ class Dompdf
      * Parse errors are stored in the global array _dompdf_warnings.
      *
      * @param string $file a filename or url to load
+     * @param string $encoding Encoding of $file
      *
      * @throws Exception
      */
-    public function loadHtmlFile($file)
+    public function loadHtmlFile($file, $encoding = null)
     {
         $this->saveLocale();
 
@@ -384,10 +385,10 @@ class Dompdf
         }
 
         list($contents, $http_response_header) = Helpers::getFileContent($file, $this->httpContext);
+
         if (empty($contents)) {
             throw new Exception("File '$file' not found.");
         }
-        $encoding = 'UTF-8';
 
         // See http://the-stickman.com/web-development/php/getting-http-response-headers-when-using-file_get_contents/
         if (isset($http_response_header)) {
@@ -409,7 +410,7 @@ class Dompdf
      * @param string $encoding
      * @deprecated
      */
-    public function load_html($str, $encoding = 'UTF-8')
+    public function load_html($str, $encoding = null)
     {
         $this->loadHtml($str, $encoding);
     }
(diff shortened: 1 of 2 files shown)


src/Dompdf.php
```diff
@@ -409,31 +410,38 @@ public function loadHtmlFile($file)
...
+    public function loadHtml($str, $encoding = null)
     {
         $this->saveLocale();
 
-        // FIXME: Determine character encoding, switch to UTF8, update meta tag. Need better http/file stream encoding detection, currently relies on text or meta tag.
-        $known_encodings = mb_list_encodings();
-        mb_detect_order('auto');
-        if (($file_encoding = mb_detect_encoding($str, null, true)) === false) {
-            $file_encoding = "auto";
+        // FIXME: Determine character encoding, switch to UTF8, update meta tag. Need better http/file stream encoding detection when $encoding parameter not used, currently relies on text or meta tag.
```
simonberger: Fixme parts of this comment should be removed.

src/Dompdf.php
```diff
@@ -409,31 +410,38 @@ public function loadHtmlFile($file)
...
+        if($encoding === null) {
```
simonberger: Little syntax problem

simonberger:
This looks really good.

src/Dompdf.php
lucasMesquitaBorges: Detect enconding when not specified still need to be improved. Do you have a better comment to this part? I couldn't think in a better comment.

src/Dompdf.php
lucasMesquitaBorges: I have fixed the syntax problem here, if you still see any problem, please warn me.

src/Dompdf.php
lucasMesquitaBorges: I improved this comment, but still accepting suggestions!

simonberger: Form my point of view the Fixme comment mainly describes the not working content type detection from loadHtmlFile which calls this loadHtml method. This is something that was indeed broken and you solved here.
An even better detection is just on top but not worth a fixme comment. What could the better detection on html document level be in your opinion? Actually using the meta information instead of just rewriting it? I wondered about this current functionally at least.

lucasMesquitaBorges: I understood your point. Just a moment and i'll fix this comment.

simonberger:
I'll test this end of week before approving it. As I wrote code is looking good.

@bsweeney Shall I merge this into develop then? I don't know how it is intended to work here at the moment.

lucasMesquitaBorges:
Ok, anything just contact me.

bsweeney:
@simonberger yes if you're happy with it approve and merge. I went ahead and did it, but anything going into develop can be set for the next release milestone (0.8.5 in this case).

lucasMesquitaBorges:
@bsweeney If this PR is aproved by @simonberger I believe #1987 should be in milestone 0.8.5 too.

src/Dompdf.php
```diff
@@ -409,31 +410,38 @@ public function loadHtmlFile($file)
...
         $this->saveLocale();
 
-        // FIXME: Determine character encoding, switch to UTF8, update meta tag. Need better http/file stream encoding detection, currently relies on text or meta tag.
-        $known_encodings = mb_list_encodings();
-        mb_detect_order('auto');
-        if (($file_encoding = mb_detect_encoding($str, null, true)) === false) {
-            $file_encoding = "auto";
+        // Determine character encoding when $encoding parameter not used
+        if ($encoding === null) {
+            $known_encodings = mb_list_encodings();
```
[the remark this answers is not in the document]
lucasMesquitaBorges: Thanks @simonberger , this variable wasn't even needed.
Waiting for your review!

simonberger:
This can be merged now. As the commit history is somewhat verbose can you rebase it?

lucasMesquitaBorges:
Ok. After work i'll rebase it!

lucasMesquitaBorges:
@simonberger and @bsweeney  Rebase complete! I believe this PR can be merged now!

Issue #2007, opened 2019-10-18
Deprecated function: PDFlib::get_value(): Deprecated, use PDF_get_option() in get_value() (line 1289 of /var/www/html/vendor/dompdf/dompdf/src/Adapter/PDFLib.php)


rsanzante:
PHP: 7.1.17
PDFlib GmbH Binary-Version: 9.2.0
Dompdf: v0.8.3

I got this error message: 

> Deprecated function: PDFlib::get_value(): Deprecated, use PDF_get_option() in get_value() (line 1289 of /var/www/html/vendor/dompdf/dompdf/src/Adapter/PDFLib.php)

This comes from those lines: 
```
$asc = $this->_pdf->get_value("ascender", $fh);
$desc = $this->_pdf->get_value("descender", $fh);
```

PDFLib 9.x deprecated the get_value function. They instruct to use get_option. Unfortunately, the params seem different and I didn't find a way to translate from one function to another.

I  found, however, a PDFLib example to get font info here: 
https://www.pdflib.com/pdflib-cookbook/fonts/font_metrics_info/php/

They use the function info_font. I blindly replace the above code with:
```
$asc = $this->_pdf->info_font($fh, "ascender",  "fontsize=$size");
$desc = $this->_pdf->info_font($fh, "descender", "fontsize=$size");
```
$size is the font size and is needed by the info_font function. The value is available in the original code.

However, the value returned seems $size times greater. So I switched to:

```
$asc = $this->_pdf->info_font($fh, "ascender",  "fontsize=$size") / $size;
$desc = $this->_pdf->info_font($fh, "descender", "fontsize=$size") / $size;
```

It seems to work, but honestly I don't know what I'm doing.

Maye someone with more knowledge on PDFLib and Dompdf can validate this or propose a better solution.

rsanzante:
PR ready: https://github.com/dompdf/dompdf/pull/2020

noone-silent:
PDFLib 6 documentation says: 

`ascender = PDF_get_value(p, "ascender", font) * fontsize;`
`descender = PDF_get_value(p, "descender", font) * fontsize;`

So if the value from info_font is larger than $size, it seems they integrated the calculation into info_font. I don't think you need to divide it by $size. Even on the example page they do not divide it

rsanzante:
> PDFLib 6 documentation says:
> 
> `ascender = PDF_get_value(p, "ascender", font) * fontsize;`
[...]

PR #2020, opened 2019-11-15
Fix deprecated function usage.

rsanzante:
This commit tries to fix https://github.com/dompdf/dompdf/issues/2007.

See comments because I'm don't know if this is ok.

### DIFF
--- a/src/Adapter/PDFLib.php
+++ b/src/Adapter/PDFLib.php
@@ -1282,8 +1282,8 @@ class PDFLib implements Canvas
 
         $this->_pdf->setfont($fh, $size);
 
-        $asc = $this->_pdf->get_value("ascender", $fh);
-        $desc = $this->_pdf->get_value("descender", $fh);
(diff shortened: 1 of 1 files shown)


noone-silent:
This looks good for me. The get_value "ascender" and "descender" is only supported for PDFLib 6, from PDFLib 7 on you already should use info_font

rsanzante:
It seems fails are not due the commit, but a CI configuration issue. 
They fail in the first 20 seconds with this kind of error: 
image

The interesting point is here:
```The command "phpenv global 5.5" failed and exited with 1 during.```

The other fails because of 5.4:
```The command "phpenv global 5.4" failed and exited with 1 during .```

bsweeney:
Yes, CI is borked right now. If nobody submits a PR I'll see about fixing it.

This also looks correct to me based on the [PDFLib 9 API reference](https://www.pdflib.com/fileadmin/pdflib/pdf/manuals/PDFlib-9.0.3-API-reference.pdf).

rsanzante:
Thanks for your comments @bsweeney. I've updated the PR according.

PR #2024, opened 2019-12-03
Simplifying get/set methods

carloscarucce:


dimitriadamou:
What about camelCasing the variables that are written out as `admin_username`. Needs to become `adminUsername`

dmitryuk:
> What about camelCasing the variables that are written out with underscores e.g. `admin_username`, needs to become `adminUsername`

It is replaces in the code 
https://github.com/dompdf/dompdf/pull/2024/files#diff-eaf8868b994390699bd8b3db948d8aafR335

PR #2034, opened 2019-12-17
Initialize font subset temp file prior to access

IonBazan:
Starting from PHP 7.4, `fread()` [raises a notice](https://www.php.net/manual/en/migration74.incompatible.php#migration74.incompatible.core.fread-fwrite) when trying to read from a write-only file resource instead of silently failing:
```
PHP Notice:  fread(): read of 8192 bytes failed with errno=9 Bad file descriptor in /var/www/html/vendor/phenx/php-font-lib/src/FontLib/BinaryStream.php on line 145
```

Creating a new font file using `DirectoryEntry::encode()` tries to read from file which does not exist and is open in write-only mode. This change should fix the issue.

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -2838,7 +2838,8 @@ EOT;
 
                         // Write new font
                         $tmp_name = $this->tmp . "/" . basename($fbfile) . ".tmp." . uniqid();
-                        $font_obj->open($tmp_name, BinaryStream::modeWrite);
+                        touch($tmp_name);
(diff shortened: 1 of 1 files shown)


anderson-comvex:
👍 This fixes our issue during the `stream()` and `download()` as well.

bsweeney:
So the issue is not that we're attempting to read a write-only file resource, it's that the resource does not exist in the first place. This might be better addressed in php-font-lib, maybe, but this is an acceptable resolution.

PR #2049, opened 2020-01-19
Avoid error if ini_set() has been disabled for security reasons

earlyhost:


bsweeney:
I'm leaning more towards try/catch so we can log errors, but until we have a better logging mechanism there's not much point in it. Will include in next release.

PR #2050, opened 2020-01-19
Removed old (php < 5.4) array syntax

simonberger:
Completely changed array() to [] in whole project. PHP < 7.1 is not supported since end of last year and we have one code line with the new syntax in the project.

Can someone please have a look at this and merge it?


bsweeney:
The syntax may be older, but it is still valid and won't be deprecated. I've read that the two initialization methods are equivalent from a compiler perspective. So this would be a style preference and not a requirement for future releases.

Still, from a style perspective I'm OK with this change. I'll save it for the next release since it touches a lot of lines of code and we want to get 0.8.4 out the door.

simonberger:
You are right. With my comment I just wanted to point out why we can do this step now. Such a transition normally takes place in a project after PHP 5.3 has been dropped. It's more of a convenience change. While working in modern projects one does not get in touch with the old syntax very often anymore.

IonBazan:
@bsweeney You can automate it using [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) (using `array_syntax` rule) or [Rector](https://github.com/rectorphp/rector).

simonberger:
There are a lot of ways to achieve it. This is an automatic change using an inspection of PHPStorm.

simonberger:
Can we merge this to develop? I want to work on a feature and I can't help to get annoyed by the old array syntax.

bsweeney:
done

PR #2070, opened 2020-01-31
Restore error handler on \Throwable

### LINKED ISSUE #2067: Deprecation errors cause an exception
Any deprecation message will cause an exception to be thrown in `record_warnings` method: https://github.com/dompdf/dompdf/blob/adcd9738253afff908eadce9cff0d790f7ae2072/src/Helpers.php#L506-L508

I see a few solutions to that:

- Add `E_DEPRECATED`, `E_USER_DEPRECATED` and `E_STRICT` to excluded error levels (trivial)
- Use `try {...} catch (\Throwable $e)` as we are already dropped PHP 5 (that would probably break streaming support in some cases)
- Don't register any error handler at all because it should be an application responsibility - not the library one (BC break) and maybe use `@` when it's ok to suppress the errors.

IonBazan:
This PR fixes #2067 by restoring a default (or previous) error handler when `TypeError` (or any other introduced in PHP7) occurs.
Besides of that fix, a few syntax flavours have been added nearby.

### DIFF
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -190,7 +190,7 @@ class CPDF implements Canvas
         }
 
         if (mb_strtolower($orientation) === "landscape") {
-            list($size[2], $size[3]) = [$size[3], $size[2]];
+            [$size[2], $size[3]] = [$size[3], $size[2]];
         }
 
         $this->_dompdf = $dompdf;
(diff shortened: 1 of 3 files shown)


src/Helpers.php
```diff
@@ -503,7 +503,7 @@ public static function dompdf_debug($type, $msg)
     public static function record_warnings($errno, $errstr, $errfile, $errline)
     {
         // Not a warning or notice
-        if (!($errno & (E_WARNING | E_NOTICE | E_USER_NOTICE | E_USER_WARNING))) {
+        if (!($errno & (E_WARNING | E_NOTICE | E_USER_NOTICE | E_USER_WARNING | E_STRICT | E_DEPRECATED | E_USER_DEPRECATED))) {
```
IonBazan: I added these levels here in case if any piece of the code between our `try..finally` block triggers a deprecation in the future.

src/Adapter/CPDF.php
```diff
@@ -190,7 +190,7 @@ public function __construct($paper = "letter", $orientation = "portrait", Dompdf
         }
 
         if (mb_strtolower($orientation) === "landscape") {
-            list($size[2], $size[3]) = [$size[3], $size[2]];
+            [$size[2], $size[3]] = [$size[3], $size[2]];
```
bsweeney: 👍 Learn something new every day ([rfc:short_list_syntax](https://wiki.php.net/rfc/short_list_syntax))!

src/Adapter/CPDF.php
```diff
@@ -614,31 +614,34 @@ protected function _convert_gif_bmp_to_png($image_url, $type)
         $func_name = "imagecreatefrom$type";
 
         if (!function_exists($func_name)) {
-            if (!method_exists("Dompdf\Helpers", $func_name)) {
+            if (!method_exists(Helpers::class, $func_name)) {
```
bsweeney: Much better than "magic strings" as a colleague would say.

bsweeney:
@IonBazan any chance you can rebase your branch with Develop?

IonBazan:
@bsweeney here you are ✅

PR #2071, opened 2020-02-03
fix css selector issue

IonBazan:
This PR fixes an issue when a numeric (invalid) CSS selector is passed, causing `Trying to access array offset on value of type int` error to be thrown on PHP 7.4. 
The fix is to simply cast `$selector` to `string` using a type hint.

This shouldn't really occur in ideal world because numeric CSS selectors are not valid, but if anyone passes an invalid input, they should be simply ignored instead of throwing an error.

PS: The error is thrown here: https://github.com/dompdf/dompdf/pull/2071/files#diff-c9ab10f725a388f7b86e26f58d4085d4R476

To consider: Using https://symfony.com/doc/current/components/css_selector.html instead.

### DIFF
--- a/src/Css/Stylesheet.php
+++ b/src/Css/Stylesheet.php
@@ -450,9 +450,8 @@ class Stylesheet
      * @throws Exception
      * @return array
      */
-    private function _css_selector_to_xpath($selector, $first_pass = false)
+    private function _css_selector_to_xpath(string $selector, bool $first_pass = false): array
     {
-
         // Collapse white space and strip whitespace around delimiters
         //$search = array("/\\s+/", "/\\s+([.>#+:])\\s+/");
         //$replace = array(" ", "\\1");
@@ -892,7 +891,7 @@ class Stylesheet
             $query = rtrim($query, "/");
         }
 
-        return ["query" => $query, "pseudo_elements" => $pseudo_elements];
+        return ['query' => $query, 'pseudo_elements' => $pseudo_elements];
     }
 
     /**


bsweeney:
It's worth considering the symfony project. Can you create an issue to track the request and we'll look into it more?

bsweeney:
Conceptually you're right but I have a hard time seeing where that error would happen in actuality. The method is private to the Stylesheet class and so outside calls are not permitted. And we specifically [throw an error](https://github.com/dompdf/dompdf/blob/9a10e0594d2a3f61f2bb83eda4320b62bb43e833/src/Css/Stylesheet.php#L273) earlier during stylesheet processing if a key is not numeric. This method is not private, but it would have to manually called by the user to make this happen. Looking further up the call chain I don't see where this would happen in a normal use case.

Of course, that's just an initial analysis of the issue so feel free to correct me.

Also, with your change wouldn't we still throw an error, just a [different error](https://www.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration)?

IonBazan:
TBH it happened on one of our apps due to invalid html/css input which resulted in following css rule:
```css
500 {
  color: black;
}
```

`500` selector was used as `int` so `$selector[0]` failed with this error. Casting it to string fixes the issue.

bsweeney:
Ack, my bad. I wasn't paying attention and my initial test wasn't performed while running PHP 7.4. I'm surprised that the value is parsed as numeric, maybe some type sniffing on the part of PHP. I'll look at this some more when I get a chance.

IonBazan:
This happens in `add_style()`, when using the selector as array key: https://github.com/dompdf/dompdf/blob/993064d3b599f59777cf9691fcf30b15ce13408c/src/Css/Stylesheet.php#L284

Mellthas:
I think the type declarations are nice to have, anyway. So I think this is good to merge, apart from the change to the quotes, which is unrelated and makes them inconsistent with the rest of the file.

PR #2072, opened 2020-02-03
New PHP version requirement

jmvezic:


jmvezic:
It should also be changed in the repo description - it's still HTML to PDF converter (PHP5)

Issue #2073, opened 2020-02-03
Change adapter class members from private to protected


rsanzante:
From https://github.com/dompdf/dompdf/issues/2069 (comment https://github.com/dompdf/dompdf/issues/2069#issuecomment-581335061 and https://github.com/dompdf/dompdf/issues/2069#issuecomment-581490023) it seems private members from adapters (mainly Dompdf\Adapter\PDFlib but let's extend to the other adapters) can be changes to protected.

This would allow creating alternative adapters reusing existing classes, like the case explained on previous issue.

rsanzante:
PR created: https://github.com/dompdf/dompdf/pull/2074

PR #2074, opened 2020-02-03
Fb change adapter members from private to protected

rsanzante:
Changes private members of Dompdf\Adapter\PDFlib, Dompdf\Adapter\CPDF and Dompdf\Adapter\GPDF to protected.

Addresses https://github.com/dompdf/dompdf/issues/2073.


simonberger:
@bsweeney What's with this PR for 0.8.5? An uncritical change we should either merge or reject.

PR #2087, opened 2020-02-12
define default value for Imagick::negateImage

Alexander1000:
Extension php-imagick with ImageMagick version 7 and above default value is changed for method negateImage
I fix it in concrete value

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -4871,7 +4871,10 @@ EOT;
             if ($imagick->getImageAlphaChannel() !== 0) {
                 $alpha_channel = $imagickClonable ? clone $imagick : $imagick->clone();
                 $alpha_channel->separateImageChannel(\Imagick::CHANNEL_ALPHA);
-                $alpha_channel->negateImage(true);
+                // Since ImageMagick7 negate invert transparency as default
+                if (\Imagick::getVersion()['versionNumber'] < 1800) {
+                    $alpha_channel->negateImage(true);
+                }
                 $alpha_channel->writeImage($tempfile_alpha);
 
                 // Cast to 8bit+palette


lib/Cpdf.php
```diff
@@ -4871,7 +4871,7 @@ protected function addImagePngAlpha($file, $x, $y, $w, $h, $byte)
             if ($imagick->getImageAlphaChannel() !== 0) {
                 $alpha_channel = $imagickClonable ? clone $imagick : $imagick->clone();
                 $alpha_channel->separateImageChannel(\Imagick::CHANNEL_ALPHA);
-                $alpha_channel->negateImage(true);
+                $alpha_channel->negateImage(true, \Imagick::CHANNEL_ALPHA);
```
bsweeney: Is this the right channel to use? Based on my reading, the default channel changed from RGB to RGBA. So for this operation should we be using `\Imagick::CHANNEL_RED | \Imagick::CHANNEL_GREEN | \Imagick::CHANNEL_BLUE`?

bsweeney:
Also, can you rebase to the devlop branch?

bsweeney:
As much for me as anyone else, here's a good post on the change:
http://blog.awm.jp/2018/11/18/im7negate/

lib/Cpdf.php
Alexander1000: for ImageMagick 7 negate in principle not need
may be Imagick::getVersion() and check version?

bsweeney: I'm open to optimizing based on version.

bsweeney:
I'm not currently set up to test with this library. I'll see what I can do about that.

I've also been thinking that usage of the IMagick/GMagick extensions should probably be configurable. No doubt if someone has either extension they're likely to use it, but we do occasionally run into a bug that breaks the integration and there's no recourse when that happens beyond modifying Dompdf locally or disabling the extension..

Alexander1000:
Don't you need a patch to work with ImageMagick7?

Issue #1388, opened 2017-02-18
pagebreak with table rowspan not created correctly


Shinseiki86:
I have a table with merged cells in several rows, but when I use dompdf and performs a page break, the cells are not created correctly.

How can I correct it?
Thanks.

Table: https://jsfiddle.net/shinseiki86/24LL3fd5/

Result: 
![image](https://cloud.githubusercontent.com/assets/22685115/23094984/74f9dcd8-f5d0-11e6-9d6c-cebfd720734a.png)

bsweeney:
I'm fairly certain there's an existing issue for this ... but I couldn't find it in a quick search.

Shinseiki86:
Right! I have encountered some similar issues, but no solution.

https://sourceforge.net/p/dompdf/bugs/39/

bsweeney:
Maybe #232 is most appropriate, which points to #55.

prtk418:
is this still unsolved?

amayer-glei:
> 
> 
> is this still unsolved? i am facing same issue

yes still broken... switched to wkhtlmpdf now.

Ajahuanex:
> > is this still unsolved? i am facing same issue
> 
> yes still broken... switched to wkhtlmpdf now.

but, wkhtmlpdf don't work in shared hosting.

simonberger:
Minimum example to test the problem (A4, landscape)

```
<html>
<head>
	<style>
		table, td, th {
			border: 1px solid black;
		}
	</style>
</head>
<body>
<table>
  <tbody>
    <tr>
		<td rowspan="3">1</td>
		<td>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br>2<br></td>
    </tr>
	<tr><td>3</td></tr>
	<tr><td>4<br>4<br>4<br></td></tr>
  </tbody>
</table>
</body>
</html>
```

Any active rowspan is not remembered on page breaks.
I currently debug this behavior but it is really deep in the renderer and/or decorator tasks.

jonathanpmartins:
having the same problem here!

```page-break-inside: avoid;``` should work with ```tbody``` group tag.

Jensovic:
[...]

PR #2090, opened 2020-02-17
Fix: page breaking on rows with rowspan.

Ujifman:
1. I added restriction to break before row, if page_break_before: avoid specified for <tr>.
2. I added logic to define if current row has spanned from previous rows cells, so one restricts page breaking on rows connected with rowspan. #1388

### DIFF
--- a/src/Frame.php
+++ b/src/Frame.php
@@ -415,7 +415,7 @@ class Frame
     }
 
     /**
-     * @return FrameList|Frame[]
+     * @return FrameList|Frame[] !! TODO fixup doc. It never returns Frame[]
      */
     public function get_children()
     {
--- a/src/FrameDecorator/Page.php
+++ b/src/FrameDecorator/Page.php
@@ -424,6 +424,14 @@ class Page extends AbstractFrameDecorator
                         $p = $p->find_block_parent();
                     }
 
+                    // Avoid page break for table row if specified
+                    if ($frame->get_style()->page_break_before === "avoid") {
+                        Helpers::dompdf_debug("page-break", "before: avoid for table-row");
+
+                        return false;
+                    }
+
+
                     // Avoid breaking before the first row of a table
                     if ($table && $table->get_first_child() === $frame || $table->get_first_child()->get_first_child() === $frame) {
                         Helpers::dompdf_debug("page-break", "table: first-row");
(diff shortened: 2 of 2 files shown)


simonberger:
@Ujifman I tested this and got the following pdf from #1388 
[dompdf_2090.pdf](https://github.com/dompdf/dompdf/files/4222031/dompdf_2090.pdf)

simonberger:
@Ujifman To precise my testing and review of the code. Am I correct that your change delays the point a table with a spanned row breaks or should it completely fix the behavior?

bsweeney:
I don't know that this change is sufficient enough for the designated issue. It does make an attempt to keep rows with a spanning cell together, but Dompdf will still split a rowspanned cell across pages when the row it's on has already been paged. And since the real issue is that the row-spanned cell is lost from the table layout after the page break this doesn't really provide much relief.

Let us know if you feel we're missing something.

Ujifman:
I wasn't clear enough. I didn't solve the problem fully. In my case I have small rowspans in pdf, i.e. rowspan is never greater than page length, so I check if page break came to spanned row, and go up to the span beggining to make a split. So as @simonberger saw in test, if rowspan is bigger then page, cells are lost. The only general solution is to copy spanned cell on every page break. So my solution is not full, but I think, it will be useful for those projects, who has spanned cells less then page length.

Ujifman:
I wasn't clear enough. I didn't solve the problem fully. In my case I have small rowspans in pdf, i.e. rowspan is never greater than page length, so I check if page break came to spanned row, and go up to the span beggining to make a split. So as @simonberger showed in test, if rowspan is bigger then page cells are still lost. The only general solution is to copy spanned cell on every page break. So my solution is not full, but I think, it will be useful for those projects, who has spanned cells less then page length.

bsweeney:
Thanks for the clarification.

simonberger:
I would prefer to fix the actual bug. There shouldn't be the need to avoid breaks on rowspans aside of the current obvious problem.
But I agree to your argument that this change can work around the issue in at least a lot of cases and vote to merge it if we won't fix it soon.
I yet failed to find the missing information in the splitted table causing the overwrite of the spanned column.

sektonic:
> 1. I added restriction to break before row, if page_break_before: avoid specified for `<tr>`.
> 2. I added logic to define if current row has spanned from previous rows cells, so one restricts page breaking on rows connected with rowspan. #1388

Thank you sir

bsweeney:
Let's push to the next release so we can better evaluate what to do with this. I don't want to implement something that we'll have to remember to roll back later even if it provides temporary relief.

fillobotto:
Do you have any update about this problem?

Mellthas:
Haven't looked into the second part, but the first change about `page-break-before: avoid` support for table rows would be nice to have independently of how the discussion is resolved. Can you put up a separate PR for just that change, @Ujifman, so it can be merged?

PR #2092, opened 2020-02-18
Passes Options to static getDefaultStylesheet

BrianHenryIE:
See https://github.com/dompdf/dompdf/issues/902#issuecomment-257152054 by @marcleimvs


Mellthas:
This should be obsolete with commit e5138acadc5276a747810be7e54ae7928e55b227.

bsweeney (approved):

bsweeney:
@Mellthas thanks for going through these PRs I think you're right.

Issue #2098, opened 2020-02-24
Allow to specify local path for styles


ghost:
According to comments in `Dompdf` class `$protocol` could be `file://`.

```
    /**
     * Protcol used to request file (file://, http://, etc)
     *
     * @var string
     */
    private $protocol;
```

The problem is that `build_url` function does not take `file://` protocol into account for relative paths.

ytsurk:
This patch helped me too to use files directly from the filesystem in a container environment. (https://www.drupal.org/project/entity_print/issues/2865185)

bsweeney:
Looks like a straightforward enough change. We'll review the PR and get back to you.

bsweeney:
Looking at this again, in what situation is the current logic problematic? Are we talking protocol-specified local paths such as `file://images/example.jpg`?

ytsurk:
In my case, running in docker w/o knowing the url/ip of the localhost (thus the desired file(s) could not be retrieved), switching from http(s):// to the file:// protocol was a painless solution that made the setup working.

dealancer:
@bsweeney any updates on this?

bsweeney:
I'm not sure of the exact nature of the issue. There are some changes in the next release that may address the issue. I had to make some changes to resource handling that included local resource handling. If anyone is able to test out what's in the master branch I'd appreciate it. The release will be posted in a day or so.

ytsurk:
I now tested. The issue remains and the proposed patch (https://github.com/dompdf/dompdf/pull/2099/commits/82456fd86db4e4b81549dd28eda6d1cbfd6e2850) still solves the problem.

@bsweeney

"We"/the Drupal module/or simply maybe also others ;) use the base_path also with the file:// protocol, whereas dompdf does not.

The question now is, is this by intention, then we will change in the Drupal module and others need to follow, or can we change dompdf to use the base_path also with the file:// protocol, then this merge request should go in!

bsweeney:
I'm still unclear on where the issue is here. Dompdf does support file-protocol for the base and Dompdf also utilizes the base properties to build out a relative URL.

On top of that I think there may be some confusion about what the base properties do. The base properties do not set a root against which URLs are set, they act more like the base href you would provide in an HTML document.

If I'm misunderstanding the issue please post back with a more detailed example.

dealancer:
@bsweeney,

Which properties you need to set in order to make relative paths work? Are you taking about following methods?

```
$dompdf
[...]

PR #2099, opened 2020-02-24
Add support for file protocol in build_url

dealancer:
Addresses https://github.com/dompdf/dompdf/issues/2098


bsweeney:
I don't think this is the right change to make. It looks like what you're trying to do is force Dompdf to adhere to the document root of your application. But that's not the intention of the base parameters. They're intended to act similar to the base href of an HTML document.

The current logic already supports file-protocol for the base protocol or URL. Current logic prepends the base path if the URL isn't absolute, which is the expected behavior. With this change when the base protocol is "file://" Dompdf would always prepend the base path. We only want to prepend the base path if the url is relative.

Issue #2097, opened 2020-02-24
the table feet is larger than the page


romanown:
when i printing the table it more then page. i tryed any version (from 0.6 to 0.8.4). layout need is portrait.
it is my html
```
<!DOCTYPE html>
<html>
<head>
  <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
</head>
<body>
  <table  border="1" cellpadding="2">
    <thead>
    <tr>
        <th width="20px">#</th>
        <th width="60px">Фото</th>
        <th width="120px">Предмет счета</th>
        <th width="70px">Штрих-код</th>
        <th width="40px">Кол-во</th>
        <th>Цена, руб.</th>
        <th>Скидка, руб.</th>
        <th>Сумма, руб.</th>
        <th width="70px">Ком-рий</th>
    </tr>
    </thead>
    <tbody>
           <tr>
            <td width="20px">1</td>
            <td width="60px"><img src="http://placehold.it/150x100" class="product-image"></td>
            <td width="120px" class="product-info">
                артикул:<br>
                размер: <br>
                бренд: <br>
                наименование: <br>
            </td>
            <td width="70px">
                        <img src="http://placehold.it/150x100" class="barcode">
            </td>
            <td width="40px">7</td>
            <td>56756</td>
            <td>89767</td>
            <td>6767</td>
            <td width="70px">орл ол олппо аппо попо поп</td>
        </tr>
    </tbody>
</table>
</body>
</html>
```

it is to `https://github.com/dompdf/dompdf/blob/master/src/FrameReflower/Text.php` line 413
`$words = array_flip(preg_split('/([^\S\xA0]+|-+)/u', $text, -1, PREG_SPLIT_DELIM_CAPTURE));`
i edited it to
`$words = array_flip(preg_split('/[^\S\xA0]+/u', $text, -1, PREG_SPLIT_DELIM_CAPTURE));`
and it worked for my.

bsweeney:
I'm unable to discern any significant different in rendering between 0.6 and 0.8.4 in the debug helper. Nor after making the suggested change locally. Is there any additional styling that could be affecting the render? Can you provide a version of the document rendered before the upgrade to 0.8.x?

The change you made should have the opposite affect, I would think, since you're removing one of the characters that is used to split a string into multiple lines.

romanown:
[...]

PR #2102, opened 2020-02-25
Fix changes

romanown:
a change in the regular expression that keeps the width of the table within the page and the width of the columns within the specified values. it is most noticeable in cells with images located in them.as writed into issue #2097

### DIFF
--- a/src/FrameReflower/Text.php
+++ b/src/FrameReflower/Text.php
@@ -408,7 +408,7 @@ class Text extends AbstractFrameReflower
                 // split the text into words
                 // The regex splits on everything that's a separator (^\S double negative), excluding nbsp (\xa0), plus dashes
                 // This currently excludes the "narrow nbsp" character
-                $words = array_flip(preg_split('/([^\S\xA0]+|-+)/u', $str, -1, PREG_SPLIT_DELIM_CAPTURE));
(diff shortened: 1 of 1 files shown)


src/FrameReflower/Text.php
```diff
@@ -408,7 +408,7 @@ function get_min_max_width()
                 // split the text into words
                 // The regex splits on everything that's a separator (^\S double negative), excluding nbsp (\xa0), plus dashes
                 // This currently excludes the "narrow nbsp" character
-                $words = array_flip(preg_split('/([^\S\xA0]+|-+)/u', $str, -1, PREG_SPLIT_DELIM_CAPTURE));
+                $words = array_flip(preg_split('/[^\S\xA0]+/u', $str, -1, PREG_SPLIT_DELIM_CAPTURE));
```
bsweeney, line 411: You've changed how Dompdf splits words within the min/max width calculation logic. But you did not make the same change to the actual line break logic on line [111](https://github.com/dompdf/dompdf/blob/5e6237e26b874f07aed6a81725515496ebbeadc6/src/FrameReflower/Text.php#L111). This will result in inconsistent results when Dompdf is laying out the lines.

bsweeney: While this particular problem with the PR is easily fixed I do not believe the change would resolve the reported issue.

Issue #2103, opened 2020-02-29
text-align: right doesn't work


maziyaryosofi:
hi
when i use text-align: right on a div not working correctly.
just lastest line of text work correctly and all of top line not set text-align:right.
how can i solve this? anyone ever found this problem too?

simonberger:
@maziyaryosofi Please add a minimal example of your problem.
I did some test which all showed no problems. If this does not work somewhere, it should be caused by a special case.

simonberger:
https://github.com/barryvdh/laravel-dompdf/issues/664

maziyaryosofi:
hi

my controller code is:
`$data['description'] = 'این یک متن تست جهت بررسی موضوع باشد. این یک متن تست جهت بررسی موضوع باشد. این یک متن تست جهت بررسی موضوع باشد.';

$pdf = PDF::loadView('printpdf', $data)->setPaper('a4', 'landscape');
return $pdf->download('print.pdf');`

and my printpdf.balde.php is:
`<!DOCTYPE html>
<html lang="fa">
<head>
    <meta charset="utf-8">
    <meta http-equiv="Content-type" content="text/html; charset=UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no, minimum-scale=1, maximum-scale=1, viewport-fit=cover, shrink-to-fit=no">
    <style type="text/css">
        @font-face {
            font-family: 'iransans';
            font-style: normal;
            font-weight: normal;
            src: url({{ storage_path('fonts\iransans.ttf') }}) format('truetype');
        }
        body{
            margin: 0 0 0 0;
            padding: 0 0 0 0;
            direction: rtl;
            width: 297mm;
            height: 210mm;
            direction: rtl;
        }
        .control-body{
            position: relative;
            left: 114mm;
            top: 48mm;
            width: 178mm;
            height: 89.5mm;
            padding: 0 0 0 0;
            margin-right: 0;
        }
        .description{
            position: absolute;
            top: 67.2mm;
            right: 87mm;
            width: 54mm;
            height: 17mm;
            /*background: #000000;*/
            color: #000000;
            font-family: "iransans";
            font-size: 13px;
            line-height: 5px;
            vertical-align: middle;
            text-align: right;
            padding: 0 0 0 0;
            margin: 0 0 0 0;
        }
    </style>
</head>

<body dir="rtl" style="direction: rtl">
    <div class="control-body">
        <div class="description">{{ $description }}</div>
[...]

Issue #1009, opened 2015-08-27
Right to left text


mrtuvn:
Hello devs team
Could you tell me this problem have fixed or not? Sometime pdf render font not properly Which fonts recommend to use for better render arabic charater. 

Sorry if this thread duplicate
Best regards
tuna

mrtuvn:
I am waiting for the response please

bsweeney:
We have yet to look into supporting this feature. I don't know that we have a specific issue related to adding support for RTL, but issue #712 has some relevant discussion.

vstyle:
Hi there.
The task opened 7 years ago and we still need RTL to be included, dear dev team could you please pay some attention on the task, we really need this feature.

Thank you!

Saifallak:
https://github.com/Ab-Alselwi/laravel-arabic-html may help, not so perfect but may help.

snipe:
@UVLabs In [my experiments](https://github.com/grokability/snipe-it/issues/17808), it cannot be used. (That would have obviated this whole feature request, tbh.) It's also a really hard problem to solve, so I get it. We ended up switching to TCPDF and doing all kinds of shenanigans to determine RTL/LTR because our users sometimes use a mix of both in their PDFs.

PR #2107, opened 2020-03-07
RTL Language Support

simonberger:
This feature adds:
- Text is aligned right when direction is set to RTL
- Text of RTL languages is reordered while LTR languages should be untouched
- Dir attribute is supported beside the direction style on the most important elements

This does not use the ar-php library which was unstable in my tests and provided often wrong results. Instead it includes https://github.com/tecnickcom/tc-lib-unicode which looks promising.

I have zero knowledge of all languages like Arabic, Hebrew, Persian this supports. If someone has some more insights and could find flaws of this implementation or on the use of this library instead of ar-php please let us know here.

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -1992,6 +1992,15 @@ class Style
         $this->_props_computed[$prop] = (is_array($munged_color) ? $munged_color["hex"] : $munged_color);
     }
 
+    public function set_direction($direction)
+    {
+        $this->_props["direction"] = $direction;
+        $this->_prop_cache["direction"] = null;
+        if ($direction === 'rtl') {
+            $this->text_align = 'right';
+        }
+    }
+
     /**
      * Sets color
      *
--- a/src/FrameReflower/Text.php
+++ b/src/FrameReflower/Text.php
@@ -8,6 +8,7 @@
  */
 namespace Dompdf\FrameReflower;
 
+use Com\Tecnick\Unicode\Bidi;
 use Dompdf\FrameDecorator\Block as BlockFrameDecorator;
 use Dompdf\FrameDecorator\Text as TextFrameDecorator;
 use Dompdf\FontMetrics;
@@ -219,7 +220,15 @@ class Text extends AbstractFrameReflower
         switch ($style->white_space) {
             default:
             case "normal":
-                $frame->set_text($text = $this->_collapse_white_space($text));
+                $text = $this->_collapse_white_space($text);
+
+                if ($style->direction === 'rtl') {
+                    $bidi = new Bidi($text, null, null, 'R', false);
+                    $text = $bidi->getString();
+                }
+
+                $frame->set_text($text);
+
                 if ($text == "") {
                     break;
                 }
@@ -228,15 +237,32 @@ class Text extends AbstractFrameReflower
                 break;
 
             case "pre":
+                if ($style->direction === 'rtl') {
+                    $bidi = new Bidi($text, null, null, 'R', false);
+                    $text = $bidi->getString();
+                }
+
                 $split = $this->_newline_break($text);
                 $add_line = $split !== false;
                 break;
 
             case "nowrap":
-                $frame->set_text($text = $this->_collapse_white_space($text));
+                $text = $this->_collapse_white_space($text);
+
+                if ($style->direction === 'rtl') {
+                    $bidi = new Bidi($text, null, null, 'R', false);
+                    $text = $bidi->getString();
+                }
+
+                $frame->set_text($text);
                 break;
 
             case "pre-wrap":
+                if ($style->direction === 'rtl') {
+                    $bidi = new Bidi($text, null, null, 'R', false);
+                    $text = $bidi->getString();
+                }
+
                 $split = $this->_newline_break($text);
 
                 if (($tmp = $this->_line_break($text)) !== false) {
@@ -249,7 +275,13 @@ class Text extends AbstractFrameReflower
 
             case "pre-line":
                 // Collapse white-space except for \n
-                $frame->set_text($text = preg_replace("/[ \t]+/u", " ", $text));
+                $text = preg_replace("/[ \t]+/u", " ", $text);
+
+                if ($style->direction === 'rtl') {
+                    $bidi = new Bidi($text, null, null, 'R', false);
+                    $text = $bidi->getString();
+                }
+                $frame->set_text($text);
 
                 if ($text == "") {
                     break;
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,8 @@
         "ext-dom": "*",
         "ext-mbstring": "*",
         "phenx/php-font-lib": "^0.5.1",
-        "phenx/php-svg-lib": "^0.3.3"
+        "phenx/php-svg-lib": "^0.3.3",
+        "tecnickcom/tc-lib-unicode": "~1.3.11"
     },
     "require-dev": {
         "phpunit/phpunit": "^7.5",
--- a/src/Css/AttributeTranslator.php
+++ b/src/Css/AttributeTranslator.php
@@ -61,6 +61,7 @@ class AttributeTranslator
             ],
             'rules' => '!set_table_rules',
             'width' => 'width: %s;',
+            'dir' => 'direction: %s;'
         ],
         'hr' => [
             'align' => '!set_hr_align', // Need to grab width to set 'left' & 'right' correctly
@@ -70,6 +71,7 @@ class AttributeTranslator
         ],
         'div' => [
             'align' => 'text-align: %s;',
+            'dir' => 'direction: %s;'
         ],
         'h1' => [
             'align' => 'text-align: %s;',
@@ -95,6 +97,7 @@ class AttributeTranslator
         ],
         'p' => [
             'align' => 'text-align: %s;',
+            'dir' => 'direction: %s;'
         ],
 //    'col' => array(
 //      'align'  => '',
@@ -115,6 +118,7 @@ class AttributeTranslator
             'nowrap' => 'white-space: nowrap;',
             'valign' => 'vertical-align: %s;',
             'width' => 'width: %s;',
+            'dir' => 'direction: %s;'
         ],
         'tfoot' => [
             'align' => '!set_table_row_align',
@@ -127,6 +131,7 @@ class AttributeTranslator
             'nowrap' => 'white-space: nowrap;',
             'valign' => 'vertical-align: %s;',
             'width' => 'width: %s;',
+            'dir' => 'direction: %s;'
         ],
         'thead' => [
             'align' => '!set_table_row_align',
@@ -142,6 +147,7 @@ class AttributeTranslator
             'bgcolor' => '!set_background_color',
             'link' => '!set_body_link',
             'text' => '!set_color',
+            'dir' => 'direction: %s;'
         ],
         'br' => [
             'clear' => 'clear: %s;',
@@ -177,6 +183,7 @@ class AttributeTranslator
         'li' => [
             'type' => 'list-style-type: %s;',
             'value' => 'counter-reset: -dompdf-default-counter %d;',
+            'dir' => 'direction: %s;'
         ],
         'pre' => [
             'width' => 'width: %s;',


src/Css/Style.php
```diff
@@ -1992,6 +1992,15 @@ protected function set_prop_color($prop, $color)
         $this->_props_computed[$prop] = (is_array($munged_color) ? $munged_color["hex"] : $munged_color);
     }
 
+    public function set_direction($direction)
+    {
+        $this->_props["direction"] = $direction;
+        $this->_prop_cache["direction"] = null;
+        if ($direction === 'rtl') {
+            $this->text_align = 'right';
+        }
```
bsweeney: Will this over-ride user-defined alignment styling?

simonberger: Yes that is another problem I noted as well and forgot about. Do we have a chance to detect a text-align: left is set by the user instead our default value or work around it somehow?
To improve it a bit we could do at least:

```
if ($direction === 'rtl') {
     if ($this->text_align === 'left') {
          $this->text_align = 'right';
     }
}
```

simonberger: I should have stepped through it earlier. :) I thought the defaults are set initially but seems to just are taken when requesting text_align.

The following change works just fine if text_align is set in the same attribute as direction.

```
if ($direction === 'rtl') {
    if (!isset($this->_props['text_align'])) { 
        $this->text_align = 'right';
    }
}
```

But inheritance has another issue:

```
<div style="text-align: left">
	<p style="direction: rtl">
		Hello World !
	</p>
</div>
```

Inheritance generally works correct and sets the value on the child before direction is applied. But the newly parsed styles (Stylesheet::_parse_properties) are added on a new Style class. I do not know there that the text-align is inherited.

bsweeney: Checking the `_props` array is exactly how I was thinking that could be done.

As for inheritance, that does present a problem. I think instead of setting the value we should add a getter for the `text-align` property. If not set it checks to see if `direction` is "rtl" and returns right in that case.

bsweeney:
What were your thoughts regarding block rendering. I was thinking we render from the last child instead of the first child for containers where direction is "rtl". But it would require knowing which children fit on the line first rather than calculating as we go.

simonberger:
I did not really think about it in detail. The problem is even more complex though when trying to implement it perfectly.

This line is rather simple:

```
<p style="direction: rtl">
	12 <span style="color: red">34</span> 56
</p>
```

We want to receive 56 34 12.
With the current implementation we get 21 34 65 due to a bug in the library I guess will be fixed soon.
But we would want to get 56 34 12 anyway.

The approach to just reorder the blocks of one line (if we would be able to solve the problem you described) would produce a wrong result here:

```
<p style="direction: rtl">
	12<span style="color: red">34</span>56
</p>
```

We would want an unchanged order of 123456. The same belongs not just to english parts in an RTL block but also to the RTL languages as far as I understand where (some) words or glyphs are not reordered.

My initial thoughts were to add the changes even earlier in the dompdf process and passing the content of the whole block into the Bidi calculation. and rearranging the inline blocks accordingly.
But as wrote  I did not evaluate if this is feasible nor how much effort this would cost.

src/Css/Style.php
simonberger: This sounds like a really good idea. I'll try that approach.

simonberger:
Starting with this very basic support (after the aligning fix) would be a good idea in my opinion. There are so many things to do or consider having a better implementation. Reverting order of table columns is just one more coming to my mind.
We could try to tackle it in a later version.

src/Css/Style.php
simonberger: I pushed a fix to this.
Caused by the way the __get method handles unset props (it sets the default value) I was forced to do a change with more implications. The default value for `text-align` is now `initial` and the default value `left` is set in the `get_text_align` method.
It works in my tests. But I do not fully understand the computing and property cache system if it could have a problem now anywhere.

bsweeney: At a high level, when settings and getting styles from the Style class the following rules is used:

- When set, either by the generic setter or the specific property setter, the various property array (_props, _props_computed, _prop_cache) are cleared.
- The property in the _props array is set to the value passed in (the [specified value](https://www.w3.org/TR/CSS2/cascade.html#specified-value)).
- If the property requires a computation it is done in a property-specific setter. The value property in the _props_computed array is set to this value (the [computed value](https://www.w3.org/TR/CSS2/cascade.html#computed-value)).
- Any final calculations on the computed value are done in property-specific getters (the [used value](https://www.w3.org/TR/CSS2/cascade.html#used-value)).
- Unset properties are set to the default when the getter is used.
- When get through the generic setter the returned value in _prop_cache is set.
- Getters are typically only used at render time, after styles have been parsed and applied to elements.

This is all kind of a work in progress. I think the lack of consistency in how properties are set/get due to partial implementation is to blame. I have an idea of how the getter logic can be tweaked to better accommodate unset properties. Shall I commit the change for review? Then I have a suggestion for how to tweak the text-align getter.

simonberger: Sure, commit directly if you can improve it.

simonberger:
Addresses:
#1009
#2103

I do not completely understand the issue in #712 but that shouldn't be fixed with this right?

src/Css/Style.php
bsweeney: Will do. I decided to go full out with the implementation of the computed value. I won't add any new logic but I will be moving some calculations into the setters. I did a quick test with what I have now and I believe the styling is working as expected though I have a bit more testing to do. I'll submit that work as a separate PR so you can review before you update this branch.

bsweeney:
I was looking at this and I have some ideas about 712. Let me collect my thoughts and get back to you.

maziyaryosofi:
it isn't work on the persian language.
also have problem with word and align:right also not working.
in this method the text like سلام print as مالس

simonberger:
Then you most likely did not set a `direction: rtl` css style

simonberger:
Activated shaping of the tc-lib-unicode library in https://github.com/dompdf/dompdf/pull/2107/commits/62e7b36a8d9bf401958d28c119b3b97ac0a3c22e which should fix this and also the issues in #712 
I deactivated this initially because of some results that seemed not correct to me and now somehow almost forgot about it. On this example it produces a correct result.
We need some more testing.

ThomasCaud:
Hello all. First of all, thank you for your work and support! I just would like to know if you know when the dompdf version will be available with this feature? I can see the PR is approved now. :)
Thomas

simonberger:
We still had one issue which now is fixed. (#2134, #2135)
This will be merged when #2134 is finished.

Releasing is managed by @bsweeney. From my side I'll just try to merge all my other pending pull requests and won't start something else for the next release.

ThomasCaud:
Thanks for these informations. :)

bsweeney:
@simonberger I went ahead and resolved a conflict in the Style class so this PR should be up-to-date with the changes I made.

simonberger:
Thank you for Merging. Results are similar to how our merge branch looked like which is good.
From my side we could merge this now.
@bsweeney 
Do you want to have another look at #712 and/or did you test my change activating shaping?

bsweeney:
I'll take another quick look, but I think we're probably good to merge.

src/FrameReflower/Text.php
```diff
@@ -221,7 +222,15 @@ protected function _layout_line()
         switch ($style->white_space) {
             default:
             case "normal":
-                $frame->set_text($text = $this->_collapse_white_space($text));
+                $text = $this->_collapse_white_space($text);
+
+                if ($style->direction === 'rtl') {
+                    $bidi = new Bidi($text, null, null, 'R', true);
```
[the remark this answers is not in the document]
simonberger: If the document says it wants RTL it should get that don't you think so?
Which problems do you want to solve with letting tc-lib-unciode deciding what to do?

If we are doing the step to pass all text it is another story surely. But I don't know if browsers are having an auto detection. I doubt that as there is just the CSS value ltr or rtl.

simonberger:
@bsweeney Thank you for the detailed review. That's what I was hoping for.

**(1) Font subsetting**
Good point. First a question. Isn't it generally possible (needed effort aside) to create the embedded font object dynamically or at the end or does this have to be complete at the start?

To fix that in the current state we have some options beside the tc-unicode-lib double run:
- Disable font subsetting if we have any RTL style (no preferred solution)
- Include all chars that could be added by shaping (https://github.com/tecnickcom/tc-lib-unicode-data/blob/develop/src/Arabic.php). This could be a good compromise but from a quick look it does not seem to work that well as the font subsetting is looking at the text instead of caring at the first place which chars the font provides. But we can include the possible substitutions whenever we hit a "source-glyph".
- Running tc-lib-unicode or a sub process of it as you mentioned

**(2) Always apply shaping**
Thought about this as well. But we maybe should do that if we are sure it works well. Also we might just enable processing it just by option.
On the other hand I don't know when and why shaping should have impact on non RTL text. It manipulates just a assessable number of glyphs.

**(3) Breaking multiline RTL text**
Took me some time to get the problem. [offtopic]It is quiet annoying to work on this with debugging tools as they switch are applying RTL direction to debugging state while it is off in the editor[/offtopic].

While writing an answer here I did a change which should work better:
https://github.com/dompdf/dompdf/pull/2107/commits/76803388963600e936a2a1464f66548914a0e353

I don't know how hard I thought about the place I set the bidi transform. I remember the idea was to be able to still do a split in case the text exceeds the limit just after the transform (I doubt that's really needed) but because the transform changes the order of the words the order gets broken if a line break happens.

simonberger:
Found the code comments suggesting the move of the font subsetting from rendering to output time.
I evaluated the change and for my knowledge of the code almost the whole logic of the `selectFont`method needs to move because of usage of the data probably coming from subsetting. But that should feasible.

bsweeney:
Regarding subsetting: So as you can see the whole reason we do subsetting the way we do is because the subset is built at the time the font is selected. I think the change would actually be fairly straightforward, but that's based on preliminary analysis. From what I see we would need to

1. Remove the font subsetting logic from Dompdf.cls.php and then within the CPDF class call registerText from the addText method.
2. In CPDF move the actual subsetting logic from the selectFont method to the "out" action of the o_font method.

There may be additional work, but that's what I see right now. We should also look at fixing up font subsetting support in PDFLib as I'm not sure we're doing it at all right now (I think all we need to do is set the global subsetting option).

This should probably be done in a separate branch. And we should probably do this for the next release if the totality of the work is as I've outlined above.

src/FrameReflower/Text.php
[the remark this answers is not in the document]
bsweeney: I think we can keep this initial release focused on the text itself. We can focus a future release on layout modification for rtl content. Conceptually this seems straightforward, just position from the right side of the container instead of the left. I think probably what will make it difficult is locating all the places where we need to add logic for this.

bsweeney:
With the most recent change to the code I'm seeing an odd character replacement with the sample I posted above using the Amiri font. Spaces are being replaced by exclamations when font subsetting is enabled. This isn't a global issue, either, because DejaVu Sans is rendered correctly.

I checked the develop branch and it's happening there as well. I must have missed this issue when I was reviewing the subset-focused pull request.

simonberger:
Works for me. Damaged font cache maybe?

sbc640964:
When will it be appropriate for Hebrew?

simonberger:
@sbc640964 It should work for Hebrew text.
If you test it please leave some feedback if it works well or if you have problems.
We have no language knowledge to really test it.

sbc640964:
@simonberger 
 I try unsuccessfully.
I am adding the

and this the document

```
<style>
  @font-face {
       font-family: 'Assistant';
       font-style: normal;
       font-weight: normal;
       src: url("the/path/to/ttf/file") format('truetype');
   }
</style>
<h2 class="text-info pb-5 font-weight-normal"  style="font-family: 'Assistant', sans-serif;">זו הכותרת בעברית</h2>
```
and this is the result
image

simonberger:
@sbc640964 
Something is wrong with your styles or your font file.

Here is a working sample with your text:

```
<html>
<head>
    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Assistant&display=swap">
</head>
<body>
    <div style="font-family: 'Assistant', sans-serif; direction: rtl;">זו הכותרת בעברית</div>
</body>
</html>
```

bsweeney:
@simonberger I think that text is _supposed_ to be rendered RTL due to strong directionality of the characters. This goes back to our discussion of enabling the bidi parsing at all time. This is a tweak I was looking at making against your branch but haven't gotten very far with (and we need to discuss more anyway).

@sbc640964 this branch is still a work-in-progress and we still have much to work out. If you set the direction of the text to RTL you'll see it rendered in the order expected. The directionality logic is only enabled when RTL is specified.

```
<html>
<head>
    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Assistant&display=swap">
</head>
<body>
    <div style="font-family: 'Assistant', sans-serif; direction: rtl;">זו הכותרת בעברית</div>
</body>
</html>
```

kory-northrop:
I have been testing out these commits on my staging site and am finding a similar issue about the RTL rendering improperly if the text string is too long to fit onto one line. This was mentioned by @bsweeney in their comment above (https://github.com/dompdf/dompdf/pull/2107#pullrequestreview-407969326).

Additionally, I'm finding that the <li> decoration is not appearing before the list item text (which _should_ be the right side of the page) instead it is staying to the left side of the page. Did anyone else have that issue?

learn05:
Solution is here [Click Here](https://github.com/dompdf/dompdf/issues/2841#issue-1188595460)

Magnatox:
Hello!
Did anyone find a solution for the first words rendered in a new line when the text is too long?

Magnatox:
> Found the best and simple trick to fix this, simply write : 
> 
> `<p dir="rtl" style="text-align: right; writing-mode: horizontal-tb; direction: rtl; font-family: XXX;">YOUR RTL TEXT</p>`
> 
> It fixes the issue with letters AND lines that are reversed

Hello @johnslm 
Can you confirm if it fixes this [issue](https://github.com/barryvdh/laravel-dompdf/issues/931#issuecomment-2259136135)?

Magnatox:
> Here is the full explanation :
> 
> 
> Add the following in the file located at vendor/dompdf/dompdf/src/Renderer/Text.php 
> 
> Above the line `$this->_canvas->text($x, $y, $text,`
> 
> Add:
> 
> ```
> if (strtolower($style->direction) == 'rtl') {
>     preg_match_all('/./us', $text, $ar);
>     $text = join('', array_reverse($ar[0]));
> }
> ```
> 
> In your text, add:
> `<p dir="rtl" style="text-align: right; writing-mode: horizontal-tb; direction: rtl; font-family: XXX;">YOUR RTL TEXT</p>`

Thank you for the explanation!

It would be better if we could fix this without touching the vendor files.

PR #2110, opened 2020-03-09
Fix warning: Illegal string offset 'hex' by adding ternary fallback to white

bassplayer7:


bsweeney:
Can you rebase your branch on the develop branch?

bsweeney:
Do you have some sample HTML + CSS that exhibits the issue?

Mellthas:
No HTML/CSS, but I can reproduce the warning with manually setting styles:

```
$style = new Style(new Stylesheet(new Dompdf()));
$style->border_top_color = "rgb(1,2,3,4,5)";
var_dump($style->get_border_top_color()); /* null */
var_dump($style->border_top_color); /* rgb(1,2,3,4,5) */
var_dump($style->get_border_top()); /* 1px none r, Warning: Illegal string offset 'hex' in ... */
var_dump($style->border_top); /* 1px none #000000FF */
```

So there are some inconsistencies to how the getters and property accessors work. `get_border_top()` calls `_get_border("top")` which returns the invalid color value, leading to the string-offset access on a string. `border_top` running through `__get()` somehow falls back to black for the color though. If you switch the order of the last two lines, the warning is gone, because the fallback black seems to be saved.

Mellthas:
So I think the commit as-is is probably not the right thing to do.

bsweeney:
Good analysis. I've been focusing on issues reproducible during rendering, but we should avoid issues like this one even in direct-usage scenarios.

Mellthas:
An example reproducing the issue is in https://github.com/dompdf/dompdf/issues/2367#issuecomment-776666583.

bsweeney:
I think the small change I made in #2536 works more within the expected return to address this issue.

Issue #2060, opened 2020-01-22
using setEncryption decreases compatibility


psyklon-project:
If I render the PDF without password, it renders correctly in Chrome, Firefox, Edge, and Acrobat Reader.

After adding this line right after $dompdf->render() and before $dompdf->output()

  $dompdf->get_canvas()->get_cpdf()->setEncryption($password, $password);

The file works properly in Chrome and Firefox. In Edge, after providing the password a blank page appears. Acrobat throws an Unknown error before showing the blank page.

Tested on both Windows, Linux, PHP 5.6, PHP 7.1 using latest version.
The Password on the attached  pdf is: 1993.06.08

Any ideas what should I do?

![screenshot](https://user-images.githubusercontent.com/35834994/72879473-3877f700-3cfd-11ea-8f62-4c8a29e08479.png)
[without-pass.pdf](https://github.com/dompdf/dompdf/files/4096075/without-pass.pdf)
[with-pass.pdf](https://github.com/dompdf/dompdf/files/4096076/with-pass.pdf)

bsweeney:
So ... uh ... what's the password to the protected PDF?

This appears to be an issue when using custom fonts and a protected PDF. If you stick to the core fonts things appear to be OK.

Maybe it's related to #1502?

psyklon-project:
> So ... uh ... what's the password to the protected PDF?
> 
> This appears to be an issue when using custom fonts and a protected PDF. If you stick to the core fonts things appear to be OK.
> 
> Maybe it's related to #1502?

The password is `1993.06.08`. The font used is DejaVu Sans, the one that comes with DomPDF.

bsweeney:
I can confirm this happens with DejaVu Sans as well. The core fonts would be ones that are included with PDF viewers (Helvetica, Times Roman, Courier). Using those, the PDF should render fine. Any other fonts (including DejaVu) are embedded in the PDF. The issue must be in how we encode the font information in a protected PDF. This may be related to 1502 or something additional, but we'll look into it.

PR #2114, opened 2020-03-12
Encryption fixes for Ordering and Registry in CIDSystemInfo dictionary

simonberger:
Addresses #2060


simonberger:
Kind of obvious problem after knowing but hard to find.
The problem of #2060 I could also reproduce is fixed and the encrypted file is valid in ghostscript and https://www.pdf-online.com/osa/validate.aspx

Issue #2111, opened 2020-03-10
Embed attachment to PDF file


tdjac0bs:
Hello, we would like to generate a PDF file with attached files inside it ( embeded ).

Is it possible to do with dompdf?

References:

- https://www.gnostice.com/nl_article.asp?id=193&t=Embed_Files_As_Attachments_To_PDF_Using_Java
- https://helpx.adobe.com/acrobat/using/links-attachments-pdfs.html#add_an_attachment

simonberger:
This is unfortunately first possible with PDF file version 1.7 and dompf implements 1.3.
It would be good to increase the version but we first need to eveluate which changes this needs on our side.

tdjac0bs:
Sounds really good @simonberger 

Is it a feature you see implemented in let's say, next 6 months? 

I ask because I need to choose the library to be used, so if you say it's in the roadmap we could go with dompdf and implement the embeded attachments later.

Thank you very much.

simonberger:
You can monitor progress and discussion on #2115 .
One of the next steps will be a team discussion if we do this 1.7 raise or not.
Dompdf has no constantly active progress looking at the last month and years which means you won't get a real estimation for features nor there is a roadmap.
But I am currently on some tasks that shall bring some new features to users and some life in dompdf. PDF 1.7 with new possibilities like file attachment is one of these.

To anwer your question: It is possible that the feature is feasible and coming in the next weeks and month without any promise. ;)

bsweeney:
FYI, this actually looks like it was first available in PDF 1.3 (section 3.10, same as in 1.7).

simonberger:
You are correct it looks like this is fully available in version 1.3. I found some version history logs saying that just in version 1.7 embedding all file types is allowed.
It was this one or similar.
https://www.pdfen.com/blog/attachments-in-a-pdf-a-what-is-allowed-and-what-is-not

But that is maybe wrong or just for PDF/A which we do not follow anyway.

sambenne:
Hi, I can see this is now in, but there does not seem to be away to actually use this method. How are you supposed to use it?

simonberger:
Hi @sambenne have a look at the pull request.
You can find a usage example there.

https://github.com/dompdf/dompdf/pull/2117

PR #2117, opened 2020-03-16
Multiple files can be embedded in a PDF

simonberger:
Addresses #2111 

This pull requests adds the embedded files feature to PDFs. Any type of files can be added.
It adds no comment mark or similar to the PDF. It is instead up to the reading program how to view the list of files.

Here a screenshot of how it is looking in Adobe Reader for example.

screenshot1

Some implementation details:
- There is no created / updated date for simplification reasons. They could easily be added (or we just use the current time).
- Just local files can be added to the PDF. No data strings. If this is a use case temporary files would have to be created. I like it that way but am open to different opinions.
- It adds 4 new PDF object to achieve the functionality.

Usage:
```
$dompdf = new \Dompdf\Dompdf();
$dompdf->loadHtml();
$dompdf->render();
/** @var \Dompdf\Cpdf $cpdf */
$cpdf = $dompdf->get_canvas()->get_cpdf();
$cpdf->addEmbeddedFile(
    'path_to_files_to_include/2100.html',
    '2100.html',
    'hello world !'
);
$cpdf->addEmbeddedFile(
    'path_to_files_to_include/circle_300px.png',
    'circle_300px.png',
    'hello circle !!'
);
$dompdf->output();
```

I did not think further if this should be available from Dompdf directly.

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -63,6 +63,16 @@ class Cpdf
     public $catalogId;
 
     /**
+     * @var integer The objectId (number within the objects array) of indirect references (Javascript EmbeddedFiles)
+     */
+    protected $indirectReferenceId = 0;
+
+    /**
+     * @var integer The objectId (number within the objects array)
+     */
+    protected $embeddedFilesId = 0;
+
+    /**
      * AcroForm objectId
      *
      * @var integer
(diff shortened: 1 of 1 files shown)


simonberger:
@bsweeney if you want want to review this for the 0.8.6 I plan to include it otherwise it can also wait. The feature itself is completely optional and just be usable via cpdf at the moment which I would not change for now. But the catalog references changed which were inlined and used for javascript. I had no idea how to test this in dompdf.

bsweeney:
I'm fine including this. I remember looking at the changes and don't recall having any issues.

Issue #2122, opened 2020-03-25
Nested span & bg-color bug


Nuranto:
HTML to reproduce : 

```
<span style="background-color:red">
            toto
            <span style="background-color:green">tata</span>
            toto
            <span style="background-color:green">tata</span>
</span>
```
HTML result : 
![Capture d’écran 2020-03-25 à 16 22 42](https://user-images.githubusercontent.com/1633463/77553578-455bc700-6eb5-11ea-9dc7-ebe306f3d7a5.png)

PDF result : 
![Capture d’écran 2020-03-25 à 16 22 29](https://user-images.githubusercontent.com/1633463/77553387-09c0fd00-6eb5-11ea-95a2-da575a528cdb.png)

As you can see, there is a part where the background is missing (or wrongly white).
There is also a background vertical align issue, but that's less important.

simonberger:
I'll have a look. 
In a first try I just tested how an outer inline div changes the result. With some negative padding we get the correct result.

```
<div style="padding-bottom: -2px; display: inline-block; background-color:red">
    toto
    <span style="background-color:green">tata</span>
    toto
    <span style="background-color:green">tata</span>
</div>
```

Nuranto:
Hello @simonberger 
I've just tested your fix in my real case, and it does fix the issue, thank you !

(For the v-align issue, I cannot edit the html in my case, but that's not an important issue.)

simonberger:
@Nuranto Thank you for confirming.
Regarding the vertical aligning problem you mentioned. First to note I am by far no CSS expert. But the background color moves with the text no matter how it is vertically aligned in its container.
I am a somewhat surprised by the text centering behavior. Dompdf renders it the way I expected it to be correct. It is the baseline position.
But it is indeed a low low priority problem I won't debug at this moment.

PR #2123, opened 2020-03-25
Fix nested inline dimensions

simonberger:
Addresses #2122


simonberger:
Adding tests on this class render function is unfortunately pointless.

Issue #2124, opened 2020-03-26
`white-space:pre-wrap` fails in two newline scenarios 


sakihl:
I think this partially overlaps with  [1315](https://github.com/dompdf/dompdf/issues/1315).
I am using PHP 7.2.28 and dompdf 0.8 (loaded by barryvdh/laravel-dompdf).
The problems are:
1. If there is a long line _without a newline_ at the end, it does not wrap.
2. If there are tags (e.g. bold) in a line _with a newline_ at the end, the newline is inserted before the part of the line after the closing tag, rather than at the end.

[ A third scenario came up later in the thread see [comment below](https://github.com/dompdf/dompdf/issues/2124#issuecomment-606063177).
3. Blank lines created by extra newlines not preserved ]

Example HTML (the monospace part is irrelevant to this issue - it makes it easier to check examples with leading spaces):

```html
<style>
	p.monospace-pre-wrap {
        font-family: Consolas, Courier, monospace;
        white-space: pre-wrap;
      }
</style>
<p class="monospace-pre-wrap">This is a long line with no newline at the end. Pre-wrap setting on the paragraph means that this line should wrap.</p>
<p class="monospace-pre-wrap">This line has <b>bold tags</b> and text afterwards ending in newline.
Another line.
</p>
```
In the PDF the long line just continues off the page without wrapping. The line with bold tags inserts the newline directly after the closing `</b>` tag instead of at the end.
![image](https://user-images.githubusercontent.com/62707603/77662338-a649d480-6f73-11ea-957d-9b134113981e.png)

I think both problems are in [FrameReflower/Text.php](https://github.com/dompdf/dompdf/blob/master/src/FrameReflower/Text.php)

In function `_layout_line()`, this piece of code (starting line 239) 
```php
case "pre-wrap":
                $split = $this->_newline_break($text);

                if (($tmp = $this->_line_break($text)) !== false) {
                    $add_line = $split < $tmp;
                    $split = min($tmp, $split);
                } else
                    $add_line = true;
```
does not handle the case where the line needs to wrap but there is no newline, so `$split` is false but `$tmp` is not.
I think it needs to be replaced with
```php
                $split = $this->_newline_break($text);
                if (($tmp = $this->_line_break($text)) !== false) {
                    if ($split !== false) {
                        $add_line = $split < $tmp;
                        $split = min($tmp, $split);
                    } else {
                        $split = $tmp;
                    }
                } else {                   
                    if ($split !== false) {
[...]

Issue #1315, opened 2016-11-24
`white-space:pre-line` not working correctly


ThomasLandauer:
This might be related to https://github.com/dompdf/dompdf/issues/800

Sample code:
```html
<style>#foo {white-space:pre-line; }</style>
Zero
<div id="foo">One
Two
Three</div>
Four
```

Expected result (as shown in Firefox & Chrome):
```
Zero
One
Two
Three
Four
```

Actual result in dompdf 0.8.1:
```
Zero

One
TwoThree
Four
```

nickwuk:
Also having issues with 'white-space: pre-line' when specified for a textarea inside a table cell. I have an example table with two columns and several rows, each row in the second column contains the same 2 paragraph content. When output to Pdf the 2nd paragraph starts on the next line, the space between the paragraphs is removed (I assume only the 1st newline character is preserved and the 2nd new line character is removed), and the 2nd paragraph is truncated to one line - because it looks like it is not wrapping.

This behaviour continues in the 2nd row, in rows 3-5 content is not truncated, then content in the last row #6 is again truncated. The extra line space between paragraphs is removed in all rows.

With 'white-space:  normal' all content is visible but there's no paragraphing, all new lines are removed.

bsweeney:
Not much progress with anything for the last few months, but I'm going to try to get back into things.

bsweeney:
I haven't looked at this yet, just a random thought ... perhaps better handled in the line box logic?

tuananhcwrs:
I've got the same issue as @ThomasLandauer .
Would someone please share any tip?

bsweeney:
specific case appears to be addressed with committed changes

PR #2125, opened 2020-03-27
Fix layout_line add_line

simonberger:
Addresses:
#2124 
#1315

### DIFF
--- a/src/FrameDecorator/Block.php
+++ b/src/FrameDecorator/Block.php
@@ -156,7 +156,7 @@ class Block extends AbstractFrameDecorator
 
         // FIXME: Why? Doesn't quite seem to be the correct thing to do,
         // but does appear to be necessary. Hack to handle wrapped white space?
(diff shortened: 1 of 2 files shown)


simonberger:
This has a little BC if _layout_line is not considered as an internal method.

sakihl:
Hi @simonberger. So far I have just been using this fix locally on a PC but I need it for a project going live in September. Before creating my own branch and setting up my project's composer to pull from there I just thought I would ask the question first. Do you know what the approximate timescale for 0.8.6 release might be?

bsweeney:
We have a few issues left to address in the current code Base for 0.8.6. Trying to push through that so we can get the release out since there are some issues introduced in 0.8.5 that are addressed in this release.

I would expect it before September, but I can’t give a hard deadline just yet.

sakihl:
Thank you for your quick and helpful response.

PR #2127, opened 2020-03-27
Images scaling to container size without specific dimension (develop branch)

simonberger:
A regression introduced by #2108 scales images without a width/height value to the full container size.
While fixing it I now did a change I wanted to initially do but was afraid of. There were some ugly 0 value handling which were not really understandable.
I now changed it to value auto. I need some more experience here as I'm not sure which values we would need/want to skip.

The fix currently is incomplete (has a undefined side effect) because of the width/height of 0 on the image tag which now will be forwarded while it had been reset to the image dimensions before.
Oddly the image is still rendered (by CPdf) but in a wrong position.
In my opinion we should remove those images from the PDF.

### DIFF
--- a/src/FrameReflower/Image.php
+++ b/src/FrameReflower/Image.php
@@ -117,7 +117,7 @@ class Image extends AbstractFrameReflower
             $style->max_height !== "none"
         ) {
 
-            list($w, $h) = $frame->get_containing_block();
+            list( /*$x*/, /*$y*/, $w, $h) = $frame->get_containing_block();
 
             $min_width = $style->length_in_pt($style->min_width, $w);
             $max_width = $style->length_in_pt($style->max_width, $w);
--- a/src/FrameReflower/Image.php
+++ b/src/FrameReflower/Image.php
@@ -58,21 +58,23 @@ class Image extends AbstractFrameReflower
      */
     function get_min_max_width()
     {
+        $frame = $this->_frame;
+
         if ($this->get_dompdf()->getOptions()->getDebugPng()) {
             // Determine the image's size. Time consuming. Only when really needed?
-            list($img_width, $img_height) = Helpers::dompdf_getimagesize($this->_frame->get_image_url(), $this->get_dompdf()->getHttpContext());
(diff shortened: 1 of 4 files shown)


simonberger:
In my tests skipping worked like this:

```
if ($w === 0.0 || $h === 0.0) {
     return;
}
```

After line https://github.com/simonberger/dompdf/blob/865649d3af8dc1280367fe5ab0f2dbb014357246/src/Renderer/Image.php#L31

simonberger:
Added this change

src/FrameReflower/Image.php
```diff
@@ -115,7 +117,7 @@ function get_min_max_width()
             $style->max_height !== "none"
         ) {
 
-            list($w, $h) = $this->_frame->get_containing_block();
+            list($w, $h) = $frame->get_containing_block();
```
bsweeney, line 120: In #2108 this was changed from:
```
list( /*$x*/, /*$y*/, $w, $h) = $this->_frame->get_containing_block();
```
to:
```
list($w, $h) = $this->_frame->get_containing_block();
```

Does the list portion need to be `list(, , $w, $h)` or `list('w' => $w, 'h' => $h)` instead?

bsweeney:
You are probably correct that an image with a width/height of 0 should probably just be removed.

bsweeney:
I think the issue I'm experiencing with that border radius sample is unrelated to this change. I thought I wasn't seeing it outside this branch but I believe I'm wrong. Perhaps best to address in a separate PR.

src/FrameReflower/Image.php
simonberger: This indeed should be a bug for that case. I'll fix that tomorrow and have a look why the tests fail to cover that.

simonberger: I fixed it with reverting the change. The tests are not good enough in this area to cover it. My change not make a difference. This is caused by too many necessary mock objects returning a new value inside the method.. It could be improved though with a lot of `shouldReceive->with` parameter checks.
On the other hand it could be a better effort to improve the structure of those classes.

simonberger:
@bsweeney Is ready to go including the other fix or did I miss something you found?

PR #2133, opened 2020-04-02
Update CSS getArray

jjsquady:
Fixes deprecation message at PHP 7.4+
**Invalid characters passed for attempted conversion, these have been ignored**

### DIFF
--- a/src/Css/Color.php
+++ b/src/Css/Color.php
@@ -296,6 +296,7 @@ class Color
             $c["alpha"] = $alpha;
             $c["hex"] = "cmyk($c[0],$c[1],$c[2],$c[3])";
         } else {
+            $color = preg_replace("/[^0-9A-Fa-f]/", "", $color);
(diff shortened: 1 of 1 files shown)


src/Css/Color.php
```diff
@@ -296,6 +296,7 @@ static function getArray($color, $alpha = 1.0)
             $c["alpha"] = $alpha;
             $c["hex"] = "cmyk($c[0],$c[1],$c[2],$c[3])";
         } else {
+            $color = preg_replace("/[^0-9A-Fa-f]/", "", $color);
```
bsweeney, line 299: This looks to be insufficient. This addresses the original error, but introduces a new one by potentially shrinking the size of the hex string so that the `mb_substr` calls result in an "illegal string offset" error.

Mellthas:
This is properly resolved with commit 236c672d5ecb9bf8763fe3a3eefcabe7c47f2568.

PR #2135, opened 2020-04-12
Merge Branch: Rtl language support and computed css

simonberger:


simonberger:
@bsweeney I tested your computed props changes and tried to get the text align right working via `direction:rtl` as well as direction:right but a `text-align: left` on a parent element. It is the same scenario I had some weeks before.

I did not manage to get both scenarios working without code changes

#1 with text-align inheritance
```
<body style="text-align: left"">
	<div style="direction: rtl"><p>
		test
	</p></div>
</body>
```

#2 without
```
<body>
	<div style="direction: rtl"><p>
		test
	</p></div>
</body>
```

If I try to set the text-align: right in the set_direction method Scenario 1 works but inheritance will be destroyed as this will be called later.

If I instead try to return text-align right in a new `get_text_align`method when direction right is set and no inherited or directly set text-align is there I have no chance to detect it. I do not know how a set_text_align method could be created to avoid those problems. 
Summed up I have the same problems as before. ;(

bsweeney:
I reviewed the status of the code and realized I had a missed a few issues:
- when I moved the letter/word spacing logic from the getter to the setter I didn't quite fill out the logic correctly
- need to validate that the passed-in value for text-align is a valid keyword
- text-align is dependent on direction and so needed to specify that relationship in the dependency map

Try with the current rev of that branch and see if it helps.

simonberger:
Ok then there seems no other way around it. Yes it works like this.

bsweeney:
Thoughts on how to proceed then? Complete review on this PR and merge, close the other PRs?

simonberger:
I'd like to close this merge request. Instead my plan is to wait until your changes can be merged into develop and merge develop into the RTL branch where I repeat the necessary small changes.

bsweeney:
OK, I've merged the computed property branch.

simonberger:
Ok I'll finalize the merge request next week. Beside some minor review comments I found an endless recursion problem in your merge request. Did you miss it or do you think it is not of much practical impact? I got it when taking the HTML of page encode.su.

bsweeney:
ah ha ha that's why those few comments were on an older commit. Anyway, I responded, take a look and let me know if you have any more thoughts. I'll commit a few changes and a new PR once we resolve current discussions.

simonberger:
It was good. Otherwise I maybe never would have noticed my fail and you wait for my approval and I wait the fixes. ;)

simonberger:
I'll wait your changes before I merge develop.

simonberger:
Closed as discussed. It was only a test branch.

PR #2138, opened 2020-04-19
silence iconv

dedobbin:
I had trouble using this (on Alpine Linux). From what i can gather it's related to the slashes in the param of the iconv call, so i silenced it. Not the neatest fix, but the same approach is used in InputStream.php, with a comment saying "non-conforming". It could indicate this gave the same problem i ran into?

### DIFF
--- a/src/FontMetrics.php
+++ b/src/FontMetrics.php
@@ -187,7 +187,7 @@ class FontMetrics
         $prefix = preg_replace("/[^\\pL\d]+/u", "-", $prefix);
         $prefix = trim($prefix, "-");
         if (function_exists('iconv')) {
-            $prefix = iconv('utf-8', 'us-ascii//TRANSLIT', $prefix);
+            $prefix = @iconv('utf-8', 'us-ascii//TRANSLIT', $prefix);
         }
         $prefix = preg_replace("/[^-\w]+/", "", $prefix);
         


bsweeney:
Based on a quick search this seems to be an issue people run into with Alpine Linux. If you replace `//TRANSLIT` with `//IGNORE` do you get the same error? Transliteration would be ideal, but I think ignoring those characters would be fine as well since all we're doing here is trying to make the persisted font file more findable.

dedobbin:
Hey, sorry for the late reply.
I tried and same result. Just using 'us-ascii' is fine. // is the problem.

keksa:
Hey @dedobbin, we had the same issue, but fixed by using "proper" iconv in alpine, see https://github.com/docker-library/php/issues/240#issuecomment-506651614

bsweeney:
I like transliterating, but right now this is the only place where Dompdf is actively using iconv (though there are some dependencies that rely on it). Maybe we can, instead, rewrite the logic to just convert the encoding. The usage here is internal, we're keeping the filename just to make debugging a bit easier. So maybe instead of the current logic we do the following:

```
mb_substitute_character("_");
$prefix = mb_convert_encoding($prefix, 'ISO-8859-1', 'UTF-8');
$prefix = preg_replace("[\W]", "_", $prefix)
```

PR #2141, opened 2020-04-30
Minor logic updates to the Style class based on feedback

bsweeney:


bsweeney:
@simonberger all done here

PR #2142, opened 2020-05-02
Prevent infinite loop when re-calculating dependent style properties

bsweeney:


bsweeney:
@simonberger this should address that last issue you pointed out. I'll play it smart this time and wait for you to review.

Issue #750, opened 2013-12-16
Generated content missing in font subset


Paktas:
I've recently setup DOMPDF_ENABLE_FONTSUBSETTING and installed/copied fresh php-font-lib to dompdf/lib/php-font-lib/.
It looks to have decreased the PDF file size (6 times), however some characters didn't end up in the subset.
I am using &lt;ol&gt;&lt;li&gt; entries styled to bold together with counter on CSS:

```
OL { counter-reset: item;}
OL LI { display: block; font-weight:bold; }
OL LI:before { content: counters(item, ".") ". "; counter-increment: item }
```

Seems like the prefixed numbers for &lt;li&gt; that are created with CSS counter do not end up in the font subset. It's showing uknown symbol for that bold-missing character in generated PDF.

For the testing purpose manually adding some of the missing numbers somewhere in html/text in bold - fixes the issue, counter generated list numbers AND entered manually in the body show up fine in li's inside pdf (as per screenshot below).
![numbers on counter](https://f.cloud.github.com/assets/6200695/1759403/e0ecca10-669d-11e3-84e2-177d72684294.JPG)
Workaround is to put bold numbers 0 through 9 in hidden span, but this doesn't look right.

Removing the subsetting option in dompdf - all characters work like a charm (and generates a big pdf file as well).

Hope this can be fixed on the road to better dompdf font subsetting.

NOTE: tested on PHP 5.2.4, linux, newest github versions of dompdf and php-font-lib.

bsweeney:
Additional improvements to content-declared text was added in commit 27ce2db2.

MarkMaldaba:
What's the status of this ticket.  It is marked as 'On Deck' and was slated for 0.8.1, but that was over 2 years ago.

I can see there is some code relating to this issue, present in 0.8.3.  I can also see that the `page_text()` function (and `text()`, presumably) don't have their characters included in the font subsetting yet.

What is needed to resolve this ticket (and in particular to get it working for `text()`/`page_text()`, which are the main issues for my use-cases)?  Is there anything I or anyone else can do to help?

bsweeney:
Status is that it is still on deck, and will remain so until someone submits a pull request or someone on the team (pretty much just me right now) is able to work on the issue.

[...]

PR #2149, opened 2020-05-10
Move font subsetting from render to output

simonberger:
The major part of the font processing is now triggered from the output instead of the render step.
The needed chars are collected from text output calls to CPDF instead of walking the frame tree another time. This also moves all work to CPDF.

Todos:

- [x] Testing
   - [x] This is covering decimal list bullets
   - [x] Multiple embedded fonts are working
   - It is working with arabic font shaping of #2107 - No reason for it to not work
- [x] Just do font subsetting when the option is activated
   - Maybe we want to activate it by default now as the performance overhead should be low.

Addresses: #750

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -874,7 +874,7 @@ class Cpdf
                 break;
 
             case 'add':
-                $font_options = $this->process_font($id, $o['info']);
+                $font_options = $this->processFont($id, $o['info']);
 
                 if ($font_options !== false) {
                     foreach ($font_options as $k => $v) {
@@ -983,7 +983,7 @@ class Cpdf
      * @return array|false
      * @throws FontNotFoundException
      */
-    private function process_font(int $fontObjId, array $object_info)
+    private function processFont(int $fontObjId, array $object_info)
     {
         $fontFileName = $object_info['fontFileName'];
         if (!isset($this->fonts[$fontFileName])) {
@@ -1108,8 +1108,7 @@ class Cpdf
             // note that pdf supports only binary format type 1 font files, though there is a
             // simple utility to convert them from pfa to pfb.
             if (
-                !$this->isUnicode
-                || strtolower($fbtype) !== 'ttf'
+                strtolower($fbtype) !== 'ttf'
                 || !$font['isSubsetting']
             ) {
                 $data = file_get_contents($fbfile);
@@ -3020,7 +3019,7 @@ EOT;
 
                 $this->o_font($this->numObj, 'new', $options);
                 $font['fontNum'] = $this->numFonts;
-                $font['isSubsetting'] = $isSubsetting;
+                $font['isSubsetting'] = $isSubsetting && !$font['isUnicode'];
 
                 // also set the differences here, note that this means that these will take effect only the
                 //first time that a font is selected, else they are ignored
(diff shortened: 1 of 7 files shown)


simonberger:
@bsweeney please have a look at this. It worked well in short examples. But I had no run with multiple fonts and I might overlook things I don't know yet. ;)
Other limitations are listed above.

lib/Cpdf.php
```diff
@@ -2329,6 +2621,13 @@ function output($debug = false)
         $content = '%PDF-1.7';
         $pos = mb_strlen($content, '8bit');
 
+        // pre-process o_font objects before output of all objects
```
simonberger, line 2624: Related to your comment https://github.com/dompdf/dompdf/pull/2107#issuecomment-626379294
I worked on this after my previous comment and my first idea was also to move it to o_font out action directly.

I decided to take this little extra effort to keep the add step of o_font. The code modifies and adds new objects.
While it should be safe to assume that all other related objects are processed after o_font it is safer and clearer this way.

bsweeney: My original thought was to process the full font without subsetting, then perform the subsetting during "out" (so "add" would have still been part of it). That way the full font information would be available right up until the PDF is generated. That would probably require a bit of extra processing but I didn't really have a chance to plan out that change fully.

Your approach isn't too far off from that, so I think it's good. I only have two potential concerns, mainly because I haven't fully analyzed the change yet:

1. Are there any internal objects that rely on the font information that is, with this change, now being generated only when the PDF is output?
2. Is there any reason a user of the library would be using that font information that will run into issues with this change?

If we're comfortable with the answer to those questions them I think this will work.

simonberger: Answers to your questions:

1. You are correct we need to analyze if any of the later steps is needed in CPdf. 
I have not too many doubts because OpenFont is still called and the font object is almost not changed in the following lines. From a quicker look just some information is added which is needed in the PDF and it also builds for example the fontDescriptor object. But I absolutely did not check this intensively enough.
2. Very much relates to 1. The main objects that are changing later are just the objects added to the pdf. The user should not want to read them but might write it which could be overwritten potentially.

I think it does not change much if we would double run the major part of this code block to have the font data in advance. If a user would interact with them he would read or write inconsistent data.
The main focus is on the data coming from openFont function and this is still processed early.

bsweeney: I think we're OK with internal use. For external ... I guess we'll see.

bsweeney:
Once we finish up with this I'll return to the RTL branch so we can finish that up.

simonberger:
Your requested changes are making sense. I'll implement them.

simonberger:
In detail I did the following changes which should cover all your points:
- selectFont gets an additional parameter isSubsetting which is true by default.
   - The calls coming from the adapter are reading the dompdf option.
   - The calls from other CPDF methods selecting the defaultFont if none is available now have enabled subsetting.
- Collecting subsetting data happens directly in CPDF addText instead of the adpater.
 - A tag SUB[A-Z][A-Z][A-Z]+ is added in front of the font name. It is increased by fontnum starting with SUBAAB and has plenty of space (26^3).
- The dompdf subsetting option is now TRUE by default.

simonberger:
Unicode and file suffix detection are now taken into account early to never collect text data unnecessarily.

PR #2178, opened 2020-06-22
Resolved error in PHP 7.4 compatibility

ravikant-tcs:
phpcs -p --standard=PHPCompatibility --runtime-set testVersion 7.4 dompdf

FILE: /Users/ravikantmane/Desktop/patch/dompdf/src/Adapter/CPDF.php
FOUND 1 ERROR AFFECTING 1 LINE
198 | ERROR | Extension 'cpdf' is removed since PHP 5.1; Use pecl/pdflib instead

### DIFF
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -195,7 +195,8 @@ class CPDF implements Canvas
 
         $this->_dompdf = $dompdf;
 
-        $this->_pdf = new \Dompdf\Cpdf(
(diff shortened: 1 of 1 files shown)


bsweeney:
This seems more like an issue with phpcs than a PHP 7.4 compatibility issue. We have appropriately namespaced our version of CPDF, I wouldn't expect that line's presence to be detected as a reference to the deprecated CPDF extension.

PR #2191, opened 2020-07-06
Develop Index creation with page number / Create a table of content (summary) 

cicoasm:
I wish I could create an index with page numbers.
I tried this: http://eclecticgeek.com/dompdf/debug.php?identifier=toc-v0.6.x but it seems not to work with the latest version.


bsweeney:
I think you meant to open an issue not a pull request.

PR #2193, opened 2020-07-07
Add `loadDOM` method

aduh95:
Gives the opportunity to users using a `DOMDocument` object to skip the string conversion and re-parsing steps.

In my use case, I am building an HTML document from several sources, and I am using the PHP DOM API to do so. I figure it would be way more efficient to pass the `DOMDocument` instance straight to DOMPDF rather than passing a string that needs parsing to turn back to a `DOMDocument`.

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -415,6 +415,24 @@ class Dompdf
         $this->loadHtml($str, $encoding);
     }
 
+    public function loadDOM($doc, $quirksmode = false) {
+        $this->saveLocale();
+
+        // Remove #text children nodes in nodes that shouldn't have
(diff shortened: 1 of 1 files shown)


src/Dompdf.php
```diff
@@ -415,6 +415,24 @@ public function load_html($str, $encoding = null)
         $this->loadHtml($str, $encoding);
     }
 
+    public function loadDOM($doc, $quirksmode = false) {
+        $this->saveLocale();
```
bsweeney, line 419: Should we save the locale as part of this method? I don't think it's necessary.

aduh95: You're right, that doesn't seem to be necessary

PR #2217, opened 2020-08-28
Change width measurements to not be case-sensitive

eddturtle:
I don't *think* there's any requirement for the width scale (like 'px' after the unit) to be lower-case in the spec. Web browsers seem to treat it them same too. But in dompdf it will ignore it if it's written in uppercase.

I couldn't see an issue for this raised, so thought I'd open a PR.

```html
<!doctype html>
<html>
<head>
  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
  <title>0px example</title>
</head>
<body>
  <p style="width:150PX; border: 1px solid red;">example in caps</p>
  <p style="width:150px; border: 1px solid red;">example in lower-case</p>
</body>
</html>
```

Example 
Screenshot - download

I tested this against the `develop` branch.

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -620,23 +620,23 @@ class Style
                 continue;
             }
 
-            if (($i = mb_strpos($l, "px")) !== false) {
+            if (($i = mb_stripos($l, "px")) !== false) {
                 $dpi = $this->_stylesheet->get_dompdf()->getOptions()->getDpi();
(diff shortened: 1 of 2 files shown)


bsweeney:
I think this is the correct interpretation. I didn't do extensive research to see what guidance is in the various specs, but at the very least the CSS3 spec indicates that "unit identifiers are ASCII case-insensitive"
https://www.w3.org/TR/css-values-3/#dimensions

PR #2227, opened 2020-09-06
fix Trying to access array offset on value of type null

xerron:
Fix to correct generate svg

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -3834,9 +3834,13 @@ EOT;
             $mode = "Normal";
         }
 
-        if ($mode === $this->currentFillTransparency["mode"] &&
-            $opacity == $this->currentFillTransparency["opacity"]
-        ) {
+        if (is_null($this->currentFillTransparency)) {
+            $this->currentFillTransparency = [];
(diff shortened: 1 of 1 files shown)


feltoar:
Thank you for this. Waiting on PR to be merged

bsweeney:
Well, this is ... interesting. Nowhere in the Dompdf library does this property ever get set to null so this change should not be necessary. Not to say that a user couldn't set the property to null since it's public, so as a safety measure I see no harm including this change.

Similar logic is used by the `setLineTransparency` method so we should probably update that method as well.

As far as where this might be happening, php-svg-lib implements its own CPDF class and in that library the value of currentLineTransparency  and currentFillTransparency is [cleared out after being set](https://github.com/PhenX/php-svg-lib/blob/v0.3.3/src/Svg/Surface/SurfaceCpdf.php#L392-L411). I'm not sure why that library is clearing the property since it acts as a cache to prevent excessively writing out the fill instruction to the PDF.

With all this in mind, is the error you're seeing reported from Dompdf\Cpdf or Svg\Surface\Cpdf?

PR #2232, opened 2020-09-09
Modify the chroot option to be a string array

timmit-nl:
This is a fix for the chroot setting introduced by 0.8.6. As it was a breaking release because we use local resources on multiple location OUTSIDE of the dompdf dir.

We don't like to use a insecure and wide "/" but we have multiple location for our assets. For example:
/glusterfs/groups/images/
/public_html/assets/
/public_html/styles/

So we like to feed the chroot multiple locations. This is the fix that allows that.
If you like the chroot to be an array, then we suggest to make an extra option like onlyFromThisDirs that will have an array containing at default [$this->chroot] but can be set to "['/glusterfs/groups/images/','/public_html/assets/','/public_html/styles/']". So we can use the insecure chroot "/" but whit a secure limitations to specific dirs.


bsweeney (approved):

bsweeney:
I think this is a solid change and I'm good moving forward. I changed the base of the PR to develop since we try to avoid committing directly to master. I'll tweak the 0.8.7 milestone tasks so we can get this out soon.

timmit-nl:
> I changed the base of the PR to develop since we try to avoid committing directly to master
ahhh offcourse sorry that I missed that. On our own projects we do the same!

bsweeney:
Should we allow both an array or a string for setChroot so that upgrading to 0.8.7 doesn't fail horribly? If it's a string just put the string into an array and set the option?

timmit-nl:
yes that is a great idea. so 
```
if(is_string($value)){
    $value = array($value);
}
```

timmit-nl:
(do I need to make the change?)

bsweeney:
You don't _have_ to make it, no. But it would be appreciated 🙏

timmit-nl:
> You don't _have_ to make it, no. But it would be appreciated

I asked just that we don't do the same ;-)

I have updated it and correct an error. But I see one diffrence in src/Dompdf.php and for example src/Ccs/Stylesheet.php

In the old code it says in src/Dompdf.php:
```
$chroot = realpath($this->options->getChroot());
if ($chroot && strpos($realfile, $chroot) !== 0) {
```

but in src/Ccs/Stylesheet.php:
```
$chroot = realpath($options->getChroot());
if (!$chroot || strpos($realfile, $chroot) !== 0) {
```
I think it has to be everywhere:
```
$chroot = realpath($this->options->getChroot());
if ($chroot!==false && strpos($realfile, $chroot) !== 0) {
```
So it fails only if it can access the $chroot dir.

bsweeney:
I agree with your analysis of how we should check that realpath returned a valid path, `!== false` would be the more correct condition.

timmit-nl:
so also the
```
$chroot = realpath($this->options->getChroot());
if ($chroot && strpos($realfile, $chroot) !== 0) {
```
in src/Dompdf.php that was the only one that was diffrent from the rest?

So all the we change all the checks to
```
$chroot = realpath($this->options->getChroot());
if ($chroot!==false && strpos($realfile, $chroot) !== 0) {
```
?

I can't figure out why it is diffrent in src/Dompdf.php that's why I ask ;-)

bsweeney:
I think that was just an oversight on my part. When I was working on the update I referenced the original logic but didn't do much to change it. So, yes, I think the logic should be made consistent in all places.

Looking at it closer, the intention of the new logic was to say, deny access to the resource if either condition is met:
- realpath is unable to parse the value of chroot (i.e. it has a bad value)
- the resource reference does not start with the realpath-parsed value of chroot

With the logic in pace in Dompdf.php we only verify the file path when realpath is valid. So, specifying an invalid value for the option is a way to bypass the check.

With that in mind I think the logic should be:
```
if ($chroot === false || strpos($realfile, $chroot) !== 0) { ... }
```

timmit-nl:
ok I have changed the logic as you suggested it!

bsweeney (approved):

williampaulo:
Hello, I was checking the reason for adding an image with a local file (/var/www/site/img.png) and configuring correctly, an error was still occurring and returning the message "Image not found or type unknown", so debugging the code I found the following in the file "src / Image / Cache.php":

line 129
```
if (strpos($realfile, $rootDir) !== 0) {
```
where according to the PHP documentation, https://www.php.net/manual/en/function.strpos.php paying attention to the "Warning" note, the correct thing would be:
```
if (strpos($realfile, $rootDir) !== false) {
```

I ask you to check and if so, apply the correction in this PR.

timmit-nl:
@williampaulo you are totaly right! I will fix it.

bsweeney:
@williampaulo we're using strict comparison so the issue should not be that we're checking that resource URL begins with the chroot setting. The warning from the documentation applies when you're checking without strict comparison (allowing type juggling), i.e. `!= 0`. Are you sure that this is the cause of your issue?

Also, your suggested change would allow the resource reference to pass so long as `realfile` returns it intact, and so long as the path includes the chroot path anywhere in the string not just at the beginning. I don't have time right now to fully investigate the security implication with that change, but I do believe that change would overall be less secure than what is suggested in this PR.

And, just to reiterate what's going on in that block, the codes says: if the file is not under the dompdf root directory (e.g. includes/html.css), then make sure it is within the chroot.

timmit-nl:
Ok I have also checked this AND I see that my reaction was wrong and a possible security issue:

`if (!$chroot || strpos($realfile, $chroot) !== false) {`
is true if chroot is in the string of $realfile
so chroot of /tmp gives:
`/tmp/file.jpg => true`
`/root/tmp/file.jpg => true`
So this in unwanted behavior!

`if (!$chroot || strpos($realfile, $chroot) !== 0) {`
gives an true if the $chroot is ONLY AT THE FIRST SEGMENT of the string
so chroot of /tmp gives:
`/tmp/file.jpg => true`
`/root/tmp/file.jpg => false`

And this is correct in my opinion.

bsweeney (approved):

Issue #2239, opened 2020-09-16
Dotted border


c-louis:
Hello

I currently have a border issue when using Dompdf, I'm suppose to make some red dotted borders in the PDF but I realized that dotted in Dompdf are square ?

That's what I'm expecting :  

![Capture d’écran 2020-09-16 à 12 04 51](https://user-images.githubusercontent.com/39031067/93323631-265f4180-f815-11ea-91ad-2f987cc18375.png)

Using : 

`td.dotted { border-style: dotted; color: maroon;}`

From [http://eclecticgeek.com/dompdf/debug.php?identifier=434d7c2a19e907aacf7f72b4d889d318](http://eclecticgeek.com/dompdf/debug.php?identifier=434d7c2a19e907aacf7f72b4d889d318)

And that's what I'm getting from Dompdf : 

![Capture d’écran 2020-09-16 à 12 05 10](https://user-images.githubusercontent.com/39031067/93323709-3ecf5c00-f815-11ea-87ae-bd653df0c949.png)

Is that a choice or just circle for dotted border are not yet implemented ?

Thank you very much for your appreciated support.
Louis

bsweeney:
The current implementation of dotted borders is a variation of the dashed border ([ref](https://github.com/dompdf/dompdf/blob/v0.8.6/src/Renderer/AbstractRenderer.php#L473)). So right now I guess you could consider it a partially implemented.

I'm not sure at this time how much work would be required to implement actual dots for the border. The logic basically passes all this through to [setLineStyle](https://github.com/dompdf/dompdf/blob/db91d81866c69a42dad1d2926f61515a1e3f42c5/lib/Cpdf.php#L4120) in the CPDF library (if you're using that backend).

c-louis:
Got some time this evening to look for a solution and changed some things to make it work : 

https://github.com/c-louis/dompdf/commit/0992c92ee1a5d9cff0d942c70f38c0efe988da07

I'm certainly breaking some of the rules the project have, feel free to tell me if somethings need to be change and I will when I got the time to do it.
This solution is working for TableCell and Div and I think anything else.
I choose the dash pattern to draw the border as I didn't see any rules about that anywhere I found out that [0.4, 2 * $width] was looking good enough for me.
I tried to understand why : 

`
if ($width <= 1) {
    $pattern = [$width, $width * 2];
} 
`

But didn't found any reason :question

[...]

PR #2245, opened 2020-09-18
Display Circle instead of Square for Dotted border

c-louis:
Change the Square dotted to real dot

https://github.com/dompdf/dompdf/issues/2239#issue-702636041

### DIFF
--- a/src/Renderer/AbstractRenderer.php
+++ b/src/Renderer/AbstractRenderer.php
@@ -474,7 +474,7 @@ abstract class AbstractRenderer
                 if ($width <= 1) {
                     $pattern = [$width, $width * 2];
                 } else {
-                    $pattern = [$width];
+                    $pattern = [0.4, $width * 2];
                 }
                 break;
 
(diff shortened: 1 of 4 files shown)


src/Renderer/AbstractRenderer.php
```diff
@@ -474,7 +474,7 @@ protected function _get_dash_pattern($style, $width)
                 if ($width <= 1) {
                     $pattern = [$width, $width * 2];
                 } else {
-                    $pattern = [$width];
+                    $pattern = [0.4, $width * 2];
```
c-louis, line 477: Maybe those value should be change

bsweeney:
I think this needs just a bit more work before it's ready (e.g. update the PDFLib and GD canvas instances). Going to push to the next release so I have more time to review.

src/Renderer/AbstractRenderer.php
Mellthas: Should be 0 for a circle, see #2926.

Mellthas:
@c-louis It doesn’t look like you are coming back to this, are you okay with being attributed as co-author in #2926?

c-louis:
@Mellthas I'm glad someone had the time to make it. I am okay with being attributed as co-author.

PR #2269, opened 2020-10-16
Issue #2098: Make the file protocol respect the base path too

ytsurk:
Addresses #2098

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -1221,13 +1221,18 @@ class Style
 
     /**
      * Returns the background image URI, or "none"
-     * 
+     *
      * @link https://www.w3.org/TR/CSS21/colors.html#propdef-background-image
      * @return string
      */
     function get_background_image()
     {
-        return $this->_image($this->_props_computed["background_image"]);
+        $tmp = $this->_props_computed["background_image"];
+        // url(none) is a non-valid CSS value for background-image, but widely used.
+        if ($tmp == "url(none)") {
+          $tmp = "none";
+        }
+        return $this->_image($tmp);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Helpers.php
```diff
@@ -83,7 +83,8 @@ public static function build_url($protocol, $host, $base_path, $url)
             //drive: followed by a relative path would be a drive specific default folder.
             //not known in php app code, treat as abs path
             //($url[1] !== ':' || ($url[2]!=='\\' && $url[2]!=='/'))
-            if ($url[0] !== '/' && (strtoupper(substr(PHP_OS, 0, 3)) !== 'WIN' || (mb_strlen($url) > 1 && $url[0] !== '\\' && $url[1] !== ':'))) {
+            //Also respect the base path for the file:// protocol
+            if (mb_strtolower($protocol) == 'file://' || $url[0] !== '/' && (strtoupper(substr(PHP_OS, 0, 3)) !== 'WIN' || (mb_strlen($url) > 1 && $url[0] !== '\\' && $url[1] !== ':'))) {
```
bsweeney, line 87: I don't think this is how we want to tweak the logic to better handle file-protocol support. The current logic already support a file-protocol base protocol or URL. Current logic prepends the base path if the URL isn't absolute, which is the expected behavior. With this change when the base protocol is "file://" Dompdf would _always_ prepend the base path. We only want to prepend the base path if the url is relative.

src/Css/Style.php
```diff
@@ -1221,13 +1221,18 @@ function get_background_color()
...
      */
     function get_background_image()
     {
-        return $this->_image($this->_props_computed["background_image"]);
+        $tmp = $this->_props_computed["background_image"];
+        // url(none) is a non-valid CSS value for background-image, but widely used.
+        if ($tmp == "url(none)") {
+          $tmp = "none";
+        }
+        return $this->_image($tmp);
```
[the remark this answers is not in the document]
ytsurk: Sorry .. this should not live in this MR ..

src/Helpers.php
ytsurk: I need to re-check this ... one by one. It may be a setting confusion ..

Issue #2244, opened 2020-09-18
dompdf and drupal entity_print


code137:
Not exactly sure if there's an issue with dompdf, drupal, or the drupal entity_print module but was running into issue where pdf generation was returning network errors and failing to load. Tracing it through I could see that 'none' would be upgraded to 'url(none)' and then translated in Style.php to create urls like http://localhostnone or <domain>none which would then fail to load correctly throwing a network error. I was able to resolve it locally by adding another catch. Any thoughts on whether this is something worth catching in Style or if it's just a symptom of an issue further up in the data passed to dompdf?

```diff --git a/src/Css/Style.php b/src/Css/Style.php
index e2fc6c1..e184fab 100644
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -1804,7 +1804,7 @@ class Style
         $DEBUGCSS = $this->_stylesheet->get_dompdf()->getOptions()->getDebugCss();
         $parsed_url = "none";
 
-        if (empty($val) || $val === "none") {
+        if (empty($val) || $val === "none" || $val === "url(none)") {
             $path = "none";
         } else if (mb_strpos($val, "url") === false) {
             $path = "none"; //Don't resolve no image -> otherwise would prefix path and no longer recognize as none
```

bsweeney:
The modification of the property from "none" to "url(none)" sounds like an issue with your drupal module. But if this is an issue with Dompdf it should be easy to identify.

I'm not sure if we want to do that level of URL parsing but it's something to think about since it's an obvious error. In the meantime, since you're able to do that level of tracing, can you confirm the source of the property transformation? Easiest way to tell if it's Dompdf is to look at the HTML passed to Dompdf.

luigimannoni:
Same started happening on our Drupal site recently, my colleague rolled back dompdf to 0.8.5 and it started working again and file_get_contents gets the images properly

ytsurk:
I can confirm that this changes makes the Drupal module work again.

And it is not related to Drupal, nor dompdf, but non-valid CSS.

The call-stack comes from 

```php
  /**
   * Returns the background image URI, or "none"
   *
   * @link https://www.w3.org/TR/CSS21/colors.html#propdef-background-image
   * @return string
   */
   function get_background_image()
   {
      return $this->_image($this->_props_computed["background_image"]);
   }
```

[...]

PR #2270, opened 2020-10-16
Handle the invalid url(none) background-image value

ytsurk:
Addresses #2244


bsweeney:
For now I'm not planning to support the non-standard url(none) syntax. Dompdf will parse this as expected, and as do the web browsers.

Issue #2242, opened 2020-09-16
PHP 8 Support


shopx:

bsweeney:
Any known incompatibilities at this time? I don't have 8 set up for testing yet.

Didn't see any migration notes on the PHP site yet, but here's a list of breaking changes: https://stitcher.io/blog/new-in-php-8#breaking-changes

shopx:
eg function set_z_index in Style.php:
round($val) gives fatal error when val=auto.
change at least  from
`round($val) != $val && $val !== "auto"`
to
`$val !== "auto" && round($val) != $val`

we tested with this. no fatal error, but scrambled pdf output

bsweeney:
Thank you for that, should help identify similar issues. That probably falls under the TypeError change mentioned in the doc I referenced.

MarkBaker:
```2) PhpOffice\PhpSpreadsheetTests\Functional\StreamTest::testAllWritersCanWriteToStream with data set #6 ('Dompdf')
Required parameter $dompdf follows optional parameter $paper
/home/travis/build/PHPOffice/PhpSpreadsheet/vendor/dompdf/dompdf/src/Adapter/CPDF.php:183
/home/travis/build/PHPOffice/PhpSpreadsheet/vendor/dompdf/dompdf/src/CanvasFactory.php:57
/home/travis/build/PHPOffice/PhpSpreadsheet/vendor/dompdf/dompdf/src/Dompdf.php:300
/home/travis/build/PHPOffice/PhpSpreadsheet/src/PhpSpreadsheet/Writer/Pdf/Dompdf.php:17
/home/travis/build/PHPOffice/PhpSpreadsheet/src/PhpSpreadsheet/Writer/Pdf/Dompdf.php:61
/home/travis/build/PHPOffice/PhpSpreadsheet/tests/PhpSpreadsheetTests/Functional/StreamTest.php:37```

PHP8 is a lot stricter about optional arguments to a method before mandatory arguments

fixpunkt:
We have a few products which depend on Dompdf and are planning our own timeline for achieving PHP 8 support. Because of this, some information about the Dompdf project's plans would be very much appreciated:

* Is there an approximate timeline for achieving PHP 8 support and if so, what does it look like?
* If we wanted to pitch in, what would be the best way to help out?

bsweeney:
I don't have a timeline for completing PHP 8 support at the moment. I had hoped to be further along (or started at least) by now, but things being the way they are ...

[...]

PR #2305, opened 2020-11-27
PHP 8 Compatibility

mralston:
I found that running v0.8.6 under PHP 8 triggered several issues.

- Unsupported operand types: string + float
- Required parameter follows optional parameter

I am not familiar with how this package works, so I've attempted to quell the above with minimal code changes. It renders, but I'm sure further changes will be required.

Of note, in order to fix the problems with the required/optional parameter order, I have simply given the out of place required parameters default values making them optional. This is probably not desired, but arguably better than breaking backwards compatibility by changing the order of the arguments. It might be worth throwing an exception if these parameters have their default values. In the case of the object typed parameters, the default value I've added (null) would probably throw anyway as I didn't include the optional ? operator.

Also, there appears to be an issue where the width or the height is 'auto' (string) and it then attempts to add borders (float|int). Obviously a string and a number cannot be added together and even if they could it would not produce the desired result. My update attempts to mimic the behaviour of PHP 7 in this situation (the string value is discarded). I expect what needs to happen is that the 'auto' value is resolved to a numeric value first, before the border is added.

So far from perfect, but it's an honest attempt at getting it to run. I fully expect that this pull request won't be merged as is, but many developers prefer a pull request to an issue, it's a start.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         }
     },
     "require": {
-        "php": "^7.1",
+        "php": "^7.1|^8.0",
         "ext-dom": "*",
         "ext-mbstring": "*",
         "phenx/php-font-lib": "^0.5.2",
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -5543,7 +5543,7 @@ EOT;
      * @param bool $mask true if the image is masked
      * @throws Exception
      */
-    function addImagePng($file, $x, $y, $w = 0.0, $h = 0.0, &$img, $is_mask = false, $mask = null)
+    function addImagePng($file, $x, $y, $w = 0.0, $h = 0.0, &$img = null, $is_mask = false, $mask = null)
     {
         if (!function_exists("imagepng")) {
             throw new \Exception("The PHP GD extension is required, but is not installed.");
@@ -5889,7 +5889,7 @@ EOT;
      * @param bool $is_mask
      * @param null $mask
      */
-    function addPngFromBuf($file, $x, $y, $w = 0.0, $h = 0.0, &$data, $is_mask = false, $mask = null)
+    function addPngFromBuf($file, $x, $y, $w = 0.0, $h = 0.0, &$data = null, $is_mask = false, $mask = null)
     {
         if (isset($this->imagelist[$file])) {
             $data = null;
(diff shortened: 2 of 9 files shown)


bsweeney:
Looks like a number of changes you made address those that were addressed in the php8-compat branch, where we're finalizing changes to support PHP 8. Have you reviewed that branch or issue #2242?

I see only two things that we haven't yet addressed, one is the issues around composer and the other is the issue around the inline renderer.

The composer issue has been brought up, and a suggested fix is here: https://github.com/peterdevpl/dompdf/commit/2f2f5e1a6141782d5826ae520e10f8cbb1ddc562.

Do you have an example of the issue with the inline renderer? I didn't run into any problems with that logic when I was testing things out.

Feel free to check out that branch and submit a PR against what we've done there.

mralston:
Thanks Brian, that’s great to hear.

I was experiencing some rendering issues after my minimal changes, but as I said I have very little understanding of what DOMPDF is doing under hood.

I hadn’t seen the php8-compat branch, but I’ll give it a go tomorrow and see how it fares.

I really appreciate the feedback! 😀

mralston:
@bsweeney Just wanted to let you know that the php8-compat branch appears to work perfectly in my tests. Here are the details of my use case for reference:

**Codebase:**
laravel/framework:v8.16.1
barryvdh/laravel-dompdf:0.8.7
dompdf/dompdf:php8-compat

**Local development:**
Docker image php:8.0-rc-fpm
Debian 10.6
PHP 8.0.0RC5

**Staging server:**
CentOS 8.2
PHP 8.0.0

**Minor tweaks required to compile in Composer:**
Forked dompdf/dompdf:php8-compat to add `"php": "^7.1|^8.0",`
Forked barryvdh/laravel-dompdf:0.8.7 to add `"dompdf/dompdf": "^0.8.6|dev-php8-compat"`

**Tested:**
1 page PDF with Bootstrap 4 markup, embedded CSS & embedded images (data:xxx)
1 page PDF with Bootstrap 4 & single table, embedded CSS & embedded images (data:xxx)
39 page PDF with Bootstrap 4 & multiple tables, embedded CSS & embedded images (data:xxx)

All looks good to me. Looking forward to the the official release of 0.8.7. 👍 

Thank you very much for your assistance. This was the last piece of the puzzle upgrading this project to PHP 8.

PR #2306, opened 2020-11-28
2244

ytsurk:
Catch also strings starting with url


bsweeney:
PRs should be based on the develop branch in the future. Also, no need to stack commits and submit a new PR. If there are already PRs you can either update those or just submit your own changes independently. They'll all merge back together when the PRs are merged.

PR #2344, opened 2021-01-09
Moving the PDF version to a `const`

ThomasLandauer:
...for easier access from outside.


bsweeney:
Curious about the use case here, but no objections.

Also, reminder to submit to the develop barnch next time.

ThomasLandauer:
Two causes:
* Make it more visible - cleaner code.
* I had this in a test, and it failed now: `$I->seeResponseContains('%PDF-1.3');`. So I was thinking about ways to make this test more robust :-)

PR #2350, opened 2021-01-20
Fix code style

PWParsons:
New to all of this so please don't shoot me 🙂. This is a real simple change that should fix the failing tests so that, hopefully, the php8 compatibility upgrades can be merged to the master branch.


PWParsons:
Some friendly advise will be appreciated if this is not the reason why the branch hasn't been merged

bsweeney:
This is a whitespace change that may impact PHPCS results but not any unit tests. Code styling updates are still welceom, however.

FYI, a PHP8-compatible version of dompdf has been released.

PR #2373, opened 2021-02-13
Fix clone

emtudo:
I'm using php 8, Imagic Mage 7.0.48

And I'm in error.

Because it is something that bugs depending on the version and we just want to clone it, then we can use the clone. Even with "gmagick" it is done like this.

I think we can use the same thing here.

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -5672,19 +5672,12 @@ EOT;
         }
         // Use PECL imagick + ImageMagic to process transparent PNG images
         elseif (extension_loaded("imagick")) {
-            // Native cloning was added to pecl-imagick in svn commit 263814
-            // the first version containing it was 3.0.1RC1
-            static $imagickClonable = null;
-            if ($imagickClonable === null) {
-                $imagickClonable = version_compare(\Imagick::IMAGICK_EXTVER, '3.0.1rc1') > 0;
-            }
-
             $imagick = new \Imagick($file);
             $imagick->setFormat('png');
 
             // Get opacity channel (negative of alpha channel)
             if ($imagick->getImageAlphaChannel() !== 0) {
-                $alpha_channel = $imagickClonable ? clone $imagick : $imagick->clone();
+                $alpha_channel = clone $imagick;
                 $alpha_channel->separateImageChannel(\Imagick::CHANNEL_ALPHA);
                 // Since ImageMagick7 negate invert transparency as default
                 if (\Imagick::getVersion()['versionNumber'] < 1800) {


adam1010:
What's needed to get this merged?  The `imagick` extension for php8 isn't on PEAR yet so we're compiling it from source -- when doing so there is no version number set and this version check fails.

bsweeney:
Trying to be more cognizant of "breaking" changes, which this would be for anyone using an older version of the library. Though that seems unlikely for anyone using PHP 7.1+ as required by the current Dompdf release.

Nuranto:
Hi @bsweeney

Just to inform you that this PR solves #2442   

What it needs to get it merged ? Can I help ?

emtudo:
Are we going to merge?

bsweeney:
Yes, I will, for the 2.0 release. This will make older versions of imagick fail which is fine but, like I said, trying to be more cognizant of changes that may break functionality for some users. There is a change incoming that should address the issue with the version check for users that falls back to using the clone method if the version string is non-existent or not a valid version string).

hebinet:
Hello, do you know when v2 will be available?
The current Milestone status is only 12% done so it doesn't seem to be available in the comming days/weeks.

Is it possible to implement a workaround instead for the current v1 version?

Maybe something like an ENV variable which, when set overwrites the version check for this clone operation?
Something like `DOMPDF_FORCE_MODERN_CLONE` could be suitable?!

This could be implemented rather quickly and wouldn't be a breaking change for v1

bsweeney:
v2 will probably be a little while yet, but I have an intermediate solution which I believe will address the issue for most if not all users.

dercodercom:
Hello, how about the solution we would need this for PHP8 since our application was developed for PHP8 and this currently means a production problem for us.

bsweeney:
@dercodercom this isn't necessarily needed for PHP8 compatibility. This is to resolve a compatibility issue with the IMagick extension. As mentioned, version 2 is probably a little ways off. The 1.0.3 version will include an update to address the issue without impacting IMagick version support (371a3b3d23e124dc29564d548e3dcc456aa7a7c1).

vs0uz4:
> I'm using php 8, Imagic Mage 7.0.48
> 
> And I'm in error.
> 
> Because it is something that bugs depending on the version and we just want to clone it, then we can use the clone. Even with "gmagick" it is done like this.
> 
> I think we can use the same thing here.

Wow, congratulations xxx, fantastic contribution in my opinion. Its PR solves the problem many of us have with this PHP8 incompatibility issue.

When will this change be part of the code? Do we have any predictions?

mwlvandermaat (approved):

bsweeney:
Has anyone here tried with the most recent release? Curious if this is still an issue after the logic change.

bsweeney:
FYI I will probably, eventually, incorporate this particular change. But I understand not wanting to keep it open until ... whenever 😬

PR #2408, opened 2021-03-11
Update Cpdf.php

Varfendell:
add null check to avoid error Trying to access array offset on value of type null


bsweeney:
Can you resubmit based on the develop branch?

I'm curious if you've run into this in practice or if you're just cleaning up potential error sources.

Varfendell:
I've been through notice error on dev environment (which are not thrown in production mode)

bsweeney:
Going to move forward with #2227 instead (with that fix applied to line 3836 as well).

PR #2411, opened 2021-03-11
Remove temporary file created when caching an image.

### LINKED ISSUE #2377: bad image URLs filling temporary directory up with empty files
Cache::resolve_url() uses tempnam() to create a $resolved_url value for remote images before it tries to load the image. If it is later unable to get a valid image, it throws an exception. The temporary file created by tempnam() is never removed. Because of some weird behavior by our CMS, we ended up with hundreds of these empty files.

We could maybe add code to the catch block to remove it?

bradbulger:
Fixes #2377

### DIFF
--- a/src/Image/Cache.php
+++ b/src/Image/Cache.php
@@ -70,6 +70,7 @@ class Cache
         $data_uri = strpos($parsed_url['protocol'], "data:") === 0;
         $full_url = null;
         $enable_remote = $dompdf->getOptions()->getIsRemoteEnabled();
+        $tempfile = false;
 
         try {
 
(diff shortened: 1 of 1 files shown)


Mellthas:
This will now break in case the image is successfully downloaded. From a quick look, removing the temporary file should be done immediately before throwing the exceptions in lines 108 and 117.

bradbulger:
Could we just put it into the catch block, then?

Mellthas:
That should also work, and from a quick test, it does!

bsweeney (approved):

PR #2414, opened 2021-03-13
Fix text-width calulation with letter spacing

### LINKED ISSUE #2404: Inline element with letter spacing results in exra space before the element
This is the same HTML as in issue #2403:
```
<!DOCTYPE html>
<html>

<head>

<meta charset="UTF-8">

<title>Test</title>

<style type="text/css">
body {
	font-family: sans-serif;
	font-size: 1em;
	letter-spacing: 5pt;
}

p {
	margin: 0;
}
</style>

</head>

<body>
	<p>Lorem ipsum <b>dolor</b> sit amet, consectetur adipisici elit.</p>
</body>

</html>
```

See the attached output; there is extra space before the word _dolor._ If the text has standard letter spacing, the issue does not show up, and it seems to get worse with larger letter spacing:
[output.pdf](https://github.com/dompdf/dompdf/files/6109861/output.pdf)

Mellthas:
The letter spacing was counted twice for every space character. Not sure, if there was a reason for that, but it seems to be wrong and leads to a different width compared to what is drawn. Support for letter spacing was introduced in commit 4bbb5feae6163409726b1544acf2fc13b2d706cd and the logic hasn't changed since then.

Sample HTML that visualizes the issue:
```
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style type="text/css">
@page {
	size: 400pt 300pt;
	margin: 50pt;
}

body {
	font-family: sans-serif;
	font-size: 1em;
}

p {
	margin: 0;
}

span {
	display: inline-block;
	border: 1pt solid red;
}

.c2 {
	letter-spacing: 2pt;
}

.c3 {
	letter-spacing: 2pt;
	word-spacing: 5pt;
}

.c4 {
	letter-spacing: 5pt;
}
</style>
</head>

<body>
	<p>
		<span class="c1">Lorem ipsum dolor sit amet</span><br>
		<span class="c2">Lorem ipsum dolor sit amet</span><br>
		<span class="c3">Lorem ipsum dolor sit amet</span><br>
		<span class="c4">L o r e m i p s u m</span><br>
	</p>
</body>

</html>
```

Before: [output-before.pdf](https://github.com/dompdf/dompdf/files/6134323/output-before.pdf)
After: [output-after.pdf](https://github.com/dompdf/dompdf/files/6134324/output-after.pdf)

Fixes #2404

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -5131,11 +5131,11 @@ EOT;
      * calculate how wide a given text string will be on a page, at a given size.
      * this can be called externally, but is also used by the other class functions
      *
-     * @param $size
-     * @param $text
-     * @param int $word_spacing
-     * @param int $char_spacing
-     * @return float|int
+     * @param float $size
+     * @param string $text
+     * @param float $word_spacing
+     * @param float $char_spacing
+     * @return float
      */
     function getTextWidth($size, $text, $word_spacing = 0, $char_spacing = 0)
     {
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -5161,7 +5161,6 @@ EOT;
         $cf = $this->currentFont;
         $current_font = $this->fonts[$cf];
         $space_scale = 1000 / ($size > 0 ? $size : 1);
-        $n_spaces = 0;
 
         if ($current_font['isUnicode']) {
             // for Unicode, use the code points array to calculate width rather
@@ -5183,14 +5182,13 @@ EOT;
                     // add additional padding for space
                     if (isset($current_font['codeToName'][$char]) && $current_font['codeToName'][$char] === 'space') {  // Space
                         $w += $word_spacing * $space_scale;
-                        $n_spaces++;
                     }
                 }
             }
 
             // add additional char spacing
             if ($char_spacing != 0) {
-                $w += $char_spacing * $space_scale * (count($unicode) + $n_spaces);
+                $w += $char_spacing * $space_scale * count($unicode);
             }
 
         } else {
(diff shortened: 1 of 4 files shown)


Mellthas:
Curiously, the PDFLib adapter _subtracts_ the number of spaces before calculating the extra space used for the letter spacing. Can't try this out without access to the PDFLib library, so I didn't touch that code.

bsweeney:
I'll take a look at the PDFLib code, I have a version of the adapter installed ... somewhere.

bsweeney:
I was able to review PDFLib and it's definitely not right. I think the logic was written with a misunderstanding of when letter-spacing applied. Namely that it did not apply to the word-break characters (e.g. the space character):

> This property specifies additional spacing (commonly called tracking) between adjacent typographic character units.
https://drafts.csswg.org/css-text/#letter-spacing-property

The discussion of a typographic character requires a bit of digestion and I'm not prepared to go into it. In comparing rendering between the various adapters and browsers, though, I'm confident that it's safe to remove the space character exclusion logic from the PDFLib adapter. Do you want to make that change here, or leave it to someone with access to PDFLib?

bsweeney:
Actually, looking through the PDFLib documentation a method exsits for making character/word spacing adjustment (set_parameter in version 7, set_text_option in version 9). I'm not sure why we're not using that functionality since it would simplify our own logic.

I'm going to change the adapter method in a separate commit since I need to see if that change would impact the rendering adversely. I also need to see if I can get version 7 of pdflib installed so I can test it out.

Mellthas:
I added an extra commit removing the space exclusion in the PDFLib adapter. Can you check that it works as expected? I think further simplifying the logic can then be done in a later step.

Issue #2347, opened 2021-01-14
A non-numeric value encountered at Renderer/Text.php


justijndepover:
This has already been reported many times. I'm encountering this error in said file:

```
A non-numeric value encountered at /var/www/production/releases/91/vendor/dompdf/dompdf/src/Renderer/Text.php:158)
```

bsweeney:
Which version of Dompdf and PHP are you using? Do you have a sample document that produces the error?

justijndepover:
php 7.1 and dompdf 0.8.6

I managed to find the culprit: a table tag was rendered inside a span tag.

```html
<span>
    <table>
        ...
    </table>
</span>
```

My clients have a wysiwyg editor, so by copy pasting, the above structure happened.

bsweeney:
Invalid structure is obviously going to be a problem. Recommend your client run with the bundled HTML5 parser enabled.

We addressed a number of non-numeric value issues with the 1.0.0 release in support of PHP8. I tried the following in the latest release and was unable to reproduce the issue.

```
<span>
    <table>
        <tr>
            <td>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et
                dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip
                ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu
                fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia
                deserunt mollit anim id est laborum.</td>
                <td><img src="https://placekitten.com/500/500"></td>    
        </tr>
    </table>
</span>
```

Possible for the client to upgrade and test the offending code again? If it's still an issue I'll need a fuller example to help track down why it's happening. I have suspicions but want to be able to confirm before making changes.

SnakeDrak:
I am using the latest version of **dompdf** (1.0.2) and I have the same issue. It is hard to reproduce it, I have created a document which reproduces the issue. However, if you remove anything (the **latest href**, some **space**, any column of any tables...) the issue cannot be reproduced.

Obviously, as @bsweeney mentioned, the issue seems to be related with an invalid structure. However, maybe with this information we can know how to fix it.

The issue is on **linea 158** of **src/Renderer/Text.php**. It is tring to sum the variable **$width** which contains the value **auto**. And the **$style** contain the __prop_cache_ as:

[...]

PR #2423, opened 2021-03-29
Prevent case exception in text renderer

mudasser1985:
On line 158 of this file width is adding up that expects a number.
In case of " auto " width, it throws an error as following and breaks the PDF render.
"ErrorException: A non-numeric value encountered in /Path_To_Project/vendor/dompdf/dompdf/src/Renderer/Text.php:158"

suggestion: type casting the $width to float will convert the string into a numeric value that will prevent the ErrorException.


bsweeney:
In the future please submit any pull requests against the develop branch.

Do you have a sample document that causes the issue?

mudasser1985:
> In the future please submit any pull requests against the develop branch.
> 
> Do you have a sample document that causes the issue?

Please find the attached zip file for the sample document that is causing the issue. Also the screenshot of actual error is attached for your reference. 

looking forward to your response for this error follow up.  

Non-Numeric-Value-Error-Screenshot
[BA.zip](https://github.com/dompdf/dompdf/files/6230458/BA.zip)

mudasser1985:
Hello bsweeney,
when can i expect this issue to be fixed in master branch?

bsweeney:
I apologize, but I've not been able to find time to work on the next release up to now. I'm still unable to reproduce the issue with your sample. It could be a combination of factors (HTML + CSS as well as page size, DPI, etc). I'll do a bit more investigation but don't see any harm including this change.

fd6130:
> I apologize, but I've not been able to find time to work on the next release up to now. I'm still unable to reproduce the issue with your sample. It could be a combination of factors (HTML + CSS as well as page size, DPI, etc). I'll do a bit more investigation but don't see any harm including this change.

Have you try with the `declare(strict_types=1);` after the `<?php` ? Only that will avoid auto convert in php and the error might occurs, i guess?

Mellthas:
This is probably related to #2347.

Mellthas:
While the proposed fix does work, I think the proper fix is to ensure that the width is resolved at this point. Does the error still occur with the current `develop` state, i.e. with 51c21a763aa9dd37eb910f409f05b6072074899c in? I cannot reproduce the error, neither with the sample provided here, nor with the one from #2347, even using version 1.0.2.

bsweeney:
responded on 2347 ... seems to be caused by a specific structure of invalid html

Mellthas:
Have managed to reproduce it now. Regarding the discussion in #2347, I think it might still be worthwhile to take this, because the source problem seems harder to track down.

PR #2436, opened 2021-04-13
Use directory for fontCache instead of the fontDir

Nyholm:
The cache file should be in the cache directory, right?


bsweeney (approved):

bsweeney (changes requested):
The placement of this file in the font directory likely the correct behavior. The file is misnamed as it's not actually a cache file but a list of available fonts. I need to think this through a bit more. You can read some relevant discussion in #2450

bsweeney:
I think any changes related to this request can be done while working on the switch to using JSON for persisting font information.

Issue #2161, opened 2020-06-02
Data Execution Vulnerability


makss:
Please pay attention.
This package saves `data` as `php files`, and then executes them.

Why aren't safe methods of saving/reading data used?
At least the simplest json_encode() / json_decode(), these are just a few lines of code.

Data must be data and you should never execute it as code.

bsweeney:
This is a hold-over from way back when Dompdf was initially developed. We've taken efforts to mitigate the exploitation vulnerability, but your comment still holds. Data should be treated differently from executable logic. We'll look at making this change in a future release.

bsweeney:
issue was originally mentioned in #1543

Issue #1543, opened 2017-09-15
About the usage of 'require' in FontMetrics::loadFontFamilies()...


tucson-tom:
In FontMetrics.php, the loadFontFamilies() has these two lines of code:

```
        $file = $rootDir . "/lib/fonts/dompdf_font_family_cache.dist.php";
        $distFonts = require $file;
```

It probably works, but that's ugly.  Whatever happened to simple variable assignment?

Loading a JSON file and then patching the paths from inside the function seems like a much better solution.

This lint issue was discovered while decomposing dompdf.

bsweeney:
This logic is pretty much a hold-over from previous versions of Dompdf. We've talked about converting to JSON but I remember seeing some potential issues at the time and dropped the idea. What those issues were I can't remember.

If you have the time and inclination feel free to submit a PR with a suggested improvement.

PR #2450, opened 2021-05-05
use JSON for font metrics cache instead of PHP

Spreeuw:
This PR stores the font metrics as JSON instead of PHP, keeping backwards compatibility with existing `.php` cache files and automatically migrating them to the JSON format.

**Background**
We use dompdf in a plugin for WordPress, and use the WordPress upload folder (`wp-content/uploads/`) for the (temporary) font files, for a lack of a better alternative. This works fine, except that there are security plugins for WordPress that don't like PHP files being written to that folder (and its subfolders), which need to be writable and thus could theoretically pose a security risk. When these files are removed in automated cleanup processes, the fonts are rendered incorrectly in dompdf.

Using JSON instead of a PHP file with `return array()` not only prevents this conflict, but also mitigates a potential attack vector that is similar to `eval()` in usage.

The provided solution is backwards compatible, following the same pattern that was used for the cache file changes introduced by dompdf 0.6 (which stored the data as plain text in a file `php_$name.ufm` and processed the contents using `eval()`).
To avoid having to introduce `eval()` into the code again, I have opted to leave the previous compatibility as-is. This means that if someone is migrating from 0.5 to 1.X, the cache files would first be converted to the `.ufm.php` format, and then to the new `.ufm.json` format in a second step. This would be a one-time conversion in an upgrade scenario that is not very likely.


bsweeney:
Thanks for the pull request. This is a long-standing request, see #1543.

FYI, PRs should be based on the Develop branch.

Spreeuw:
Thanks! This PR currently only addresses the font metrics files themselves, not the index ('dompdf_font_family_cache.php`). I'll see if I can add that too.

Spreeuw:
The last commits also changed the index files (`dompdf_font_family_cache.php` and `dompdf_font_family_cache.dist.php`) to the JSON format.

Sharing some of the  design choices I have made:

Rather than storing an absolute path to the font files, in the current release, `FontMetrics` processes the data during the save process, replacing known paths from the `fontDir` and `rootDir` options by variable names (which used to be contants in dompdf 0.6):
https://github.com/dompdf/dompdf/blob/3ed5eab9a259c909679d781daab6b71d4de94e57/src/FontMetrics.php#L100-L101

Upon reading the files, these variables are then set so that the resulting array has the full paths again:
https://github.com/dompdf/dompdf/blob/3ed5eab9a259c909679d781daab6b71d4de94e57/src/FontMetrics.php#L125-L133

In this PR, I've changed this slightly, leaving out the direct reference and instead splitting the different locations into three categories: `rootDir` (for the bundled `.dist` fonts), `fontDir` (for user defined fonts inside the `fontDir` defined in options) and `absolute` (for fonts that reside in neither folders*). Now, instead of loading the families directly into the `$fontLookup` property as before (which needs full paths and would thus modify the data directly), I load them separately into their own properties `$distFonts` and `$userFonts`. This ensures that the data representation itself is not changed and can be saved back without any preprocessing (removing base paths).
After loading this data, it is processed into a separate property `$fontFamilies` that replaces the previous `$fontLookup`. This is a singular list with all the families from the 3 categories combined (user fonts preferred over dist fonts), with their full working paths.

The rest is pretty similar, although I needed to shuffle some variables in the `registerFont()` method to be able to maintain the separation between full paths (for saving the font data such as `.ttf` and `.ufm` into the right location). I didn't touch the structure itself, but think that some helper methods for getting a path for a family file with a specific extension would make this code more readable.

There is some wild code here that I didn't touch either, because honestly I couldn't grasp what this was doing:
https://github.com/dompdf/dompdf/blob/3ed5eab9a259c909679d781daab6b71d4de94e57/src/FontMetrics.php#L186-L197
It feels like that could (should?) be simpler...

Legacy `.php` cache files are converted into the new JSON format automatically and saved, but **not** deleted (this holds for both the indexes (`.php` and `.dist.php` as well as the metrics cache `.ufm.php`). This is suboptimal and I am certainly not opposed to unlinking the files (especially the index file because it would be processed on each run), but I'd like input on this. @bsweeney thoughts?

I've taken much care to make this as backwards compatible as possible, and being a non-breaking change I think this could be included in a 1.X update rather than 2.X (but I don't know your expected timeline for that release).

*One of the 'features' of the current implementation using PHP is that the user defined fonts index (`dompdf_font_family_cache.php`) can be edited directly to (theoretically) link to locations outside of the fonts folder defined in the dompdf Options. While perhaps not super likely, these will still function properly by being saved as absolute paths in the new JSON file.

Spreeuw:
by the way, I didn't change the filename itself, but I think the name `dompdf_font_family_cache` is not very accurate. It is not cache but simply a list of font families. `font-families.json` and `font-families-bundled.json` may be more approriate?

bsweeney:
>  I didn't change the filename itself, but I think the name `dompdf_font_family_cache` is not very accurate. It is not cache but simply a list of font families. `font-families.json` and `font-families-bundled.json` may be more approriate?

I agree that the name doesn't quite match the purpose. I don't have a good feel for what would be a better name. What you have would work.

bsweeney:
> There is some wild code here that I didn't touch either, because honestly I couldn't grasp what this was doing:

That's just trying to make sure that the file name is simple enough not to cause problems for the local filesystem. That's in there mostly to protect from weirdly named remote files.

makss:
@Spreeuw, Thanks for your PR, this solves issue #2161 too.

bsweeney:
I'm ready to pull this in. If you have time to rebase and address conflicts that'd be great, but if not no worries.

Spreeuw:
@bsweeney unfortunately I don't have time to resolve this at the moment

bsweeney:
That's awesome I can take it from here.

PR #2466, opened 2021-06-01
Fix undefined variable

### LINKED ISSUE #2465: Undefined variable: rootDir
I'm getting undefined variable roorDir here: https://github.com/dompdf/dompdf/blob/0ccafeb142dd04f556b43f1e8b5541aa06de1c26/lib/fonts/dompdf_font_family_cache.dist.php#L2

It seems to be some sort of side-effect when using PhpUnit's `@runInSeparateProcess` feature - I think the new process is trying to include all files that were included in the original process, but the variable this file expects to exist in the context is of course missing.

My question is, why is the variable there? Why don't we simply use `$distFontDir = __DIR__;` instead?

enumag:
Closes #2465

I prepared a fix that works for me. Could you review it @bsweeney? Basically I have the cache file (both dist and generated) return a closure instead of an array. That way I can pass the variable into the closure and the file no longer needs to be loaded in a context where the variable exists.

### DIFF
--- a/src/FontMetrics.php
+++ b/src/FontMetrics.php
@@ -92,7 +92,8 @@ class FontMetrics
     public function saveFontFamilies()
     {
         // replace the path to the DOMPDF font directories with the corresponding constants (allows for more portability)
-        $cacheData = sprintf("<?php return array (%s", PHP_EOL);
+        $cacheData = sprintf("<?php return function (%s) {%s", '$fontDir', PHP_EOL);
+        $cacheData .= sprintf("return array (%s", PHP_EOL);
         foreach ($this->fontLookup as $family => $variants) {
             $cacheData .= sprintf("  '%s' => array(%s", addslashes($family), PHP_EOL);
             foreach ($variants as $variant => $path) {
@@ -103,7 +104,8 @@ class FontMetrics
             }
             $cacheData .= sprintf("  ),%s", PHP_EOL);
         }
-        $cacheData .= ") ?>";
+        $cacheData .= ");" . PHP_EOL;
+        $cacheData .= "}; ?>";
         file_put_contents($this->getCacheFile(), $cacheData);
     }
 
@@ -130,14 +132,16 @@ class FontMetrics
         if (!defined("DOMPDF_FONT_DIR")) { define("DOMPDF_FONT_DIR", $fontDir); }
 
         $file = $rootDir . "/lib/fonts/dompdf_font_family_cache.dist.php";
-        $distFonts = require $file;
+        $distFontsClosure = require $file;
+        $distFonts = $distFontsClosure($rootDir);
 
         if (!is_readable($this->getCacheFile())) {
             $this->fontLookup = $distFonts;
             return;
         }
 
-        $cacheData = require $this->getCacheFile();
+        $cacheDataClosure = require $this->getCacheFile();
+        $cacheData = $cacheDataClosure($fontDir);
 
         $this->fontLookup = [];
         if (is_array($this->fontLookup)) {
@@ -195,7 +199,7 @@ class FontMetrics
         mb_substitute_character($substchar);
         $prefix = preg_replace("[\W]", "_", $prefix);
         $prefix = preg_replace("/[^-_\w]+/", "", $prefix);
-        
+
         $localFile = $fontDir . "/" . $prefix . "_" . $remoteHash;
 
         if (isset($entry[$styleString]) && $localFile == $entry[$styleString]) {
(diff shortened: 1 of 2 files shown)


src/FontMetrics.php
```diff
@@ -130,14 +132,16 @@ public function loadFontFamilies()
...
+        $distFonts = $distFontsClosure($rootDir);
 
         if (!is_readable($this->getCacheFile())) {
             $this->fontLookup = $distFonts;
             return;
         }
 
-        $cacheData = require $this->getCacheFile();
+        $cacheDataClosure = require $this->getCacheFile();
+        $cacheData = $cacheDataClosure($fontDir);
```
enumag, line 144: Do I need to worry about the possibility of the cache file being from older dompdf? It could be fixed like this:

```suggestion
        $cacheDataClosure = require $this->getCacheFile();
        $cacheData = is_array(cacheDataClosure) ? $cacheDataClosure :  $cacheDataClosure($fontDir);
```

enumag:
ping @bsweeney

bsweeney:
The generated dompdf_font_family_cache.php file utilizes both $fontDir and $rootDir variables. That doesn't cause problems? Even if not, since the cause is PhpUnit and the generated file isn't present, shouldn't the generated closure accept both variables?

enumag:
In my case the dist file was used so I didn't notice the issue. Thanks for catching it, I updated the code accordingly.

src/FontMetrics.php
bsweeney: Just want to check this comment and see what happens. Did you test the potential of running before and after the code is modified?

enumag: no, my app isn't using this part of the code

bsweeney: We would definitely need to add this logic to handle the alternate format.

bsweeney:
Since this is more than just a bug fix I'm going to move it to the 1.1 release. Compatible with older versions but it is new functionality.

src/FontMetrics.php
```diff
@@ -130,14 +132,16 @@ public function loadFontFamilies()
...
+        $distFonts = $distFontsClosure($rootDir);
 
         if (!is_readable($this->getCacheFile())) {
             $this->fontLookup = $distFonts;
             return;
         }
 
-        $cacheData = require $this->getCacheFile();
+        $cacheDataClosure = require $this->getCacheFile();
+        $cacheData = $cacheDataClosure($fontDir, $rootDir);
```
enumag, line 144:
```suggestion
        $cacheData = is_array(cacheDataClosure) ? $cacheDataClosure : $cacheDataClosure($fontDir, $rootDir);
```

src/FontMetrics.php
enumag: Alright, I added the condition.

src/FontMetrics.php
```diff
@@ -130,14 +132,16 @@ public function loadFontFamilies()
...
+        $distFonts = $distFontsClosure($rootDir);
 
         if (!is_readable($this->getCacheFile())) {
             $this->fontLookup = $distFonts;
             return;
         }
 
-        $cacheData = require $this->getCacheFile();
+        $cacheDataClosure = require $this->getCacheFile();
+        $cacheData = is_array(cacheDataClosure) ? $cacheDataClosure : $cacheDataClosure($fontDir, $rootDir);
```
enumag, line 144:
```suggestion
        $cacheData = is_array($cacheDataClosure) ? $cacheDataClosure : $cacheDataClosure($fontDir, $rootDir);
```

enumag:
@enumag Can't say I agree with that. It's just a change of the internal logic regarding fonts to fix a rare bug. 1.1.0 has a lot of other issues in it's scope so this would force me to use a fork for a long time.

src/FontMetrics.php
```diff
@@ -141,7 +141,7 @@ public function loadFontFamilies()
         }
 
         $cacheDataClosure = require $this->getCacheFile();
-        $cacheData = $cacheDataClosure($fontDir, $rootDir);
+        $cacheData = is_array(cacheDataClosure) ? $cacheDataClosure : $cacheDataClosure($fontDir, $rootDir);
```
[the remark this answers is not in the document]
enumag: Hmm? I fixed that already... https://github.com/dompdf/dompdf/pull/2466/commits/cfd85ec74af7c42540a907ca876c8ffd633b14cd

bsweeney:
I'll think it over. The issues addressed in a milestone are fluid. I can push a lot of those to 1.1.1 but I want to release 1.0.3 to address a number of bugs that aren't quite so rare.

src/FontMetrics.php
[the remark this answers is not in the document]
bsweeney: thanks, github 😞

enumag:
@bsweeney I'm sorry but it seems I accidentally targeted wrong branch with this PR. Could you revert it in the develop branch and cherry-pick it to master?

enumag:
Ah no, other PRs are also targeting develop... meaning [this](https://github.com/dompdf/dompdf/blob/4c65810c797674d0ceb8918859c645f298b5e358/composer.json#L54) is outdated...

PR #2497, opened 2021-07-27
Improve page break related to inline(-block) elements

Mellthas:
* Inline frames now properly report their height, the same as text frames. This resolves inline frames extending into the page margins instead of breaking onto a new page.
* The positioning of inline frames has also been improved, allowing line breaks within them in more instances (instead of breaking the whole frame).
* Allow line breaks and page breaks before inline-block elements, which were not allowed before.

Nested inline frames with margin, border, or padding still cause bugs at line breaks, because they are added to the text children on-the-fly. To properly fix these cases, it would probably be best to refactor the code in such a way that margin, border, and padding are properly handled on the inline frame itself, instead of being added to first resp. last child. It might be even worth considering always containing text frames in an anonymous inline box if needed; that could allow handling these properties consistently while simplifying the code. But the changes should improve the situation for now, nonetheless.

I have a few test cases in preparation that will demonstrate several broken scenarios that are fixed by this PR.

### DIFF
--- a/src/Frame.php
+++ b/src/Frame.php
@@ -465,42 +465,12 @@ class Frame
     //........................................................................
 
     /**
-     * Return the height of the margin box of the frame, in pt.  Meaningless
-     * unless the height has been calculated properly.
-     *
-     * @return float
-     */
-    public function get_margin_height()
-    {
-        $style = $this->_style;
-
-        return (
-            (float)$style->length_in_pt(
-                [
-                    $style->height,
-                    (float)$style->length_in_pt(
-                        [
-                            $style->border_top_width,
-                            $style->border_bottom_width,
-                            $style->margin_top,
-                            $style->margin_bottom,
-                            $style->padding_top,
-                            $style->padding_bottom
-                        ], $this->_containing_block["w"]
-                    )
-                ],
-                $this->_containing_block["h"]
-            )
-        );
-    }
-
-    /**
      * Return the width of the margin box of the frame, in pt.  Meaningless
      * unless the width has been calculated properly.
      *
      * @return float
      */
-    public function get_margin_width()
+    public function get_margin_width(): float
     {
         $style = $this->_style;
 
@@ -516,29 +486,30 @@ class Frame
     }
 
     /**
+     * Return the height of the margin box of the frame, in pt.  Meaningless
+     * unless the height has been calculated properly.
+     *
      * @return float
      */
-    public function get_break_margins()
+    public function get_margin_height(): float
     {
         $style = $this->_style;
 
-        return (
-            (float)$style->length_in_pt(
-                [
-                    //$style->height,
-                    (float)$style->length_in_pt(
-                        [
-                            $style->border_top_width,
-                            $style->border_bottom_width,
-                            $style->margin_top,
-                            $style->margin_bottom,
-                            $style->padding_top,
-                            $style->padding_bottom
-                        ], $this->_containing_block["w"]
-                    )
-                ],
-                $this->_containing_block["h"]
-            )
+        return (float)$style->length_in_pt(
+            [
+                $style->height,
+                (float)$style->length_in_pt(
+                    [
+                        $style->border_top_width,
+                        $style->border_bottom_width,
+                        $style->margin_top,
+                        $style->margin_bottom,
+                        $style->padding_top,
+                        $style->padding_bottom
+                    ], $this->_containing_block["w"]
+                )
+            ],
+            $this->_containing_block["h"]
         );
     }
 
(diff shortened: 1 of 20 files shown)


Mellthas:
I have added one more commit that should improve auto-width calculation for inline-block and floating elements by sticking more closely to the spec.

Mellthas:
Added another commit to trim trailing whitespace after line breaks in more cases, notably after inline and inline-block frames.

Mellthas:
For a sample of the scenarios the commits here intend to fix, see the test cases included and listed in #2510.

bsweeney:
Lots to review here, in the meantime when running through my standard test bed I noticed two issues. I haven't had a chance to look and see what's going on yet, so here's the errors.

First, input buttons render with no width:
```
<input type="submit" value="submit" />
```

Second, rendering http://eclecticgeek.com/dompdf/core_tests/encoding_utf-8_all.html throws the following exception:
```
[11-Sep-2021 19:33:25 America/New_York] 
[11-Sep-2021 19:33:25 America/New_York] TypeError: Return value of Dompdf\FrameReflower\Text::_collapse_white_space() must be of the type string, null returned in C:\Users\datho\Documents\code\dompdf\dompdf\src\FrameReflower\Text.php:86
Stack trace:
[0] C:\Users\datho\Documents\code\dompdf\dompdf\src\FrameReflower\Text.php(261): Dompdf\FrameReflower\Text->_collapse_white_space('D800  \xED\xA0\x80\xED\xA0\x81\xED\xA0\x82...')
[1] C:\Users\datho\Documents\code\dompdf\dompdf\src\FrameReflower\Text.php(396): Dompdf\FrameReflower\Text->_layout_line()
[2] C:\Users\datho\Documents\code\dompdf\dompdf\src\FrameDecorator\AbstractFrameDecorator.php(902): Dompdf\FrameReflower\Text->reflow(Object(Dompdf\FrameDecorator\Block))
[3] C:\Users\datho\Documents\code\dompdf\dompdf\src\FrameReflower\Block.php(872): Dompdf\FrameDecorator\AbstractFrameDecorator->reflow(Object(Dompdf\FrameDecorator\Block))
[4] C:\Users\datho\Documents\code\dompdf\dompdf\src\FrameDecorator\AbstractFrameDecorator.php(902): Dompdf\FrameReflower\Block->reflow(NULL)
[5] C:\Users\datho\Documents\code\dompdf\dompdf\src\FrameReflower\Page.php(141): Dompdf\FrameDecorator\AbstractFrameDecorator->reflow()
[6] C:\Users\datho\Documents\code\dompdf\dompdf\src\FrameDecorator\AbstractFrameDecorator.php(902): Dompdf\FrameReflower\Page->reflow(NULL)
[7] C:\Users\datho\Documents\code\dompdf\dompdf\src\Dompdf.php(837): Dompdf\FrameDecorator\AbstractFrameDecorator->reflow()
[8] C:\Users\datho\Documents\code\dompdf\test\run-tests.php(145): Dompdf\Dompdf->render()
[9] {main}
```

Mellthas:
The second error is due `preg_replace` returning `null` in case of an error. In case of https://eclecticgeek.com/dompdf/core_tests/encoding_utf-8_all.html, it seems to be related to the surrogate code points starting at D800. Notice how there is no output for these when running through the debug helper at https://eclecticgeek.com. On my system (Ubuntu 20.04 with the default PHP 7.4.3), `preg_replace` runs without errors in these cases, the resulting output has the characters included as question marks (which I think is as expected). See the attached [output file](https://github.com/dompdf/dompdf/files/7149696/encoding_utf-8_all.pdf).

I have pushed an update to resolve the error by explicitly retaining the current behavior: Effectively returning the empty string in case of a `null` return value of `preg_replace`. Whether that behavior is fine, or if it would make more sense to preserve the original string in case of an error is I think a discussion unrelated to the current pull request.

Mellthas:
By the way, there is a broken `br` tag down there in _encoding_utf-8_all.html_ ;-).

Mellthas:
Input buttons rendering without width is probably related to the change to use shrink-to-fit width resolution for float and inline-block elements. Generated content is only set right before reflowing the generated-content frame, so it is not yet set when calling `get_min_max(_content)_width`. This already affects the width calculations for table cells.

Mellthas:
I don't see a straight-forward way for handling the generated content right now, so I am going to split up the PR.

Mellthas:
The split is done: I would review #2545, #2546, #2547 first, then #2548. That finally leaves #2549, which might require more work to prevent the mentioned regressions related to generated content.

bsweeney:
> 
> 
> By the way, there is a broken `br` tag down there in _encoding_utf-8_all.html_ ;-).

heh, yes thanks

bsweeney:
> 
> 
> The split is done: I would review #2545, #2546, #2547 first, then #2548. That finally leaves #2549, which might require more work to prevent the mentioned regressions related to generated content.

Are those replacements for this PR or in addition to it? Should I go ahead and finish reviewing this one before I move on to those?

Mellthas:
Those PRs replace this one completely, so no need to continue reviewing here.

PR #2505, opened 2021-08-01
Add support for any inline-positioned frames in text alignment

### LINKED ISSUE #1761: `text-align:center` not working for child elements with `display:inline-block`
This is probably the same bug as #1635 but without any overhead.

Sample code:
```html
<div style="text-align:center"><div style="display:inline-block">foo</div></div>
```
Expected result (as shown in Firefox & Chrome):
"foo" is centered on the page.

Actual result in dompdf 0.8.2:
"foo" is aligned to the left.

Mellthas:
This notably adds support for inline-block.

The second commit needs #2414 to work properly. What the old code did, was basically undo the error in text-width calculation with letter spacing for the CPDF backend when text was justify aligned.

Fixes #1761

### DIFF
--- a/src/FrameReflower/Block.php
+++ b/src/FrameReflower/Block.php
@@ -12,6 +12,7 @@ use Dompdf\Frame;
 use Dompdf\FrameDecorator\Block as BlockFrameDecorator;
 use Dompdf\FrameDecorator\TableCell as TableCellFrameDecorator;
 use Dompdf\FrameDecorator\Text as TextFrameDecorator;
+use Dompdf\Positioner\Inline as InlinePositioner;
 use Dompdf\Exception;
 use Dompdf\Css\Style;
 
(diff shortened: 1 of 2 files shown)


src/FrameReflower/Block.php
```diff
@@ -521,11 +520,10 @@ protected function _text_align()
 
                 // Adjust the last line if necessary
                 if ($last_line->left) {
-                    foreach ($last_line->get_frames() as $frame) {
-                        if ($frame instanceof BlockFrameDecorator) {
-                            continue;
+                    foreach ($line->get_frames() as $frame) {
```
bsweeney, line 523: Should this be `$last_line->get_frames()`?

bsweeney (changes requested):

src/FrameReflower/Block.php
Mellthas: Good catch, thanks!

Mellthas: Fixed

bsweeney (approved):

PR #2507, opened 2021-08-01
Fix orphans check

### LINKED ISSUE #1617: Orphans calculation seems to be off by one
See the attached test case (document size 400pt×300pt):
[orphans-bug.zip](https://github.com/dompdf/dompdf/files/1548572/orphans-bug.zip)

The CSS specifies `orphans: 3`, but two lines stay on the first page. Using `orphans: 4`, the second paragraph is moved to the second page completely. After trying out different paragraph lengths and orphan numbers, it seems the orphan count is always off by one.

Mellthas:
This also effectively forbids line breaks before the first line.

Based on work in #1356
Fixes #1617

### DIFF
--- a/src/FrameDecorator/Page.php
+++ b/src/FrameDecorator/Page.php
@@ -364,7 +364,12 @@ class Page extends AbstractFrameDecorator
 
                 // Rule C
                 $block_parent = $frame->find_block_parent();
-                if (count($block_parent->get_line_boxes()) < $frame->get_style()->orphans) {
+                $parent_style = $block_parent->get_style();
+                // The line number of the frame can be less than the current
+                // number of line boxes, in case we are backtracking. As long as
+                // we are not checking for widows yet, just checking against the
+                // number of line boxes is sufficient in most cases, though.
+                if (count($block_parent->get_line_boxes()) <= $parent_style->orphans) {
                     Helpers::dompdf_debug("page-break", "orphans");
 
                     return false;


Mellthas:
See the example HTML for the effect of preventing line breaks before the first line box:

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style type="text/css">
@page {
	size: 400pt 300pt;
	margin: 50pt;
}

body {
	font-family: sans-serif;
	letter-spacing: 0.8pt;
	border: 1pt solid black;
	line-height: 19pt;
}

p {
	margin: 0 0 30pt 0;
	border: 5pt solid black;
	orphans: 1;
	widows: 1;
}
</style>
</head>

<body>
	<p>Lorem ipsum dolor sit amet, consectetur adipisici elit, sed eiusmod tempor
	incidunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis
	nostrud exercitation ullamco laboris nisi ut aliquid ex ea commodi consequat.</p>

</body>

</html>
```

[orphans-before.pdf](https://github.com/dompdf/dompdf/files/6919710/orphans-before.pdf)
[orphans-after.pdf](https://github.com/dompdf/dompdf/files/6919711/orphans-after.pdf)

bsweeney:
I wonder if we should go ahead and implement the full update to the orphan logic from 1356 ([lines 360-380](https://github.com/dompdf/dompdf/pull/1356/files#diff-f02f8bb15cc0eb9e1233ca8a51d8a5a4f33e63ab75b12ab7718627c1c7fccf68R360-R380)). Just returning a count of the number of lines is inaccurate since it always returns number of rendered lines up to this point instead of the line we're currently checking. This is important considering that the page break logic [walks back through previous frames](https://github.com/dompdf/dompdf/blob/v1.0.2/src/FrameDecorator/Page.php#L559-L571) when a page break is disallowed.

Mellthas:
That is true. But I think the thing is (and probably the reason why the original code didn't try to find the exact line number): As long as we are only checking for orphans, checking against the current number of line boxes is actually sufficient. Backtracking within an inline context can only happen due to:

1. The orphans check failing for the current frame. Then it will also fail for every previous sibling, as `count($block_parent->get_line_boxes())` hasn't changed.
2. The `page-break-inside: avoid` check failing. Then it will also fail for any previous sibling, as they share the same block parent.

That said, I just realized that the orphans property is checked on the current frame, so it could change while backtracking. But the spec states that `orphans` and `widows` only apply to block container elements, so that potential point is moot.

If I have not missed anything, then I would hold off implementing the full logic for getting the exact line number until it is actually needed (as it would only incur a slight performance hit right now, without achieving anything). I have updated the PR with a short comment and changed the code to check the `orphans` property on the block parent. What do you think?

Mellthas:
That said, there seem to be cases where a new line box has already been added without backtracking involved. I’ll have to investigate that further.

bsweeney:
> **Edit:** Nevermind, that had backtracking involved, a block element in an inline context (`ul` within an `li`).

Is that a correct example? An LI doesn't represent inline context, unless I'm misunderstanding the example. Via https://www.w3.org/TR/CSS22/generate.html#lists:

> An element with 'display: list-item' generates a principal block box for the element's content

I haven't had a chance to think of an example where this could produce incorrect results. I suspect it would be a rare case, but basically the line after the acceptable orphan count would have to result in a non-allowed page break. When Dompdf rewinds to see if there is an acceptable breaking point earlier it would do so within the lines that should not be orphaned since we're only checking the line count and not the current line.

Truthfully it's probably a rare case, and anyway it's how things are handled at present. It will become more of an issue when widows are supported.

Mellthas:
Sorry, was a bit unclear there. What I mean is something like the following, when a page break is to occur before the inner `ul`:

```html
<ul>
  <li>Inline content
    <ul>
    …
    </ul>
  </li>
</ul>
```

The current code will forbid a page break before the inner `ul` because it thinks that the `ul` has no preceding block-level sibling. The inner `ul` has resulted in a new line box, and so when backtracking for an allowed page break, the line number for the inline content will be ‘reported’ as 2 instead of 1, and so the orphans check will not disallow a page break before it (if orphans is 2 or less without this PR, if orpahns is 1 with this PR). See the following full sample:

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
    size: 400pt 300pt;
    margin: 50pt;
}

ul {
    counter-reset: li;
}

li {
    counter-increment: li;
}

li::before {
    content: counter(li);
}
</style>
</head>

<body>
<ul>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li><ul><li></li></ul></li>
<li></li>
<li></li>
<li></li>
</ul>
</body>

</html>
```
</details>

Output: [list-bullet.pdf](https://github.com/dompdf/dompdf/files/7162266/list-bullet.pdf)

Mellthas:
#2552 fixes the sample. It will still fail if the inner `ul` has `page-break-before: avoid`. But that would be the only remaining case that I can think of were just counting the number of line boxes is not enough.

bsweeney (approved):

PR #2509, opened 2021-08-03
Add support for webp images

### LINKED ISSUE #1276: Add support for WebP
Could you add webp image format support to pdf document? webp is one of the best image that compess image file size, google already used it and google chrome already supported it

Spreeuw:
WebP is supported by PHP7.1+ which is also the current minimum requirement for dompdf.

This PR utilizes the same methods used for converting GIF & BMP to PNG, adding support for webp.

closes #1276

### DIFF
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -606,7 +606,7 @@ class CPDF implements Canvas
     }
 
     /**
-     * Convert a GIF or BMP image to a PNG image
+     * Convert image to a PNG image
      *
      * @param string $image_url
      * @param int $type
@@ -614,7 +614,7 @@ class CPDF implements Canvas
      * @throws Exception
      * @return string The url of the newly converted image
      */
-    protected function _convert_gif_bmp_to_png($image_url, $type)
+    protected function _convert_to_png($image_url, $type)
     {
         $func_name = "imagecreatefrom$type";
 
@@ -859,12 +859,14 @@ class CPDF implements Canvas
                 $this->_pdf->addJpegFromFile($img, $x, $this->y($y) - $h, $w, $h);
                 break;
 
+            case "webp":
+            /** @noinspection PhpMissingBreakStatementInspection */
             case "gif":
             /** @noinspection PhpMissingBreakStatementInspection */
             case "bmp":
-                if ($debug_png) print '!!!bmp or gif!!!';
-                // @todo use cache for BMP and GIF
-                $img = $this->_convert_gif_bmp_to_png($img, $type);
+                if ($debug_png) print "!!!{$type}!!!";
+                // @todo use cache for converted images (BMP/GIF/WebP)
+                $img = $this->_convert_to_png($img, $type);
 
             case "png":
                 if ($debug_png) print '!!!png!!!';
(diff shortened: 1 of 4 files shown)


Spreeuw:
Apologies: I accidentally based this on `master` instead of `develop`

bsweeney (changes requested):
Almost there. Works well when using the adapters for CPDF and GD, but not for PDFLib. If you don't have time to figure that out I'll pull into another branch and add the PDFLib support.

Spreeuw:
We're trying to setup PDFLib but so far haven't succeeded locally. My host provides it out of the box but it seems dompdf requires a newer version (running 7.0.5p3 I get a lot of errors even without the `webp` code from this PR/branch... Am I correct in assuming that dompdf requires version 9 of PDFLib (even though there's still code for version 7 in the Adapter)?

And just to be sure: does PDFLib support webp at all or would this require a workaround converting to another format first (using ImageMagick?).

Spreeuw:
I also wonder if it would hurt if webp support is phased in for the different adapters at different times. In my opinion it doesn't hurt anyone if webp support is introduced for CPDF & GD while PDFLib support is still being worked on (I'm guessing 99% uses the [default CPDF](https://github.com/dompdf/dompdf/blob/0ccafeb142dd04f556b43f1e8b5541aa06de1c26/src/Options.php#L260) anyway?) . It doesn't break any existing functionality for PDFLib while people using CPDF & GD can benefit from being able to use webp... I only see advantages to introducing this now rather than to wait for PDFLib support!😉

Mellthas:
Apart from PDFLib support this looks fine, although setting a WebP image as background image via CSS does not seem to apply transparency.

Can you do a clean rebase based on `develop`, so the unrelated `README.md` changes are not included?

Spreeuw:
@Mellthas done!

Spreeuw:
@bsweeney can you tell me what the current status of PDFLib integration is? It seems to crash on basic stuff like using fonts (errors like `Exception: Handle parameter or option of type 'font' has bad value 0`, at least on PHP8). This happened to me with version 7 but after installing version 9 (which is distributed under a commercial license and has watermarks for the unlicensed version), I got the same errors. Version 7 is the latest version available via PECL but that version throws even more errors because the code seems to have been tested with version 9 only...

bsweeney:
Regarding the state of support for this feature, I think it's a fair argument that we should move forward with it even if it's not fully supported. We'll mark webp as partially supported for now.

As for PDFLib, not sure what's going on there. I have two versions of PHP with PDFLib:
- OS X 10.15.7 + PHP 7.3.13 + PDFLib 9.2.0: ✔️ 
- Windows 10 + PHP 7.4.11 + PDFLib 9.3.1: ❌

I thought I also had a setup with PDFLib 7 but I can't seem to find it so I'll have to do more research. As you mentioned, though, it _is_ over 10 years old. If it's broke it's not worth fixing.

All that to say, don't spend any time on PDFLib support if you're not a regular user and can't get it working. We can move forward with what we have and I'll review the PDFLib implementation.

bsweeney (approved):

Mellthas (approved):

PR #2510, opened 2021-08-03
Implement a basic framework for output tests

Mellthas:
This implements a very basic framework for automatically testing the final output of different documents, to help with finding and fixing regressions, and documenting expected rendering. It uses the GD backend for automatic tests, because comparing the images was simple to implement. This shows several inconsistencies between the CPDF and GD backend, but I think it should still be helpful. PDFs are included as reference for manual checking. I am not too happy about the name `OutputTest`, maybe someone has a better idea, or maybe it is okay. In relation to the debug helper at https://eclecticgeek.com/dompdf/debug.php, the test cases here are intended to be more constrained and specific, covering edge cases and documenting fixed bugs.

Reference images and PDFs can be updated via the helper script `tests/update-reference-output.php`. When running the tests, the output of all failed test cases is saved to the `tmp` directory, so the differences to the reference rendering can be easily checked manually.

If have initially included several test cases that are all broken with the current code, but will be fixed by different PRs (and they have helped greatly in judging the code changes). The font is specified in a way that mimics the rendering of the GD backend <del>as close as possible, as that does only have one fixed font without support for different styles, weights, or letter or word spacing</del>.

Fixed by #2414:
* `inline/letter-spacing`

Fixed by #2499:
* `inline/extended-parent-width`

Fixed by #2506:
* `page-break/blocks-whitespace`
* `page-break/non-empty-first-page`

Fixed by #2507:
* `page-break/orphans`

Fixed by PR #2497:
* `page-break/inline-element-1`
* `page-break/inline-element-2`
* `page-break/inline-element-3`
* `page-break/inline-block-1`
* `page-break/inline-block-2`
* `page-break/inline-block-3`
* `page-break/paragraph`
* `inline/inline-block-width`
* `float/float-width`
* `table/block-in-table`
* `text-align/inline-elements` (not completely, but the rendering is improved)

Fixed by #2505:
* `text-align/inline-block` (except for the vertical alignment)

Fixed by #2500  (together with #2497):
* `position-relative/block`
* `position-relative/inline`
* `position-relative/inline-block`
* `position-relative/inline-nested`
* `position-relative/nested`

Fixed by #2542 and #2543:
* `text-align/center`
* `text-align/justify`
* `text-align/right`

<del>The reference rendering reflects the current state and needs to be updated once the different changes are merged.</del>


bsweeney:
This is very, very nice. Comparing renderings against a reference in testing is one of the things I've wanted to add to the unit tests.

bsweeney:
My renderings are failing on two different systems, one Windows and one Mac. The rendering on those systems appear to be similar, though I only did a quick visual inspection. The primary culprit seems to be line-height.

Perhaps we should specify a value for most of the settings to ensure a more consistent configuration? Also, what do you think about using the Ahem font developed for use with the CSS conformance suite: https://www.w3.org/Style/CSS/Test/Fonts/

Mellthas:
Okay, I’ll have to take a closer look, have only tested this on a Linux system so far. Can you attach (one of) the failing renders?

The Ahem font looks interesting, though it is probably not very useful as long as the GD back end does not have support for different fonts. Maybe an option is to add support for just that font for use in the tests, but haven't looked into that.

bsweeney:
Font support in GD seems OK to me, though I haven't tested on various systems.

I used the following render on my two system:

```
<link href='http://fonts.googleapis.com/css?family=Give+You+Glory&amp;v2' rel='stylesheet' type='text/css'>
<style> @page { margin: 0px; size: 350px 400px; } </style>
<p style="font-family: 'Give You Glory'; font-size: 30pt;">
  Grumpy wizards make toxic brew for the evil Queen and Jack
</p>
```

and it looks OK:

glory

bsweeney:
Here's what I got for float-width.html:

float-width p1 fail

Mellthas:
:-D Oh my, I think I simply assumed it had no support for different fonts, because `font-family: serif` didn't work …

bsweeney:
I had no idea serif _wasn't_ working!

Mellthas:
It seems specifying any of the generic font families results in DejaVu Sans being used, but without support for font weights and styles.

Mellthas:
I updated the `float-width` test to specify the font explicitly. Does that make your rendering match the updated reference?

Mellthas:
The line-height issue seems to be exhibited with DejaVu Sans. In the image you posted, @bsweeney, the line height seems to be calculated correctly: With a font size of 12pt and a line height of 1.2 (the default values), the actual line height calculates to 14.4 pt, which would fit 13, almost 14 lines into the box. With the CPDF back end the actual line height is much larger, even larger than using the GD back end on my system, which is still too large. Here is the output of `text-align/center` on my system which specifies a line-height of 14pt, which should fit 14 lines onto one page:

**GD**
center p1
center p2

**CPDF**
[center.pdf](https://github.com/dompdf/dompdf/files/7148607/center.pdf)

But as you can see, the GD output only fits 12, the CPDF one only 11 lines. Switching to the default sans-serif font, 14 lines fit using the CPDF back end, as expected with a line height of 14pt.

Mellthas:
Apart from the line-height issue there also seem to be differences wrt. kerning (and probably anti-aliasing), which could also change between different PHP versions and the (system) libraries behind. I think checking for a pixel-perfect replication is a bit too simplistic for the automatic tests. Even allowing for a small margin of error might not be enough, as the layout might change due to the text width being slightly off.

Mellthas:
GD rendering is significantly improved with the recent rounding fixes from master. PNG and PDF output are much closer now.

bsweeney:
now's a good time to revisit this one

Mellthas:
For the record, what needs to be investigated:
* Using a similarity measure for the image comparisons
* Using the Ahem font

We’d then have to see whether comparing images via the GD back end is really feasible at all, or under what circumstances. If not, maybe converting the reference PDF and current PDF output to images on-the-fly and comparing these might be an angle to pursue. I still see quite a lot of questions to be answered and details to be worked out here.

It would also be good to update the current set of tests, there is some overlap, and for some it might be enough to check the frame tree for box dimensions instead of doing full output comparison.

I don’t have the time to pursue this further currently. If you or anyone wants to pick this up, feel free to do so.

8ctopus:
Is it correct that until this PR gets merged, there's currently no way to test the generated pdf? If that's the case, I'm really excited about the PR.

bsweeney:
@8ctopus do you mean test user-generated output during runtime? That's not the purpose of this PR. This PR adds unit tests that would allow us to validate that changes to code do not break expected rendering (based on the unit tests that are available).

8ctopus:
@bsweeney Thank you for the explanation, I got it.

johanadivare:
@Mellthas if we would run these test in docker wouldn't it be consistent no matter if you run the docker test in windows or linux?

Mellthas:
Hadn’t thought of it, but that might work, yes. If you want to try that out, feel free to do do so.

JohJohan:
@Mellthas i setup basic docker file: https://github.com/Mellthas/dompdf/pull/1 note i still had some problems with test failing although the files where generated in the container. Maybe you can have a look if you have time.

Start the box:  `docker build -t dompdf . `
And then enter the box and you can execute `vendor/bin/phpunit`

johanadivare:
@Mellthas i now know that i rendered the references incorrect and now they are good. and `vendor/bin/phpunit` works for me. Then we only need to change `/.github/workflows/test.yml` to use a docker setup right?

tests/OutputTest.php
```diff
@@ -0,0 +1,122 @@
...
+     * @dataProvider outputTestProvider
+     */
+    public function testOutput(Dataset $dataset): void
+    {
+        $document = $dataset->render("gd");
+        $pageCount = $document->getCanvas()->get_page_count();
+        $referenceImages = $dataset->referenceImages();
+
+        if ($pageCount !== count($referenceImages)) {
+            $this->saveFailedOutput($dataset, $document);
```
johanadivare, line 54: No longer need to pass `$document`

johanadivare:
@Mellthas Do you have time to look at my pull request https://github.com/Mellthas/dompdf/pull/1?

Mellthas:
> @Mellthas Do you have time to look at my pull request [Mellthas#1](https://github.com/Mellthas/dompdf/pull/1)?

Hey, sorry, haven’t had the time yet. Hope I can find some time next week.

johanadivare:
@Mellthas just a poke so you dont forget

Mellthas:
@johanadivare I will look at this next, just want to finish up the on some small updates to selector parsing.

tests/OutputTest.php
```diff
@@ -0,0 +1,122 @@
...
+            if (!$matches) {
+                $this->saveFailedOutput($dataset);
+                $this->fail("Output does not match reference rendering (difference on page $page)");
+            }
+        }
+
+        $this->addToAssertionCount(1);
+    }
+
+    protected function compareImages(string $referencePath, string $imageData): bool
```
IonBazan, line 72: This class is `final` so the method should be `private`

tests/OutputTest.php
```diff
@@ -0,0 +1,122 @@
...
+        $dir = new RecursiveDirectoryIterator(Dataset::DATASET_DIRECTORY, $flags);
+        $files = new CallbackFilterIterator(new RecursiveIteratorIterator($dir), $filter);
+
+        foreach ($files as $file) {
+            $dataset = new Dataset($file);
+            yield $dataset->name() => $dataset;
+        }
+    }
+
+    public function outputTestProvider(): Iterator
```
IonBazan, line 37: This could be merged with `datasets()`.

tests/OutputTest/Dataset.php
```diff
@@ -0,0 +1,94 @@
+<?php
+namespace Dompdf\Tests\OutputTest;
+
+use Dompdf\Options;
+use Dompdf\Dompdf;
+use SplFileInfo;
+
+class Dataset
```
IonBazan, line 8: Let's make it `final`

tests/OutputTest/Dataset.php
```diff
@@ -0,0 +1,94 @@
...
+{
+    public const DATASET_DIRECTORY = __DIR__;
+
+    /**
+     * @var SplFileInfo
+     */
+    protected $file;
+
+    public function __construct(SplFileInfo $file)
```
IonBazan, line 17: I think it's good to put a comment here explaining that `$file` is the source HTML file.

IonBazan (approved):
Amazing work here! This will definitely make it easier to keep regressions under control.

Minor comments from me. Also, consider moving the source and output files to `tests/_files`, to separate fixtures and assets from test files.

johanadivare:
> @johanadivare I will look at this next, just want to finish up the on some small updates to selector parsing.

@Mellthas can you give a update?

johanadivare:
> > @johanadivare I will look at this next, just want to finish up the on some small updates to selector parsing.
> 
> @Mellthas can you give a update?

@Mellthas if you dont have time maybe someone else can continue now its stuck for a while

Mellthas:
@johanadivare I am sorry for the long silence. I will see what I can do in the coming days.

Mellthas:
Okay, here we are. I reworked the tests to compare the PDF output produced by the CPDF back end with the help of Ghostscript. The PDF files are converted to images for comparison during testing only now. This should avoid any problems with differing font rendering. @johanadivare This should also avoid the need to use a Docker setup. Ghostscript is needed to run these tests now, and if it is not available, the tests are skipped.

@IonBazan Thanks for the review! I have incorporated your review points, and moved the fixture files as suggested. I also moved the `update-reference-output.php` helper to the `bin` directory.

tests/OutputTest.php
Mellthas: Also applied this suggestion to the other protected methods.

tests/OutputTest.php
Mellthas: Kept these separate because this is the data-provider version of the method. Merging the two would make the usage in `update-reference-output.php` a bit awkward.

Mellthas:
The image comparison uses Imagick if it is available now. It should be a bit faster than using the GD functions. The `Imagick::compareImages()` method also returns an image visualizing the differences, which could be written to a file on failure.

IonBazan (approved):
LGTM! Thanks for picking it up and wrapping it up again.

tests/OutputTest/OutputTest.php
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+
+    protected function setUp(): void
+    {
+        $process = new Process(["gs", "-v"]);
+        $exitCode = $process->run();
+
+        if ($exitCode === 127) {
+            $this->markTestSkipped("Output tests need Ghostscript to be installed");
```
johanadivare, line 61: We could suggest installing `sudo apt-get install ghostscript`
I dont know how we have setup phpunit for skipping test but i think it should definitely fail. is `stopOnSkipped="true"`?

tests/OutputTest/OutputTest.php
```diff
@@ -0,0 +1,209 @@
...
+                "-sOutputFile=-", $file
+            ];
+        };
+        $process1 = new Process($command($referenceFile));
+        $process2 = new Process($command($actualOutputFile));
+
+        $process1->mustRun();
+        $process2->mustRun();
+
+        $error1 = $process1->getErrorOutput();
```
johanadivare, line 104: Does this give output as `mustRun` docs says:
```
The mustRun() method is identical to run(), except that it will throw a [ProcessFailedException](https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Process/Exception/ProcessFailedException.php) if the process couldn't be executed successfully (i.e. the process exited with a non-zero code):
```

tests/OutputTest/OutputTest.php
```diff
@@ -0,0 +1,209 @@
...
+
+        if ($error2 !== "") {
+            throw new RuntimeException("Ghostscript execution failed: $error2");
+        }
+
+        $referenceImages = $this->outputToImageData($process1->getOutput());
+        $actualImages = $this->outputToImageData($process2->getOutput());
+
+        $expectedCount = count($referenceImages);
+        $failureMessage = "Output does not match reference rendering (number of pages is different).";
```
johanadivare, line 119:
```suggestion
        $failureMessage = sprintf('Output (count: `%d`) does not match reference (count: `%d`) rendering (number of pages is different).', $expectedCount, $actualImages);
```

It would be nice to have the numbers in the failure message just like in the default error message for `assertCount`

Also should we not always use single quote `'` instead of double `"` or are there no styling guidelines

tests/OutputTest/OutputTest.php
```diff
@@ -0,0 +1,209 @@
...
+     *
+     * @return string[] A list of the PNG images defined by the output.
+     */
+    private function outputToImageData(string $output): array
+    {
+        $pngSignature = "\x89\x50\x4E\x47\x0D\x0A\x1A\x0A";
+        $elements = explode($pngSignature, $output);
+
+        if (count($elements) <= 1 || $elements[0] !== "") {
+            throw new RuntimeException("Unexpected Ghostscript output: $output");
```
johanadivare, line 142:
```suggestion
            throw new RuntimeException("Unexpected Ghostscript output: `$output`");
```

I think its always nice to put quote around it then you know what the about was because you could get something like this: `Uncaught PHP Exception RuntimeException : "Unexpected Ghostscript output: " at /...` then you dont know if the output was ` ` or it was ` `

johanadivare (changes requested):
Looking good

It would also be nice if there was a way for projects that use this package to use the test in there own project by passing reference files and compare files. But that might be better for new pull request

tests/OutputTest/OutputTest.php
Mellthas: Yes, the message for `ProcessFailedException` contains all the relevant information, so something like:

```
Symfony\Component\Process\Exception\ProcessFailedException: The command "'gs' '-q' '-dBATCH' '-dNOPAUSE' '-sstdout=%stderr' '-sDEVICE=incorrect' '-dGraphicsAlphaBits=4' ...'" failed.

Exit Code: 1(General error)

Working directory: ...

Output:
================


Error Output:
================
Unknown device: incorrect
```

I added a short comment to the error-output handling below and changed the wording, so that it should be more obvious what is happening. Error output could be produced with e.g. the option `-dFirstPage` with a non-exisiting page, which would output something like:

```
Requested FirstPage is greater than the number of pages in the file: 2
   No pages will be processed (FirstPage > LastPage).
```

The command does not fail in that case, in contrast to specifying an incorrect value for`-sDevice`.

tests/OutputTest/OutputTest.php
Mellthas: > We could suggest installing `sudo apt-get install ghostscript`

Added a short note.

> I dont know how we have setup phpunit for skipping test but i think it should definitely fail. is `stopOnSkipped="true"`?

I think we want to have `failOnSkipped="true"`? CI should fail without the output tests, but locally you might want to run the testsuite without Ghostscript, and still have results for all the other tests displayed.

tests/OutputTest/OutputTest.php
Mellthas: I think I left them out because the numbers are contained in the default error message, which is also displayed. But I think you are right, this is nicer. Included you suggestion with a slightly different wording.

Mellthas: > Also should we not always use single quote `'` instead of double `"` or are there no styling guidelines

Other than the `phpcs.xml` configuration, there are no style guidelines that I am aware of. The code base uses double quotes predominantly, though (apart from the `Cpdf` class), so that is what I have been sticking to.

tests/OutputTest/OutputTest.php
Mellthas: Done.

johanadivare (approved):

tests/OutputTest/OutputTest.php
bsweeney: Yeah, no real styling guide. I think the majority of changes in recent years has leaned towards double quotes. I believe primarily due to the decrease in performance differential between the two and to ensure more consistent code presentation.

Mellthas:
Okay, I have looked over the initial set of test cases and added some new ones for border rendering. I think this is good to go now. Further enhancements can be done in new PRs.

bsweeney (approved):

Issue #476, opened 2012-12-06
min-height & max-height not working


PhenX:
_Original author: johngbel...@gmail.com (April 24, 2012 19:26:20)_

<b>What steps will reproduce the problem?</b>
&lt;div style=&quot;min-height:100px&quot;&gt;...&lt;/div&gt;

<b>What is the expected output? What do you see instead?</b>
expected: 100px high div
seen: div shrunk to content

<b>What version of dompdf are you using? What version of PHP? On what</b>
<b>operating system?</b>
Trunk, Php 5.3.10, Centos

<b>Please provide the HTML source code you want to convert, or any additional</b>
<b>information.</b>

will work on patch

_Original issue: http://code.google.com/p/dompdf/issues/detail?id=470_

bsweeney:
patch files from original issue on Google Code:

block_frame_reflower.patch:
```patch
# This patch file was generated by NetBeans IDE
--
# Following Index: paths are relative to: C:\Users\john\Documents\dev\dompdf\dompdf\include
# This patch can be applied using context Tools: Patch action on respective folder.
# It uses platform neutral UTF-8 encoding and \n newlines.
# Above lines and this line are ignored by the patching process.
Index: block_frame_reflower.cls.php
--- block_frame_reflower.cls.php Base (BASE)
+++ block_frame_reflower.cls.php Locally Modified (Based On LOCAL)
@@ -224,7 +224,7 @@
* Determine the frame's restricted height
* @return array
*/
-  protected function _calculate_restricted_height() {
+  protected function _calculate_height() {
$frame = $this->_frame;
$style = $frame->get_style();
$content_height = $this->_calculate_content_height();
@@ -349,49 +349,12 @@
} else {
 
// Expand the height if overflow is visible
-      if ( $height === "auto" && $content_height > $height /* && $style->overflow === "visible" */)
+      if ( $height === "auto" && $content_height > $height  && $style->overflow === "visible" )
$height = $content_height;
 
-      // FIXME: this should probably be moved to a seperate function as per
-      // _calculate_restricted_width
-
-      // Only handle min/max height if the height is independent of the frame's content
-      if ( !($style->overflow === "visible" \|\|
-             ($style->overflow === "hidden" && $height === "auto")) ) {
-
-        $min_height = $style->min_height;
-        $max_height = $style->max_height;
-
-        if ( isset($cb["h"]) ) {
-          $min_height = $style->length_in_pt($min_height, $cb["h"]);
[...]

PR #2511, opened 2021-08-04
BlockFrameReflower: Fixes to height calculations

Mellthas:
Brings the code closer to the spec, basically by only removing code:
1. Always resolve `auto` heights
2. Apply `min-/max-height` independently of the `overflow` property value
3. `min-height` overrules `max-height` (swapping removed)

Partially addresses #476.

The first point fixes the following HTML, which resulted in an empty page before:
```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style type="text/css">
@page {
	size: 400pt 300pt;
	margin: 50pt;
}

.box {
	padding-top: 200pt;
	background-color: red;
}
</style>
</head>

<body>

	<div class="box"></div>

</body>

</html>
```

### DIFF
--- a/src/FrameReflower/Block.php
+++ b/src/FrameReflower/Block.php
@@ -374,48 +374,48 @@ class Block extends AbstractFrameReflower
             }
 
         } else {
-            // Expand the height if overflow is visible
-            if ($height === "auto" && $content_height > 0 /* && $style->overflow === "visible" */) {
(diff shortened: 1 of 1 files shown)


Mellthas:
This does not add complete support for `min-/max-height`, it is just a quick fix which enables it for all block containers which are not absolutely positioned.

bsweeney (approved):

Issue #2189, opened 2020-07-05
Page breaking on table


hhjung2013:
I am trying to bring a weekday table into a PDF. It probably won't fit on one landscape page, so I am looking for a way to have more control over where to do the page break. I have tried `page-break-inside: avoid` on the table and `page-break-after: always` on the row for testing but both didn't have any effect at all.

Here is my basic code:
```
<!DOCTYPE html>
<html>
<head>
<link type="text/css" rel="stylesheet"
	href="https://www.devah.de/wp-content/themes/infinity-pro/style-pdf.css">
</head>
<body>
	<table class="dv-kursplan kp-large">
		<thead>
			<tr></tr>
		</thead>
		<tbody>
			<tr></tr>
			<tr></tr>
			<tr></tr>
		</tbody>
	</table>
</body>
</html>
```
```
table {
	page-break-inside: avoid;
}	

.dv-kursplan {
	background-color: #eee;
	font-family: sans-serif;
	font-size: 1.2rem;
}

.dv-kursplan th {
	text-align: center;
	border: 1px solid #fff;
	font-weight: normal;
}
	
.dv-kursplan td {
	width: 14%;
	padding: 0;
	border: 1px solid #fff;
}
```

The full page is available here:
https://www.devah.de/pdf-test/

This is the rendered PDF result:
[devah_kursplan(4).pdf](https://github.com/dompdf/dompdf/files/4875710/devah_kursplan.4.pdf)

Any help on how to apply page breaking is greatly appreciated.

bsweeney:
If I recall correctly Dompdf does not handle page break declarations on table elements (#404). I'd have to review the current state of the code to say for certain if that has changed at all.

hhjung2013:
This post made me think it was possible:
https://stackoverflow.com/questions/8723487/dompdf-page-break-if-element-is-exceeding-page-height
If you could check the code for clarity it would be very helpful.

If page breaking inside of tables is not supported, I guess using table-styled divs would be the way to go. Any other ideas?

bsweeney:
It's possible to use the page-break declarations, just not on table elements. What I would have to review is if any work has been done to address that deficiency, but I do not believe so.

If you want this type of layout now using table-styled divs would probably work best. Though, I'd create a sample document before you invest too much time into it.

hhjung2013:
But the above stackoverflow topic is about page breaks on tables, isn't it?

[...]

Issue #404, opened 2012-12-06
Honor page-break styling on table grouping elements


PhenX:
_Original author: eclecticgeek (December 22, 2011 19:37:28)_

Background:
We had a request to provide some kind of option to allow for row groups to avoid being split between pages (ref Grouping certain table rows on one page, https://groups.google.com/forum/#!topic/dompdf/HfWt70SB0OU). The most obvious solution is for the user to surround rows with a tbody element (multiple tbody elements would define multiple row groups). The tbody element would then be styled page-break-inside: avoid.

Defect:
page-break-inside: avoid is not honored for thead, tbody, and tfoot elements. This might seem to be appropriate, tbody is a grouping element and a table layout element (http://www.w3.org/TR/html401/struct/tables.html#edef-TBODY). However, the CSS2.1 spec allows for page-break-inside to be applied to &quot;to other elements, e.g., 'table-row' elements&quot; (http://www.w3.org/TR/CSS2/page.html#page-break-props).

Resolution:
Honor page-break styling on elements that behave similar to block-level elements (e.g. tr, tbody, thead).

<details>
<summary>sample html</summary>

```
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
--
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
<title>Issue 398</title>
<style>
tbody, p { page-break-inside: avoid; }
</style>
 
</head>
 
<body>
 
<div>
<table>
<tbody>
<tr><th>h1c1</th><th>h1c2</th><th>h1c3</th><th>h1c4</th></tr>
<tr><td>r1c1</td><td>r1c2</td><td>r1c3</td><td>r1c4</td></tr>
<tr><td>&nbsp;</td><td>&nbsp;</td><td>&nbsp;</td><td>&nbsp;</td></tr>
<tr><td>r3c1</td><td>r3c2</td><td>r3c3</td><td>r3c4</td></tr>
</tbody>
<tbody>
<tr><th>h1c1</th><th>h1c2</th><th>h1c3</th><th>h1c4</th></tr>
<tr><td>r1c1</td><td>r1c2</td><td>r1c3</td><td>r1c4</td></tr>
[...]

PR #2512, opened 2021-08-04
Add support for `page-break-before/-after` on table rows

### LINKED ISSUE #329: Page-break on tr element
_Original author: nikosp...@gmail.com (July 08, 2011 07:23:20)_

<b>What would you like dompdf to do:</b>
Be able to apply page-break-before: always on tr elements as described here: http://www.w3.org/TR/CSS2/page.html#page-break-props

<b>Do you have an example:</b>
&lt;table&gt;
  &lt;tr&gt;
    &lt;td&gt;&lt;/td&gt;
    &lt;td&gt;&lt;/td&gt;
  &lt;/tr&gt;
  &lt;tr style=&quot;page-break-before: always;&quot;&gt;
    &lt;td&gt;&lt;/td&gt;
    &lt;td&gt;&lt;/td&gt;
  &lt;/tr&gt;
&lt;/table&gt;

_Original issue: http://code.google.com/p/dompdf/issues/detail?id=323_

Mellthas:
Caveat: `page-break-after: always` does not work if the row is the last of its table row group. I want to keep things simple now; this might be better fixed in a proper rework of the page-break-properties handling, e.g. by properly propagating the property values as per https://www.w3.org/TR/css-break-3/#break-propagation.

Fixes #329
Might address #2189


bsweeney:
@Mellthas what do you think about applying the logic to other table grouping objects, such as `tbody`, `thead`, and `tfoot` as requested in #404?

bsweeney:
fyi related to 2189 it looks like `page-break-*` declarations on a table and row work as expected. That is so long as they're performed in isolation. Conflicting rules can be specified (e.g. avoid table break but always break after a row) and I haven't done enough research to say if we're handling those edge cases correctly.

Mellthas:
Support for table grouping elements would definitely be nice. I though about it while implementing the changes here, but then realized that it might better be handled by a more thorough refactor of the code, as there is quite a bit of overlap between the different paths in `_page_break_allowed` now, albeit with slight differences related to e.g. `page-break-inside: avoid`, and it is a bit hard to find your way around in the code. You'd also have to check the impact breaking before table grouping elements would have on rendering, e.g. with a separated table border model. So restricting it to table rows here was simply the lowest impact change, which is quite useful at the same time.

Mellthas:
Conflicting rules are already handled properly I think. Forced breaks take precedence, as they are handled separately before unforced ones. See https://www.w3.org/TR/css-break-3/#forced-breaks for the spec.

bsweeney:
I think that's fine. We could shim tbody handling into the table-row logic with minimal changes. However, it's not a widely requested feature so maybe adding it while refactoring is the way to go so that we avoid complicating the logic even more.

PR #2547, opened 2021-09-12
Small code cleanups

Mellthas:
Split out from #2497.

### DIFF
--- a/src/Frame.php
+++ b/src/Frame.php
@@ -465,42 +465,12 @@ class Frame
     //........................................................................
 
     /**
-     * Return the height of the margin box of the frame, in pt.  Meaningless
-     * unless the height has been calculated properly.
-     *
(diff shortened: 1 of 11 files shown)


bsweeney:
I see why I never noticed that there was an error rendering surrogate code points before ... no type checking. Plus I wasn't actually reviewing the full document (shame on me) after render. The error I get is "Malformed UTF-8 characters, possibly incorrectly encoded" (PREG_BAD_UTF8_ERROR).

I _was_ thinking it would be better to return the original string, but because text is parsed by RegEx elsewhere the same issue is encountered in multiple locations. I'll open a new issue as a reminder to investigate but the error parsing surrogate code points doesn't seem an urgent issue at this point.

bsweeney (approved):

PR #2549, opened 2021-09-12
Use shrink-to-fit width resolution for float and inline-block

Mellthas:
As per spec.

<del>The problem right now is that generated content is only set on reflow of the actual frame, so it is not yet set for any descendants when `get_min_max_content_width` is called. As a result, generated content is not honored when calculating the minimum and maximum width. This  is very apparent with elements that only contain generated content, e.g. an `input` button, as reported in https://github.com/dompdf/dompdf/pull/2497#issuecomment-917504662.</del>

<del>Applies on top of #2548.</del>
Applies on top of #2559.

### DIFF
--- a/src/FrameDecorator/AbstractFrameDecorator.php
+++ b/src/FrameDecorator/AbstractFrameDecorator.php
@@ -930,7 +930,7 @@ abstract class AbstractFrameDecorator extends Frame
     /**
      * @return array
      */
-    final function get_min_max_width()
+    final function get_min_max_width(): array
     {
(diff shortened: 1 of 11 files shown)


Mellthas:
This now applies on top of #2559 and includes another change to set generated content early in `get_min_max_content_width`. This ensures that the shrink-to-fit width honors generated content, which also affected table column widths.

Mellthas:
@bsweeney, can you have a look at this for 1.1.0, now that generated content is accounted for?

bsweeney (approved):

PR #2559, opened 2021-09-19
Clean up handling of counters involving page breaks

### LINKED ISSUE #2167: Outlined numbering of CSS counters do not properly render across page breaks.
There is a problem with outlined nested numbering being properly rendered over page breaks in DomPDF version 0.8.5.  I have attached a text file of the HTML used to create the attached PDF output.  Take notice the numbering does not continue from "3.2" to "3.3", but instead restarts at "3.1".

I've gone thru the code, but can't seem to figure it out:  I'm sorry that I cannot propose an answer to fix this issue.  However, I do have an inkling that this bug has something to do with the numbering counters being reset upon the adding of a new page to the PDF (as opposed to being preserved).

[numbering-problem.txt](https://github.com/dompdf/dompdf/files/4754893/numbering-problem.txt)
[numbering-problem.pdf](https://github.com/dompdf/dompdf/files/4754861/numbering-problem.pdf)

**Notes:**  I've blacked-out a little bit of the PDF and replaced that text with blanks in the HTML.  Also, the red text is added _after_ PDF generation so you won't find that in the HTML code.

### LINKED ISSUE #2558: Counter value off after page break
The `li` counter is incremented one extra time after the page break in the following sample:

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
    size: 400pt 300pt;
    margin: 50pt;
}

li {
    counter-increment: li;
}

li::before {
    content: counter(li);
}
</style>
</head>

<body>
<ul>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li><ul><li></li></ul></li>
<li></li>
<li></li>
<li></li>
</ul>
</body>

</html>
```

Mellthas:
The `content_set` property is introduced in preparation for allowing to set generated content early when calculating shrink-to-fit widths.

Fixes #2167
Fixes #2558

### DIFF
--- a/src/FrameDecorator/AbstractFrameDecorator.php
+++ b/src/FrameDecorator/AbstractFrameDecorator.php
@@ -95,6 +95,13 @@ abstract class AbstractFrameDecorator extends Frame
     private $_cached_parent;
 
     /**
+     * Whether generated content and counters have been set.
+     *
+     * @var bool
+     */
+    public $content_set = false;
+
+    /**
      * Whether the frame has been split
      *
      * @var bool
@@ -147,7 +154,7 @@ abstract class AbstractFrameDecorator extends Frame
      *
      * @param DOMNode $node
      *
-     * @return Frame
+     * @return AbstractFrameDecorator
      */
     function copy(DOMNode $node)
     {
@@ -160,7 +167,7 @@ abstract class AbstractFrameDecorator extends Frame
     /**
      * Create a deep copy: copy this node and all children
      *
-     * @return Frame
+     * @return AbstractFrameDecorator
      */
     function deep_copy()
     {
@@ -183,13 +190,13 @@ abstract class AbstractFrameDecorator extends Frame
         return $deco;
     }
 
-    /**
-     * Delegate calls to decorated frame object
-     */
     function reset()
     {
         $this->_frame->reset();
+        $this->reset_generated_content();
+        $this->revert_counter_increment();
 
+        $this->content_set = false;
         $this->_counters = [];
 
         $this->_cached_parent = null; //clear get_parent() cache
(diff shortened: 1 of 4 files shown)


src/FrameDecorator/AbstractFrameDecorator.php
```diff
@@ -171,13 +179,13 @@ function deep_copy()
...
     }
 
-    /**
-     * Delegate calls to decorated frame object
-     */
     function reset()
     {
         $this->_frame->reset();
+        $this->reset_generated_content();
+        $this->revert_counter_increment();
```
Mellthas, line 186: Reverting incremented counters on frame reset handles cases where a child that already had a reflow (and thus counter increment) is moved to the next page.

src/FrameDecorator/AbstractFrameDecorator.php
```diff
@@ -749,11 +766,9 @@ public function split(Frame $child = null, bool $force_pagebreak = false)
...
         $this->get_parent()->split($split, $force_pagebreak);
 
-        // If this node resets a counter save the current value to use when rendering on the next page
-        if ($style->counter_reset && ($reset = $style->counter_reset) !== "none") {
-            $vars = preg_split('/\s+/', trim($reset), 2);
-            $split->_counters['__' . $vars[0]] = $this->lookup_counter_frame($vars[0])->_counters[$vars[0]];
-        }
+        // Preserve the current counter values. This must be done after the
+        // parent split, as counters get reset on frame reset
+        $split->_counters = $this->_counters;
```
Mellthas, line 771: This also handles cases where the split element does not reset the counter itself. The counter reset is then prevented on the split node by setting `counter-reset` to `none` above.

src/FrameDecorator/AbstractFrameDecorator.php
```diff
@@ -694,29 +707,33 @@ public function split(Frame $child = null, bool $force_pagebreak = false)
...
+            $split_style->margin_top = 0;
+            $split_style->padding_top = 0;
+            $split_style->border_top = 0;
+            $split_style->page_break_before = "auto";
         }
 
+        $split_style->text_indent = 0;
+        $split_style->counter_reset = "none";
+
+        $split->set_style(clone $split_style);
```
Mellthas, line 733: This is what `$split->reset()` effectively did. All the other resets are unnecessary because a newly decorated frame is created for the split node.

bsweeney (approved):
This is a much cleaner process for resetting counters. 🎉

PR #2569, opened 2021-09-22
Fix justification of last line before page break

Mellthas:
Before 1b8f3edc3f8d0e60914dc4d66fa200e87904a722, lines were not removed from line boxes after a split, and thus the actual last line was still contained in the list of of the original frame’s line boxes, so the issue was essentially masked.

I think the description of `$_splitted` was misleading, does the wording make sense now in contrast to `$is_split`? I did not rename the property to preserve backwards compatibility, though it would be nice to clean that up, too.

### DIFF
--- a/src/Frame.php
+++ b/src/Frame.php
@@ -160,7 +160,7 @@ class Frame
     public $_float_next_line = false;
 
     /**
-     * Tells whether the frame was split
+     * Whether the frame is a split-off frame
      *
      * @var bool
      */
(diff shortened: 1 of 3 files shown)


bsweeney:
I think the updated description and naming makes sense. I can't imagine `$_splitted` is being used outside the code, but I agree that since it's public we should preserve it for compatibility.

bsweeney (approved):

PR #2578, opened 2021-09-29
Clear border radius between split frames

jorycadence:
When frames are split across pages, the border radius is still applied at the split point:

Screen Shot 2021-09-29 at 3 05 15 PM

PDF document: [before.pdf](https://github.com/dompdf/dompdf/files/7255034/before.pdf)

This PR introduces a change such that border radius is cleared at the split point: 

Screen Shot 2021-09-29 at 3 05 38 PM
PDF document: [after.pdf](https://github.com/dompdf/dompdf/files/7255038/after.pdf)

This can be reproduced with the following: 

HTML markup: [example-html.txt](https://github.com/dompdf/dompdf/files/7255021/examplehtml.txt)
Config: default


Mellthas:
Nice!

Mellthas (approved):

Issue #2570, opened 2021-09-23
DIV Strange page-break if doesn't fit on a page.


Nagrom33:
I have a problem if my `<div>` doesn't fit, than i got an strange page break.
I think it's something with the padding. Can someone help me out?

HTML:
```
<div class="info">
      <h2 class="font--xxl margin--m">Creatable checkboxes tooltip</h2>
      <p class="font--m margin--m">Lorem creatable checkbox lorem</p>
</div>
```

CSS:
```
.info {
  width: 570px;

  padding: 32px 48px 40px;
  border-radius: 8px;
  margin-top: 24px;

  color: #000;
  background: #d9ebf7;

  page-break-inside: avoid;
}
```

Example of export:
<a href="https://ibb.co/ZdycfsV"><img src="https://i.ibb.co/TczPwXk/Schermafbeelding-2021-09-23-om-14-15-13.png" alt="Schermafbeelding-2021-09-23-om-14-15-13" border="0"></a>

bsweeney:
FYI easier to debug with a full example, this seems to do it:

```
<style>
    @page { size: 650px 185px; margin: 0px; }
    .info {
      width: 570px;
    
      padding: 32px 48px 40px;
      border-radius: 8px;
      margin-top: 24px;
    
      color: #000;
      background: #d9ebf7;
    
      page-break-inside: avoid;
    }
</style>

<div class="info">
        <h2 class="font--xxl margin--m">Creatable checkboxes tooltip</h2>
        <p class="font--m margin--m">Lorem creatable checkbox lorem</p>
</div>
```

There does seem to be an issue here since the containing element is pushed to the next page but still has a presence on the previous page. There also seems to be an issue determining whether or not it's appropriate to push to the next page (as in my example where it's the only element in the document).

The upcoming 1.1.0 is much improved in this regard but still appears to produce a suboptimal rendering.

Mellthas:
`page-break-inside` is only checked if the frame has a previous sibling. This should be addressed by #2571.

Nagrom33:
Thanks @Mellthas this works:

```
.info {
  page-break-inside: avoid !important;
}

.info > *:first-child {
  page-break-before: avoid !important;
}

.info > *:last-child {
    page-break-after: avoid !important;
}
```

PR #2580, opened 2021-10-04
Tweak gap detection when checking for allowed page breaks

Mellthas:
This is to resolve cases like in #2570, where the page breaks before the first block-level child of a block container. The current code basically checks for a gap between the *padding edge* of the parent and the *border edge* of the child when the spec text demands a gap between *content edge* and *margin edge* resp. See the following sample for an illustration of the current issue. Especially without borders or outlines it can lead to unexpectedly leaving the (invisible) padding on a previous page:

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
    size: 400pt 300pt;
    margin: 50pt;
}

body {
    outline: 1pt solid black;
}

.block {
    margin-bottom: 20pt;
    outline: 2pt solid red;
}

.block div {
    height: 160pt;
    background-color: lightblue;
}

.padding {
    padding-top: 20pt;
}
</style>
</head>

<body>

<div class="block">
    <div></div>
</div>

<div class="block padding">
    <div></div>
</div>

</body>

</html>
```
</details>

[padding-before.pdf](https://github.com/dompdf/dompdf/files/7277679/padding-before.pdf)
[padding-after.pdf](https://github.com/dompdf/dompdf/files/7277680/padding-after.pdf)

https://www.w3.org/TR/CSS21/page.html#allowed-page-breaks
https://www.w3.org/TR/CSS21/box.html#box-dimensions


Mellthas:
The following sample is I think a case where an actual gap occurs. The floating box pushes the text down below the content edge of its parent frame:

<details>
<summary>Actual gap</summary>

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
	size: 400pt 300pt;
	margin: 50pt;
}

body {
	font-family: sans-serif;
	letter-spacing: 0.8pt;
	outline: 1pt solid black;
	line-height: 19pt;
}

p {
	margin: 0 0 20pt 0;
	padding: 10pt 0;
	border: 5pt solid black;
	orphans: 1;
	widows: 1;
}

.float {
	display: block;
	float: right;
	width: 96%;
	height: 30pt;
	background-color: red;
}
</style>
</head>

<body>
	<p>Lorem ipsum dolor sit amet, consectetur adipisici elit, sed eiusmod tempor
	incidunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis
	nostrud exercitation ullamco laboris nisi ut aliquid ex ea commodi consequat.</p>
	<span class="float"></span>
</body>

</html>
```
</details>

This is the inline case though, which is not currently handled by the code. The break is allowed there because it seems the text is pushed down via empty line boxes. Checking this in Chrome and Firefox, the text is pushed just enough so that the float fits.

bsweeney (approved):

PR #2583, opened 2021-10-05
reset counter parent instead create local one

eltharin:
Hi I had a problem in counter numbers, when they were reset, after one page they were erased, I found this, before that create on counter in the frame instead take parent's one.

### DIFF
--- a/src/FrameDecorator/AbstractFrameDecorator.php
+++ b/src/FrameDecorator/AbstractFrameDecorator.php
@@ -755,7 +755,7 @@ abstract class AbstractFrameDecorator extends Frame
      */
     function reset_counter($id = self::DEFAULT_COUNTER, $value = 0)
     {
-        $this->get_parent()->_counters[$id] = intval($value);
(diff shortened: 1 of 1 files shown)


src/FrameDecorator/AbstractFrameDecorator.php
```diff
@@ -755,7 +755,7 @@ function split(Frame $child = null, $force_pagebreak = false)
      */
     function reset_counter($id = self::DEFAULT_COUNTER, $value = 0)
     {
-        $this->get_parent()->_counters[$id] = intval($value);
+        $this->lookup_counter_frame($vars[0])->_counters[$id] = intval($value);
```
Mellthas, line 758: You probably meant `$id` there instead of `$vars[0]`.

Mellthas:
Please check whether you case has not already been fixed on the `develop` branch. Judging from the requested change, it might be the same as #2167, which is already fixed on `develop`. While your change fixes the sample from that issue, it breaks nested counters, so it is not the right thing to do.

eltharin:
yes $this->lookup_counter_frame($id) ... sorry

Issue #682, opened 2013-07-08
Large inline element pages incorrectly (?)


bsweeney:
To be honest, I don't know what's going on here. But the following breaks rendering. I tried some simplified markup and could not reproduce the issue.

The following text pages at the start of the italics. It appears to be treating the italics element as a positioned node (sort of, but not really).

``` html
<!DOCTYPE html>
<html>
<head>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
</head>
<body>

To better understand why this course of action was necessary, 
A'th'n learned why<br> <br> it was the responsibility, shame of each of the incarnations for ridding this 
world of<br> <br> this threat.<br> <br> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; H'br'a, spoken 
through A'yd'a, explained the CHTHONIAN collective and why, since<br> <br> A'yd'a through T'li'a, H'br'a, now, A'th'n it had to come to this point.<br> <br> <br> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; 
<i>Since her initial landfall, every time that the Primae had released a life force<br> <br> from a body she 
killed, this being the only logical solution to correct the crisis,<br> <br> the spirit had nowhere to go. <br> 
<br> It hung in the ether seeking spiritual release. Never finding any, the spirits soon <br> <br> learned that 
joining together made them capable of<br> <br> creating entities that could exist, again, in the living world, 
albeit as Sapen<br> <br> mockeries. <br> <br> Learning swiftly, the collected spirits came to the 
realization that not<br> <br> only could they exist on the living plane they could alter it.<br> <br> 
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; It was this "awakening" that started their raiding, terrorizing, 
destroying all<br> <br> living beings they encountered.<br> <br> 
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; The end result of their intermingling was to, first, confuse, next, 
annihilate<br> <br> the outer husks of the tortured to absorb them into their collective existence.<br> <br 
/> <br> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; In time, the first few spirits realized they were capable of 
manipulation,<br> <br> allowing them to become very adept at their methods with the one goal being 
total<br> <br> domination of all species, races.<br> <br> Keeping their combined numbers small enabled 
[...]

PR #2589, opened 2021-10-08
Some fixes around frame splitting

Mellthas:
* Make sure to always hand through the `$force_pagebreak` parameter, so that inline elements properly split when needed.
* Rename `$force_pagebreak` to `$page_break`, to make clearer what it is about.
* Add a `$forced` parameter to be able to tell the difference.
* Recursively split inline frames and split inline frames on `br`, so that large nested inline elements are properly paged.
* Checking for forced page breaks on inline elements is not covered by the spec, and the actual check in `FrameDecorator\Inline` is pretty funky, leading to some funny output when defining forced breaks on nested inline elements. Just remove the check there.
* Clear top margins after an unforced page break, per spec. Clearing bottom margins/improving the page-break check is still TODO. See https://www.w3.org/TR/CSS21/page.html#allowed-page-breaks.

I think it would be event better to `split` (!) the method itself into two, one for page-break split and one for line-box split (`FrameDecorator\Text` already has a dedicated method `split_text`). This would make clearer what it does. But for now, just a small step here.

Addresses #682
Addresses #2347

### DIFF
--- a/src/FrameDecorator/TableRowGroup.php
+++ b/src/FrameDecorator/TableRowGroup.php
@@ -32,8 +32,8 @@ class TableRowGroup extends AbstractFrameDecorator
     }
 
     /**
-     * Override split() to remove all child rows and this element from the
-     * cellmap
+     * Split the row group at the given child and remove all subsequent child
+     * rows and all subsequent row groups from the cellmap.
      */
     public function split(?Frame $child = null, bool $page_break = false, bool $forced = false): void
     {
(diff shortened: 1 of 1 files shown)


Mellthas:
Sample which is partially addressed by clearing the top margin:

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
	size: 400pt 300pt;
	margin: 50pt;
}

body {
	outline: 0.5pt solid black;
}

.block {
	page-break-inside: avoid;
	margin: 10pt 0;
	background-color: lightblue;
	height: 40pt;
}
</style>
</head>

<body>

<div class="block">
</div>

<div class="block">
</div>

<div class="block">
</div>

<div class="block">
</div>

<div class="block">
</div>

<div class="block">
</div>

<div class="block">
</div>

<div class="block">
</div>

</body>

</html>
```

</details>

Output [before.pdf](https://github.com/dompdf/dompdf/files/7307599/before.pdf)
Output [after.pdf](https://github.com/dompdf/dompdf/files/7307601/after.pdf)

I think ideally, we want four boxes per page here.

Mellthas:
Added a small fix for bottom table cell borders not being drawn in some cases on top, as it relates to the split logic.

Issue #497, opened 2012-12-06
table column width incorrectly accounts for non-flowed content width


PhenX:
_Original author: eclecticgeek (May 31, 2012 02:29:50)_

<b>What steps will reproduce the problem?</b>
Create a table with two cells where one cell contains only an absolutely positioned element.

<b>What is the expected output? What do you see instead?</b>
When rendered the column with the positioned element should render with a width similar that that of an empty cell. Instead it is rendered to the width of the positioned element.

Originally reported in the support group: https://groups.google.com/d/topic/dompdf/OkpARUqiRQE/discussion

Confirmed in trunk r493.

Simple test case:
&lt;table border=&quot;1&quot;&gt;&lt;tr&gt;
&lt;td&gt;&lt;div style=&quot;position: absolute;&quot;&gt;positioned div with some text&lt;/div&gt;&lt;/td&gt;
&lt;td&gt;&lt;/td&gt;
&lt;td&gt;A normal table cell with a little bit of text&lt;/td&gt;
&lt;/tr&gt;&lt;/table&gt;

_Original issue: http://code.google.com/p/dompdf/issues/detail?id=491_

PR #2596, opened 2021-10-11
Consider all inline-level types in `get_min_max_width`

Mellthas:
Quick fix to get more accurate min/max widths when inline-block, inline-table, or image elements are involved.


Mellthas:
Updated to also exclude children with absolute position, which addresses #497.

PR #2651, opened 2021-11-21
Clean up `add_frame_to_line`

### LINKED ISSUE #1259: Incorrect indent after pagebreak with inline-block elements.
# Problem

When rendering a list of DIVS that shouldnt break inside`(page-break-inside: avoid)`, and inside the DIV there are two DIVS put next to each other using `display:inline-block`, the first item on the next page will have an unintended left indent. 
# Steps to reproduce
## HTML

```
<html>
<link rel="stylesheet" type="text/css" href="test.css">
<body>

<article>
    {{ repeat the following HTML x 100 }}
    <div class="item">
        <div class="label">
            <div class="label">Lorem</div>
            <div class="description">Lorem</div>
        </div>
        <div class="text">Dolet
        </div>
    </div>
    {{ // end repeat }}
</article>
```
## CSS

```
article .item {
    page-break-inside: avoid;
}
article .item .text {
    width: 200px;
    display: inline-block;
}
article .item .label {
    width: 200px;
    display: inline-block;
}
```
## PHP

```
require_once("dompdf/autoload.inc.php");
$html = file_get_contents('test.html');

$dompdf = new \Dompdf\Dompdf();
$options = new \Dompdf\Options();
$options->setIsRemoteEnabled(TRUE);
$dompdf->setOptions($options);

$dompdf->load_html($html);
$dompdf->set_paper('A4');
$dompdf->render();
$dompdf->stream('pdf.pdf');
```
# Screenshot

image

### LINKED ISSUE #2638: Leading spaces on same line as tag are lost with 'white-space:pre-wrap'
I am using PHP 7.4.16 and dompdf/dompdf 1.0.2 plus the fixes in [Pull Request 2584](https://github.com/dompdf/dompdf/pull/2584) by @Mellthas.

When formatting paragraphs with `white-space: pre-wrap`  in some cases a leading space on the same line as a paragraph tag is not being preserved.

Example HTML (the monospace part is irrelevant to this issue - it makes it easier to check examples with leading spaces):
```
<style>
	p.monospace-pre-wrap {
        font-family: Consolas, Courier, monospace;
        white-space: pre-wrap;
      }
</style>

<p class="monospace-pre-wrap"> Leading space on same line as tag in single-line paragraph is lost.</p>
<p class="monospace-pre-wrap">
 Leading space on separate line from tag is OK.
</p>
<p class="monospace-pre-wrap"> Leading space on same line as tag in multi-line paragraph is not lost
 Second line.
</p>
<p class="monospace-pre-wrap"> <u>Leading space with underline</u> is lost again
 Second line.
</p>
```
In the PDF it looks like this:
image

I think it might be related to the bit in [FrameReflower/Text.php](https://github.com/dompdf/dompdf/blob/fcf6583887fce4b9a779d5be7e84789490b1c922/src/FrameReflower/Text.php) around line 350 where it does an ltrim
```
            if ((!$is_inline_frame && !$frame->get_prev_sibling()) /* ||
            ( $is_inline_frame && !$parent->get_prev_sibling())*/
            ) { //  <span><span>A<span>B</span> C</span></span> fails (the whitespace is removed)
                $t = ltrim($t);
            }
```
but I have not managed to work out why some of these examples are trimmed and not others.

Mellthas:
The <del>first</del> second commit is a followup to #2647, which breaks the following test case:

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
    size: 400pt 300pt;
    margin: 50pt;
}

div {
    outline: 2pt solid red;
}

p {
    margin: 0;
    height: 100pt;
    background-color: lightblue;
}
</style>
</head>

<body>
    <div>
        <p>First</p>
        <p>Second</p>
        <p>Third</p>
        <p>Fourth</p>
    </div>
</body>

</html>
```
</details>

[line-boxes-before.pdf](https://github.com/dompdf/dompdf/files/7577042/line-boxes-before.pdf)
[line-boxes-after.pdf](https://github.com/dompdf/dompdf/files/7577043/line-boxes-after.pdf)

The zero-width check in the `add_frame_to_line` method was not only used for handling wrapped white space, but also for preventing frames which were moved to the next page from being added to the line box on the previous page. This worked because of the style reset after a page break, unless they had a specified width, in which case the check failed (can be observed by adding a specific width to the `p` elements in the test case).

Following that, the second commit cleans up the white-space-trimming logic, which includes addressing #2638.

Fixes #1259
Fixes #2638

### DIFF
--- a/src/Cellmap.php
+++ b/src/Cellmap.php
@@ -732,7 +732,7 @@ class Cellmap
     {
         $key = $row->get_id();
         if (!isset($this->_frames[$key])) {
-            return; // Presumably this row has alredy been removed
+            return; // Presumably this row has already been removed
         }
 
         $this->__row = $this->_num_rows--;
@@ -780,7 +780,7 @@ class Cellmap
     {
         $key = $group->get_id();
         if (!isset($this->_frames[$key])) {
-            return; // Presumably this row has alredy been removed
+            return; // Presumably this row has already been removed
         }
 
         $iter = $group->get_first_child();
@@ -802,10 +802,15 @@ class Cellmap
     public function update_row_group(Frame $group, Frame $last_row)
     {
         $g_key = $group->get_id();
-        $r_key = $last_row->get_id();
 
-        $r_rows = $this->_frames[$g_key]["rows"];
-        $this->_frames[$g_key]["rows"] = range($this->_frames[$g_key]["rows"][0], end($r_rows));
+        $first_index = $this->_frames[$g_key]["rows"][0];
+        $last_index = $first_index;
+        $row = $last_row;
+        while ($row = $row->get_prev_sibling()) {
+            $last_index++;
+        }
+
+        $this->_frames[$g_key]["rows"] = range($first_index, $last_index);
     }
 
     /**
--- a/src/FrameDecorator/Block.php
+++ b/src/FrameDecorator/Block.php
@@ -164,7 +164,7 @@ class Block extends AbstractFrameDecorator
 
         // FIXME: Why? Doesn't quite seem to be the correct thing to do,
         // but does appear to be necessary. Hack to handle wrapped white space?
-        if ($w == 0 && $frame->get_node()->nodeName !== "hr" && !$frame->is_pre()) {
+        if ($w === 0.0 && $frame->is_text_node() && !$frame->is_pre()) {
             return;
         }
 
--- a/src/FrameDecorator/Block.php
+++ b/src/FrameDecorator/Block.php
@@ -108,12 +108,6 @@ class Block extends AbstractFrameDecorator
      */
     function add_frame_to_line(Frame $frame)
     {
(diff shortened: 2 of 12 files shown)


Mellthas:
Includes the zero-width fix now, which I reverted on `master`, so that we can get 1.1.1 out.

bsweeney:
Considering the regression should I go ahead and review this PR (and the other one?) for inclusion in 1.1.1?

Mellthas:
I am a bit confused, I reverted the commit in da6a2f4fa55ab431b5001bebb50d9ac85e57f2c3 so that the regression is gone from master? If you have the time for a review though, I would certainly like to see them being included in a release soon :-).

bsweeney:
Sorry, no, it's me who was confused (lack of sleep and all that).

bsweeney (approved):

PR #2659, opened 2021-11-28
file_get_contents use_include_path parameter set from null to false a…

Madnx:
If I'm not mistaken, dompdf should set the second parameter of file_get_contents as false instead of null. Can't see why it was null in the first place but it's now deprecated and displays warnings.

### DIFF
--- a/src/Helpers.php
+++ b/src/Helpers.php
@@ -871,9 +871,9 @@ class Helpers
                     $uri = Helpers::encodeURI($uri);
                 }
                 if (isset($maxlen)) {
-                    $result = file_get_contents($uri, null, $context, $offset, $maxlen);
+                    $result = file_get_contents($uri, false, $context, $offset, $maxlen);
                 } else {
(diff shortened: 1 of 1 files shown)


Mellthas (approved):
Thanks!

bsweeney:
Indeed, "passing `null` to non-nullable parameters of built-in functions" was deprecated with 8.1
https://www.php.net/manual/en/migration81.deprecated.php

I _think_ this was maybe done because we didn't have preference for that argument and just wanted to use the default. But we should have explicitly specified the default instead of passing null.

dejury:
Can someone please tell me what the ETA of this release is? It is pretty annoying that PHP8.1 still throws errors.

bsweeney:
Expecting to have this out next week.

ddkpdev:
Any updates on release date?

tiagofrancafernandes:
https://github.com/dompdf/dompdf/issues/2732#issuecomment-1126734141

PR #2668, opened 2021-12-01
Additional fixes for PHP 8.1 deprecations

Mellthas:
Following #2659. See the individual commits, which also include some small related cleanups.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,8 +35,8 @@
         "php": "^7.1 || ^8.0",
         "ext-dom": "*",
         "ext-mbstring": "*",
-        "phenx/php-font-lib": "^0.5.2",
-        "phenx/php-svg-lib": "^0.3.3"
+        "phenx/php-font-lib": "^0.5.4",
+        "phenx/php-svg-lib": "^0.3.3 || ^0.4.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^7.5 || ^8 || ^9",
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -5051,17 +5051,17 @@ EOT;
 
         $text = str_replace(["\r", "\n"], "", $text);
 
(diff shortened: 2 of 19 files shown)


bsweeney:
Added two more changes. I'm working on an update to php-font-lib to address issues there as well. We should update the php-font-lib version referenced once a new release is available.

php-svg-lib looks to be compatible at a base level, at least that's how it appears after some limited testing. However, the php-css-parser dependency does not yet have a release fully compatible with PHP 8.1.

bsweeney:
I think this is ready to merge.

Mellthas:
Since php-svg-lib requires PHP 7.4 now, won’t raising the php-svg-lib dependency implicitly raise the requirement for dompdf, too? So maybe the requirement should be `^0.3.3 || ^0.4.0`.

Mellthas:
I still get a deprecation warning in php-font-lib when using any font other than the standard PDF ones (e.g. DejaVu):

`Deprecated: mb_strlen(): Passing null to parameter #1 ($string) of type string is deprecated in .../dompdf/vendor/phenx/php-font-lib/src/FontLib/Glyph/Outline.php on line 99`

bsweeney:
> So maybe the requirement should be `^0.3.3 || ^0.4.0`

Of course you are correct. Re-pushed.

> I still get a deprecation warning in php-font-lib

How did I miss that? 😞

bsweeney:
That was _very_ sloppy of me. I had an uncommitted change in my local that I _totally_ missed.

bsweeney:
Well _that_ was embarrassing! Thanks for spotting my mistakes there and for helping get php-font-lib back in order, @Mellthas. How are we looking now?

Mellthas:
I think we are good now, haven’t found any further issues!

bsweeney (approved):

PR #2686, opened 2021-12-08
Small simplification for callback handling

Mellthas:
Code cleanup plus some very basic tests for callback handling.

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -1411,7 +1411,7 @@ class Dompdf
      * Sets callbacks for events like rendering of pages and elements.
      *
      * The callbacks array should contain arrays with `event` set to a callback
-     * event name and `f` set to a function or object plus method to be called.
+     * event name and `f` set to a function or any other callable.
      *
      * The available callback events are:
      * * `begin_page_reflow`: called before page reflow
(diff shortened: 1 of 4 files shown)


src/Dompdf.php
```diff
@@ -875,18 +868,14 @@ public function add_info($label, $value)
...
      * Writes the output buffer in the log file
      *
-     * @return void
+     * @param string $logOutputFile
+     * @param float $startTime
      */
-    private function write_log()
+    private function writeLog(string $logOutputFile, float $startTime): void
     {
-        $log_output_file = $this->getOptions()->getLogOutputFile();
```
bsweeney, line 882: Would it make more sense to move the log checking/creation logic into this method? It's more of a change but when we add PSR-3 support for logging the caller will just make the call and won't concern itself with the work of determining if the log exists.

src/FrameReflower/Page.php
```diff
@@ -174,31 +174,26 @@ function reflow(BlockFrameDecorator $block = null)
...
+
             foreach ($fs as $f) {
-                if (is_callable($f)) {
-                    if (is_array($f)) {
-                        $f[0]->{$f[1]}($info);
-                    } else {
-                        $f($info);
-                    }
-                }
+                $f($info);
```
bsweeney, line 196: Just an FYI. I believe the method was _supposed_ to support both a function and an array with an object/string combination representing a callable class method. I don't know if that was ever documented anywhere. And it looks like we pretty much never really supported it since one of the conditions for adding a callback was that it was callable 😕 

I think this change is fine since we never supported the other method and, really, who would use it anyway.

bsweeney (approved):

src/Dompdf.php
Mellthas: Would have to be two methods I think, because of the output buffering, which is only enabled if the log file is writable. Anyway, this was just a quick drive-by cleanup, since the method is private.

src/FrameReflower/Page.php
Mellthas: Hm, but `is_callable` and the `$callable()` syntax both support an array with object/string! They also support an array with class name/string to call static methods, which would previously fail because of the `is_array` check. See https://www.php.net/manual/en/language.types.callable.php for reference.

Support for the object/string combination was documented in the `setCallbacks` doc block (_… set to a function or object plus method to be called_).

bsweeney: ah ha ha shows what _I_ know, I obviously haven't been using this feature of PHP very much

PR #2688, opened 2021-12-08
Make dompdf setters available as options

### LINKED ISSUE #2682: Why `set_http_context` is not available from `Options()`
Why is not available `set_http_context` on [Dompdf\Options](https://github.com/dompdf/dompdf/blob/master/src/Options.php)?
https://github.com/dompdf/dompdf/blob/61c86c04d2a483187ff9f6a73c50d42669be5b4d/src/Dompdf.php#L1252-L1255

erikn69:
Closes #2682
Setters moved to  [Dompdf\Options](https://github.com/dompdf/dompdf/blob/master/src/Options.php), and prevent overwriting setter value on `setOptions` when new options brings empty values(BC)

```batch
/var/www/github/dompdf# ./vendor/bin/phpunit
PHPUnit 9.5.10 by Sebastian Bergmann and contributors.

...................................................               51 / 51 (100%)

Time: 00:00.305, Memory: 12.00 MB

OK ( 51 tests, 209 assertions)
```

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -77,27 +77,6 @@ class Dompdf
     private $version = 'dompdf';
 
     /**
-     * DomDocument representing the HTML document
-     *
-     * @var DOMDocument
-     */
-    private $dom;
-
-    /**
-     * FrameTree derived from the DOM tree
-     *
-     * @var FrameTree
-     */
-    private $tree;
-
-    /**
-     * Stylesheet for the document
-     *
-     * @var Stylesheet
-     */
-    private $css;
-
-    /**
      * Actual PDF renderer
      *
      * @var Canvas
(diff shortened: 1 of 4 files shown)


Mellthas:
Just a quick note: I think moving `$dom`, `$tree`, `$css`, and `$callbacks` to `Options` doesn’t make any sense. They are document-specific, while `Options` contains general options that you might want to share between different documents. Not sure about the others right now.

erikn69:
Maybe you are right, feel free to close the PR if it is not useful, i don't need this, i just trying to help with #2682

Mellthas:
Didn’t want to sound harsh, was just a bit surprised that you moved all the properties.

erikn69:
Don't worry, sounds normal to me, i'm still learning english, and using google translate, 
> i can revert everything and only move set_http_context to options, no problem for me

I can't do it, if it is better for the package

PR #2689, opened 2021-12-08
Add support for passing a callback to `page_script` method

### LINKED ISSUE #1838: page_script callback?
I'm trying to set up a slightly tricky header and footer (page numbers in each footer, no header on the first page, and user information on subsequent pages' headers). I see that I can use `page_script` to do different headers for different page numbers, but setting up PHP code as text feels awkward.

It would be nice of page_script supported invoking a callback (and passing the $pdf, $fontMetric, $PAGE_NUM, and $PAGE_COUNT parameters) as well as calling eval. If there's interest, I can put together a PR demonstrating this.

Mellthas:
Implementation notes:
* The `$type` parameter is unused, and since calling user-defined functions with extra parameters is not an error, I simply removed the parameter.
* Added commented `page_script` and `page_line` signatures to the `Canvas` interface, to be enabled with the next major release, as to not break existing custom implementations.
* Didn’t add the string option to the (commented) interface declaration, as it feels like a legacy feature. Didn’t realize this before, but the code string is also only evaluated with the `$isPhpEnabled` option set (though that would be easy to resolve).

Fixes #1838

### DIFF
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -1061,20 +1061,29 @@ class CPDF implements Canvas
     }
 
     /**
-     * Processes a script on every page
+     * Processes a callback or script on every page
      *
-     * The variables $pdf, $PAGE_NUM, and $PAGE_COUNT are available.
+     * The callback function receives the four parameters `$pageNumber`,
+     * `$pageCount`, `$pdf`, and `$fontMetrics`, in that order. If a script is
+     * passed as string, the variables `$PAGE_NUM`, `$PAGE_COUNT`, `$pdf`, and
+     * `$fontMetrics` are available instead.
      *
-     * This function can be used to add page numbers to all pages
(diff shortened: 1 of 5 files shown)


bsweeney:
Maybe for the 2.0 release we mark string values as deprecated and remove it with 3.0? I don't see much need to continue supporting it, though, since there are few situations where the user can't just pass in a callable.

Generated logic being the exception, but in those cases the user can wrap variables. Maybe add an issue for discussion in case anyone with a vested interest in the feature wants to provide feedback?

bsweeney (approved):

Mellthas:
Maybe we could even mark string values as deprecated with the next release and remove it in 2.0 already? I can’t imagine any situation right now where only passing a string would work, but passing a callable wouldn’t. I’ll open an issue for discussion.

PR #2705, opened 2021-12-22
Drop PHP support older than 7.4

PowerKiKi:
Because those are EOL, and this would solve the confusion expressed in  https://github.com/dompdf/php-svg-lib/pull/78#issuecomment-999455409

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,9 @@
 language: php
 
 php:
-  - 7.1
-  - 7.2
-  - 7.3
   - 7.4
   - 8.0
+  - 8.1
   - nightly
   - hhvm
 
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Follow us on [![Twitter](http://twitter-badges.s3.amazonaws.com/twitter-a.png)](
  
 ## Requirements
 
- * PHP version 7.1 or higher
+ * PHP version 7.4 or higher
  * DOM extension
  * MBString extension
  * php-font-lib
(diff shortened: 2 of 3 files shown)


Spreeuw:
Dropping support for PHP 7.1/7.2/7.3 makes sense if you need to implement changes that can't be done on those versions - you can then do that when you release those changes. Dropping it just because it's EOL is not a strong argument to me.

PowerKiKi:
You cannot modernize code as long as you declare supporting old versions. Dropping those versions now means that any contributions, even the smallest one, can start using modern syntax. Stronger typings comes to mind first, but there are a lot of nice things in recent PHP versions that will never be naturally adopted if we don't drop support for old PHP version first.

PowerKiKi:
Also I'd like to remind that [dropping PHP version is not a breaking change](https://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html), and it can be release in a minor version just fine. Doctrine and PHPSpreadsheet are projects, amongst others, that do that routinely without any issues.

So this PR should target milestone 1.2.0, not 2.0.0.

Mellthas:
I see no actual benefit in dropping support with 1.2.0. The only advantage would be clearing up the possible confusion about the `php-svg-lib` dependency, which should only be an actual problem when using different PHP versions between different environments as described in the post that @PowerKiKi linked. And as mentioned there (_Making sure you get a compatible version_), that situation can be handled with the `platform.config` setting.

If `php-svg-lib` keeps the PHP 7.4 requirement, we could put a note about the setting into the release notes, and keep it until the PHP requirements for the latest versions are compatible between the projects again.

PR #2722, opened 2022-01-05
Refactor handling of computed values

### LINKED ISSUE #612: Support the "initial" CSS3 keyword
_Original author: petrglushchenko, (March 19, 2013)_

**What steps will reproduce the problem?**
`#sidebar h2 { border-color: initial; }`
https://developer.mozilla.org/docs/CSS/initial

**What is the expected output? What do you see instead?**
Error 500: _border_initial is not valid

**What version of dompdf are you using? What version of PHP? On what operating system?**
0.6.0 beta 3

**Please provide the HTML source code you want to convert, or any additional information.**
To resolve issues I added this lines to dompdf/include/abstract_renderer.cls.php :

``` php
protected function _border_initial($x, $y, $length, $color, $widths, $side, $corner_style = "bevel") {
  return;
}
```

_Original issue: http://code.google.com/p/dompdf/issues/detail?id=562_

### LINKED ISSUE #1515: caption border is not rendered - upgrade from 0.7 to 0.8
In my HTML I use this:

``<caption style="border: 1px solid #BEBEBE;  line-height: 250%; vertical-align: middle;" class="grey">
``

after migrating from dompdf 0.7 to 0.8 the border is not rendered anymore.
here is my css file for table and caption:
```
table {
	border-collapse: collapse;
}

caption {
	color: inherit;
	text-align: center;
	vertical-align: middle;
}

```

Did I miss something? Is it an issue (did not find anything about it so far)?
Thanks

Mellthas:
* Improve performance due to more efficient handling of computed values and shorthand properties
* Fix incorrectly calculated `rem` lengths in some cases, due to the root style not being available at the time of computation
* Fix incorrectly calculated `em` and `ex` lengths in some cases, due to incorrect caching in `length_in_pt`
* Fix several bugs related to `!important` handling, e.g. important declarations were incorrectly considered during inheritance
* Start dropping invalid declarations. This can be improved upon by extending the existing property setters or implementing missing ones
* Move `display` value computation to the `Style` class, which makes borders and backgrounds work on `table-caption` and `flex` elements, as they are fully treated as `block` elements now
* Add support for the `initial` and `unset` keywords
* Add support for the `currentcolor` color keyword
* Support `background-size` in `background` shorthand

This changes the `Style` class to only compute values when actually needed, instead of immediately on set. Previously, values were computed immediately while parsing style blocks, which meant unnecessary computation of values from unused styles and unnecessary invalidations later because of property dependencies. It also makes sure that the the root style is available when computing lengths, as values are now never computed before style merge and inheritance, at which point the root style is defined already. This fixes some cases of incorrectly computed `rem` lengths. Additionally, the handling of shorthand properties and `!important` declarations is cleaned up.

Together, these changes result in quite a noticeable performance improvement in the range of about 15–35% of the total render time, judging from some rudimentary testing with different documents.

Fixes #612
Fixes #1515

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -140,14 +140,104 @@ class Style
         "double", "groove", "ridge", "inset", "outset"];
 
     /**
-     * List of CSS shorthand properties
+     * Map of CSS shorthand properties and their corresponding sub-properties.
+     * The order of the sub-properties is relevant for the fallback getter,
+     * which is used in case no specific getter method is defined.
      *
      * @var array
      */
-    protected static $_props_shorthand = ["background", "border",
-        "border_bottom", "border_color", "border_left", "border_radius",
-        "border_right", "border_style", "border_top", "border_width",
-        "flex", "font", "list_style", "margin", "padding", "outline"];
+    protected static $_props_shorthand = [
+        "background" => [
+            "background_image",
+            "background_position",
+            "background_size",
+            "background_repeat",
+            // "background_origin",
+            // "background_clip",
+            "background_attachment",
+            "background_color"
+        ],
+        "border" => [
+            "border_width",
+            "border_style",
+            "border_color"
+        ],
+        "border_top" => [
+            "border_top_width",
+            "border_top_style",
+            "border_top_color"
+        ],
+        "border_right" => [
+            "border_right_width",
+            "border_right_style",
+            "border_right_color"
+        ],
+        "border_bottom" => [
+            "border_bottom_width",
+            "border_bottom_style",
+            "border_bottom_color"
+        ],
+        "border_left" => [
+            "border_left_width",
+            "border_left_style",
+            "border_left_color"
+        ],
+        "border_width" => [
+            "border_top_width",
+            "border_right_width",
+            "border_bottom_width",
+            "border_left_width"
+        ],
+        "border_style" => [
+            "border_top_style",
+            "border_right_style",
+            "border_bottom_style",
+            "border_left_style"
+        ],
+        "border_color" => [
+            "border_top_color",
+            "border_right_color",
+            "border_bottom_color",
+            "border_left_color"
+        ],
+        "border_radius" => [
+            "border_top_left_radius",
+            "border_top_right_radius",
+            "border_bottom_right_radius",
+            "border_bottom_left_radius"
+        ],
+        "font" => [
+            "font_family",
+            "font_size",
+            // "font_stretch",
+            "font_style",
+            "font_variant",
+            "font_weight",
+            "line_height"
+        ],
+        "list_style" => [
+            "list_style_image",
+            "list_style_position",
+            "list_style_type"
+        ],
+        "margin" => [
+            "margin_top",
+            "margin_right",
+            "margin_bottom",
+            "margin_left"
+        ],
+        "padding" => [
+            "padding_top",
+            "padding_right",
+            "padding_bottom",
+            "padding_left"
+        ],
+        "outline" => [
+            "outline_width",
+            "outline_style",
+            "outline_color"
+        ]
+    ];
 
     /**
      * Maps legacy property names to actual property names.
@@ -359,7 +449,7 @@ class Style
             $d["border_right_width"] = "medium";
             $d["border_bottom_width"] = "medium";
             $d["border_left_width"] = "medium";
-            $d["border_width"] = "medium";
+            $d["border_width"] = "";
             $d["border_bottom_left_radius"] = "0";
             $d["border_bottom_right_radius"] = "0";
             $d["border_top_left_radius"] = "0";
@@ -749,8 +839,8 @@ class Style
         }
 
         foreach (self::$_inherited as $prop) {
-            // don't inherit shorthand properties, the specific properties will inherit
-            if (in_array($prop, self::$_props_shorthand, true)) {
+            // Inherit is handled via the specific sub-properties for shorthands
+            if (isset(self::$_props_shorthand[$prop])) {
                 continue;
             }
 
@@ -780,11 +870,6 @@ class Style
         }
 
         foreach ($this->_props as $prop => $value) {
-            // don't inherit shorthand properties, the specific properties will inherit
-            if (in_array($prop, self::$_props_shorthand, true)) {
-                continue;
-            }
-
             if ($value === "inherit") {
                 if (isset($parent->_important_props[$prop])) {
                     $this->_important_props[$prop] = true;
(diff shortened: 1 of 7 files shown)


Mellthas:
### Test case for the `rem` fix

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
	size: 400pt 500pt;
	margin: 50pt;
}

html {
	font-size: 40pt;
}

body {
	outline: thin solid black;
}

div {
	margin: 1rem;
	height: 40pt;
	font-size: 10pt;
	background-color: lightblue;
}

.box1 {
	font-size: 10pt;
}

.box2 {
	font-size: 12pt;
}

.box3 {
	font-size: 14pt;
}
</style>
</head>

<body>
	<div class="box1">margin: 1rem; font-size: 10pt</div>
	<div class="box2">margin: 1rem; font-size: 12pt</div>
	<div class="box3">margin: 1rem; font-size: 14pt</div>
</body>

</html>
```
</details>

[rem-before.pdf](https://github.com/dompdf/dompdf/files/7815016/rem-before.pdf)
[rem-after.pdf](https://github.com/dompdf/dompdf/files/7815015/rem-after.pdf)

### Test case for the `length_in_pt` cache fix

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
	size: 400pt 300pt;
	margin: 50pt;
}

html {
	font-size: 2.5rem;
}

div {
	margin-bottom: 10pt;
	height: 2em;
	background-color: lightblue;
}

.box1 {
	font-size: 10pt;
}

.box2 {
	font-size: 20pt;
}
</style>
</head>

<body>
	<div class="box1">font-size: 10pt</div>
	<div class="box2">font-size: 20pt</div>
	<div class="box3">font-size: 2.5rem</div>
</body>

</html>
```
</details>

[cache-before.pdf](https://github.com/dompdf/dompdf/files/7815013/cache-before.pdf)
[cache-after.pdf](https://github.com/dompdf/dompdf/files/7815014/cache-after.pdf)

Mellthas:
Some implementation notes: I tried not to break compatibility here, so that this does not require a major release. There are some small behavioral differences, like the `set_*` methods, which are used for computing values, not setting the specified value anymore. But, I would ideally go even further here, making all the property-specific setters and getters `protected`. The setters could then deal with computation only, which would allow further simplification; they could then just return the computed value and would not have to deal with setting the prop arrays. Having these method `public` seems confusing to me, as they are available only for a subset of the properties, and they make it less clear what the proper way to access style properties is (use `$style->$prop`, `$style->get_prop($prop)`, `$style->get_$prop()`?). Maybe the initial idea there was to have the `Style` class of more general use outside dompdf?

My idea for further refactoring then would be to have dedicated methods to get or set specified/computed/used values instead:
* `__get()` is already a used-value getter effectively. Make `__set()` operate on used values, too, as that it was is actually mostly needed during layout. This would improve performance further, as no computation is needed then.
* `get_prop` currently retrieves the computed or specified value. Split that into two methods, one for retrieving the specified, one for retrieving the computed value. Uses of `original_style` to read the specified value of a property could then simply be replaced by the getter for the specified value.
* Introduce parallel (specified/computed) setter methods. The specified-value setter would be the current `__set()`. This would allow replacing uses of `original_style` to set values during layout that should not be reset on frame reset.
* Following that, `original_style` could be removed. On frame reset, the style would be reset instead, clearing only used values. This should improve performance and memory usage, as only one style object would be needed per frame, and the cloning could be avoided on reset.

Mellthas:
Got a bit deeper than expected. Noticed that the initial changes regressed handling of invalid declarations, so I fixed that up. The result is that invalid declarations are ignored now (though invalid declarations can still mask valid ones within a single style block). This depends on the property setters setting the computed value to `null`, which most of them already did, although others cleared the specified value instead. Now the behavior is consistent, and so can be more easily improved upon by extending the existing property setters or implementing missing ones. To validate the new logic, I extended the invalid-declaration handling for margins, padding, and border widths, and implemented a setter for `display`.

Mellthas:
The idea for further refactoring outlined in the previous comment is now almost trivial to implement. With the `display` computation logic now inside the `Style` class, dedicated methods for computed values are not even needed for removing `original_style`. `__set()` would just utilize the logic of `set_used_value()` from #2723, just setting the used value, and that would be mostly it.

Mellthas:
### Test case for handling of invalid declarations

Not all cases are handled correctly yet. Also demonstrates some features that are not covered by the spec, like `none` as keyword for margin and padding and  support for unit-less lengths, which are treated as `pt` values. Just wanted to make sure that I didn’t remove these accidentally.

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
    size: 400pt 400pt;
    margin: 50pt;
}

html {
    display: inline-block;
}

body {
    outline: thin solid black;
}

div {
    padding: 5pt;
    border: 1pt solid red;
}

div + div {
    margin-top: 10pt;
}

.box1 {
    padding-left: 20pt;
    padding-left: -50pt;
}

b {
    padding: inherit;
}

.box2 {
    width: 150;
}

.box3 {
    width: fit-content;
    margin: 10;
}

.box4 {
    padding: none;
}

.box5 {
    border: 5% dashed;
}

.box6 {
    border-radius: asd;
}

.inline-flex {
    display: inline-flex;
    vertical-align: top;
    border: solid;
    background: lightblue;
}

.invalid {
    display: invalid;
    background: lightblue;
}
</style>
</head>

<body>
    <div class="box1">Negative <b>padding</b></div>
    <div class="box2">Unitless width</div>
    <div class="box3">Fit content</div>
    <div class="box4">padding: none</div>
    <div class="box5">Percentage border</div>
    <div class="box6">Invalid border radius</div>
    <div class="box7">display <span class="inline-flex">inline-flex</span> <span class="invalid">invalid</span></div>
</body>

</html>
```
</details>

[invalid-before.pdf](https://github.com/dompdf/dompdf/files/7831601/invalid-before.pdf)
[invalid-after.pdf](https://github.com/dompdf/dompdf/files/7831600/invalid-after.pdf)

bsweeney:
> Together, these changes result in quite a noticeable performance improvement in the range of about 15–35% of the total render time, judging from some rudimentary testing with different documents.

😮 That's a pretty big jump! This class has definitely been one source of performance issues so I'm glad to see it.

> I would ideally go even further here, making all the property-specific setters and getters protected. Maybe the initial idea there was to have the Style class of more general use outside dompdf?

I agree that these shouldn't be public. I don't think this class was ever meant to be used independent from Dompdf. I suspect, though would have to remind myself by reviewing the commit history, that some of the property-specific methods were called directly within Dompdf. When I was refactoring the logic to its current state I made sure that the getter/setter overload methods were used throughout. It's probably safe to plan on making those methods protected. I'd say go ahead and open an enhancement issue so we can talk through your suggestions for a future major release.

src/Css/Style.php
```diff
@@ -589,212 +757,183 @@ function get_stylesheet()
...
-                }
+            if (isset($parent->_props[$prop])) {
+                $parent_val = \array_key_exists($prop, $parent->_props_computed)
+                    ? $parent->_props_computed[$prop]
+                    : $parent->compute_prop($prop, $parent->_props[$prop]);
+
+                $this->_props[$prop] = $parent_val;
+                $this->_props_computed[$prop] = $parent_val;
+                $this->_prop_cache[$prop] = null;
             }
```
bsweeney: TODO: Does this break for properties that are dependent on other properties?

bsweeney (approved):

Mellthas:
Thanks for the quick review! Noticed a small oversight in the `merge` method, which is fixed now (broke custom margins on non-base page styles).

> I'd say go ahead and open an enhancement issue so we can talk through your suggestions for a future major release.

I’ll do that once the picture in my mind is a bit clearer.

Issue #449, opened 2012-12-06
Support anonymous table objects


PhenX:
_Original author: eclecticgeek (March 17, 2012 02:07:35)_

When a table is missing a structural element as outlined in CSS 2.1 it can cause problems during rendering. This is especially true of the element is not rendered normally because its display has been set to something other than the default for that element. See issue 438.

Full support for anonymous table objects would resolve the cited issue.

Ref: http://www.w3.org/TR/CSS2/tables.html#anonymous-boxes

_Original issue: http://code.google.com/p/dompdf/issues/detail?id=443_

Mellthas:
This should be mostly covered by #2733, apart from generating missing table parents for table-internal elements out in the wild.

PR #2733, opened 2022-01-12
Normalize tables by wrapping invalid children instead of moving them

### LINKED ISSUE #1363: Stylesheet::_styles one or two dimensional array ?
`Stylesheet::Stylesheet::apply_styles` and [Stylesheet::add_style](https://github.com/dompdf/dompdf/blob/master/src/Css/Stylesheet.php#L268-L280) are using `_styles` as a two dimensional array. [Stylesheet::lookup](https://github.com/dompdf/dompdf/blob/master/src/Css/Stylesheet.php#L291-L298) instead as a one dimensional.

I am pretty sure I miss something or the callers to lookup won't be called itself. Stopped on investigating at some point cause it would be a long dead route.

### LINKED ISSUE #2529: table:before pseudo-elements rendered at the bottom of the table
`:before` pseudo-elements appear to be rendered such that the element top is at the bottom of the table. In addition to the visually incorrect placement of the pseudo-element, and layout adjustments (e.g. float clearing) is applied after the table instead of before.

```
<style>
  .float { float: left; height: 100px; width: 10px; background-color: green; }
  table { border: 1px solid red; }
  .collapse { border-collapse: collapse; }
  table:before { display: block; height: 10px; width: 10px; background-color: blue; margin: 10px; }
  table.clear:before { display: block; height: 10px; width: 10px; background-color: blue; clear: both; }
</style>

<table>
  <tr>
    <td>dompdf</td>
  </tr>
</table>

<table class="collapse">
  <tr>
    <td>dompdf</td>
  </tr>
</table>

<div class="float"></div>
<table class="clear">
  <tr>
    <td>dompdf</td>
  </tr>
</table>
```

Mellthas:
Implements the rules according to the spec, but makes sure that table rows are always wrapped in row groups. This is conceptually much cleaner than the previous logic and provides a better fallback for invalid table structure.

Generating missing `table`/`inline-table` parents is not covered.

https://www.w3.org/TR/CSS21/tables.html#anonymous-boxes
https://www.w3.org/TR/CSS21/visuren.html#anonymous

Partially addresses #449
Fixes #1363
Fixes #2529


bsweeney (approved):

bsweeney:
FYI 2529 may have been addressed by some of your earlier commits. It is significantly improved on master already.

src/FrameDecorator/Table.php
```diff
@@ -232,143 +232,190 @@ public function set_max_width($width)
...
-                    $table_row_group = Factory::decorate_frame($frame, $this->_dompdf, $this->_root);
+            if (in_array($display, self::$ROW_GROUPS, true)) {
+                $this->normalize_group($child);
+            }
+        }
+    }
 
-                    // Create an anonymous table row
-                    $tr = $this->get_node()->ownerDocument->createElement("tr");
+    private function normalize_group(AbstractFrameDecorator $frame): void
```
bsweeney, line 340: Should the child normalization methods go into the relevant FrameDecorator class (TableRowGroup, TableRow)?

PR #2752, opened 2022-01-19
Remove superfluous iterator indirection

Mellthas:
Instead of returning an object which can only be iterated and nothing else, return the iterator directly. `FrameList` was likely modeled after the DOM `NodeList`, but the latter does provide additional capabilities like direct access to its length, or the ability to retrieve children by position.

<del>For `FrameTree`, it might be even more natural to make it iterable directly by implementing `IteratorAggregate`, instead of needing to call `get_frames()`, but I left that out for now.</del>


Mellthas:
@bsweeney Can you have a look at this for 2.0? Essentially just a minor simplification and optimization.

Issue #1946, opened 2019-06-10
Extra margins when using inline-block and margin-top


PowerKiKi:
The combination of `display: inline-block;` and `margin-top` will mess up the margin of the containing element and each contained element after the first one. That makes it very difficult to have a proper alignment and margin.

In my real use-case there is no container, so I cannot apply a padding on the container to workaround this issue. Also `vertical-align: top;` on any elements does not allow to workaround the issue either.

# Reproducible example 

See http://eclecticgeek.com/dompdf/debug.php?identifier=69da216620f4370b949628b93ae3f072 or the code below:

```php
<?php

use Dompdf\Dompdf;

require_once '../vendor/autoload.php';

$html = <<<STRING
<!doctype html>
<html lang="en">
    <head>
        <title>My title</title>
        <style>
            .container {
                background: pink;
            }
            
            .foo {
                background: gold;
                width: 30%;
            }
            
            .bar {
                background: lightgreen;
                width: 70%;
            }
            
            .foo, .bar {
                display: inline-block;
                height: 50px;
                margin-top: 200px;
            }
        </style>
    </head>
    <body>
        <p>lorem lipsum</p>
        <div class="container">
            <div class="foo">foo foo foo</div><div class="bar">bar bar bar</div>
        </div>
    </body>
</html>
STRING;

$dompdf = new Dompdf();
$dompdf->loadHtml($html);
$dompdf->setPaper('A4');
$dompdf->render();

// Comment out to see HTML
$dompdf->stream('test.pdf', ['Attachment' => false]);

echo $html;
```

# Expected

HTML as rendered by Chrome:

![image](https://user-images.githubusercontent.com/72603/59166727-c8a76880-8b80-11e9-8853-d50b5e9566ad.png)

# Actual

PDF as rendered by dompdf/dompdf v0.8.3 will show extra margins in different places:

![image](https://user-images.githubusercontent.com/72603/59166958-cfcf7600-8b82-11e9-9210-fd674034505b.png)

# Versions

- dompdf/dompdf v0.8.3, also failing on develop branch 5c20bb461558b2a37f9f9ff6f902337e6ca4f8c2
- PHP 7.2.19, also failing on PHP 7.3 according to eclecticgeek.com/dompdf

PowerKiKi:
Re-reading the following issues I realize they are duplicates, or related:

[...]

PR #2757, opened 2022-01-21
Revert "Use margin height instead of border height in vertical alignment"

### LINKED ISSUE #2756: Margin ignored on ul element
This example shows the top margin of an `<ul>` element being ignored:

HTML: <details>

```html
<style type="text/css">
	ul {
		display: inline-block;
		list-style: none;
		padding: 0;
		margin: 0;
		margin-top: 10px;
	}

	ul li {
		display: inline-block;
		padding: 0;
		margin: 0;
	}

	ul li strong {
		display: inline-block;
		font-weight: normal;
	}

	ul li p {
		display: inline-block;
		margin: 0;
	}
</style>


<div>
	<span>Biozertifiziert</span> | <span>DE-&Ouml;KO-XXX</span> |
	<ul>
		<li><strong>Variante:</strong> <p>1kg (20% Rabatt)</p></li>
	</ul>
</div>
```
</details>

dompdf-margin-top-dev-master

In 1.0.2 it looked like this:
dompdf-margin-top-1 0 2

Neither situation seems to be rendered the same as Chrome renders the HTML, which simply moves down the entire line when the margin of that `<ul>` element is increased (rather than moving just the `<ul>` element itself. So it seems like the top margin was patching a bug to begin with, but now that it's ignored it appears to be even worse than before because it can no longer be controlled via that margin.

Mellthas:
This reverts commit 96a93459e06df326afee96a579f50d2594406237. It seems this was not an innocent fix as it seemed to be. Retain the existing behavior, which unfortunately undoes the small improvement to #1946.

Fixes #2756

### DIFF
--- a/src/FrameReflower/Block.php
+++ b/src/FrameReflower/Block.php
@@ -599,8 +599,8 @@ class Block extends AbstractFrameReflower
                         continue;
                     }
 
-                    $marginHeight = $frame->get_margin_height();
-                    $imageHeightDiff = $height * 0.8 - $marginHeight;
+                    $frameBox = $frame->get_border_box();
+                    $imageHeightDiff = $height * 0.8 - $frameBox["h"];
 
                     $align = $frame->get_style()->vertical_align;
                     if (in_array($align, Style::$vertical_align_keywords, true)) {
@@ -635,7 +635,7 @@ class Block extends AbstractFrameReflower
                                 break;
                         }
                     } else {
-                        $y_offset = $baseline - (float)$style->length_in_pt($align, $style->font_size) - $marginHeight;
+                        $y_offset = $baseline - (float)$style->length_in_pt($align, $style->font_size) - $frameBox["h"];
                     }
                 } else {
                     $parent = $frame->get_parent();


Spreeuw (approved):
Fixes #2756 for me indeed! 🚀

bsweeney (changes requested):
I don't think reverting this change is necessarily the way to go. I've noticed that alignment calculation is universally ... not great? I think the original logic just attempts to approximate the correct alignment without actually doing as the spec requires. Let's leave this open for now but I'm working on a change that I think will do a better job.

Mellthas:
Yeah, this would be the safe-route fix here, just not meddling with the logic for now.

But having looked at it again now, the top margin in #2756 seems to be a workaround for the misalignment really. The same can be achieved with `vertical-align: -10px` on the `ul` there, which works both with current master and with older versions. An advantage of the original change is that the difference in the baseline positions is now relatively stable no matter the margin, where it would differ greatly depending on the provided margin before. So the baselines are not correctly, but at least roughly, aligned. And any remaining misalignment can be worked around by using `vertical-align` with a specific length.

Let’s see what you have got :-); independently, it may be better to drop this PR, and consider the original change a small improvement instead.

bsweeney:
I agree, this does provide significant improvement generally despite the issue cited in 2756. Since there is a work-around I think we can drop this one. What I'm working on is a significant change to the logic so some review will definitely be in order. So far I'm pretty happy with the results.

Mellthas:
Alright then, let’s close this.

Issue #2755, opened 2022-01-21
Incorrect inline-block flow


Spreeuw:
It looks like there's an error in the width calculations of `inline-block` elements:

HTML: <details>

```html
<style type="text/css">
	ul {
		display: inline-block;
		list-style: none;
		padding: 0;
		margin-bottom: 0;
	}

	ul li {
		display: inline-block;
		padding-left: 3px;
		padding-right: 3px;
		margin-bottom: 0;
	}


	ul li strong {
		display: inline-block;
		font-weight: normal;
	}

	ul li p {
		margin: 0;
		padding: 0;
		display: inline-block;
	}
</style>
<ul><li><strong>F&uuml;llmenge:</strong> <p>500g</p></li></ul>
```
</details>

![fullmenge-dev-master](https://user-images.githubusercontent.com/3217773/150517361-d1609009-4ad9-4382-91ca-80c8db2f03f1.png)

In 1.0.2 it looked like this:
![fullmenge-1 0 2](https://user-images.githubusercontent.com/3217773/150517405-0b1d23f8-1a1d-4c48-8f2d-4ecf6ac09a8c.png)

Mellthas:
I need to set the font to DejaVu Sans, cannot reproduce with the default font.

Mellthas:
I think this is the same issue as #2560, and the problem is that float values are compared naively, not accounting for inaccuracies with float computation. The shrink-to-fit width is calculated first, then the inline frames are laid out, checking the available space left in each step; so the numbers being added up are not exactly the same, which can result in the values in the end being marginally off. For inline blocks, this was not an issue prior to version 1.1.0, since they didn’t use proper shrink-to-fit width. Only column widths in tables were affected.

I had a look at this a while ago, but didn’t pursue it further. Maybe we can resolve it now, will have a look in the coming days.

Mellthas:
Thanks for reporting these issue, by the way!

Spreeuw:
> I need to set the font to DejaVu Sans, cannot reproduce with the default font.

Sorry about not doing the example/testing with the default serif font. I think you are right about this being dependent on rounding in specific cases. After playing with the font size a bit I was able to reproduce with the default serif font too, at `21pt`, adding this to the example above:
```css
	body {
		font-family: serif;
		font-size: 21pt;
	}
```

Mellthas:
Yes, that also works for me. Small differences in numbers can make the issue appear and disappear.

[...]

PR #2761, opened 2022-01-23
Account for rounding errors when comparing float lengths

### LINKED ISSUE #2560: Table cell contents sometimes don't fit the calculated width
See the following HTML. Both table cells (the black bordered boxes) should be single-line, but the second cell wraps. Slight changes to the text or the table styling can make the difference between wrapping or not for each cell (e.g. setting `border-collapse: separate;` makes the first cell wrap, but not the second):

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
    size: 400pt 300pt;
    margin: 50pt;
}

table {
    border-collapse: collapse;
}

td {
    border: 1pt solid black;
    padding: 0.05pt;
}

span {
    display: inline-block;
    border: 1pt solid red;
}
</style>
</head>

<body>
<table>
    <tr>
        <td>Some content</td>
        <td>Some content Some content</td>
    </tr>
</table>
<p>
    <span>Some content</span>
    <span>Some content Some content</span>
</p>
</table>
</body>

</html>
```

Mellthas:
Adapted from [1] and [2]. More details on the underlying issue can also be found there.

We neither need to deal with very large, nor with very small numbers in layout. So using an arbitrary epsilon should be perfectly fine (`PHP_FLOAT_EPSILON` is only available starting from PHP 7.2); it is selected so that it results in a precision of at least:

* 7 decimal digits at around 1
* 4 decimal digits at around 1000 (around the size of common paper formats)
* 2 decimal digits at around 100,000 (100,000pt ~ 35.28m)

This should be more than enough for any reasonable number used. If anything, the epsilon value can easily be tuned later; just note that the tests would need change, as many of them demonstrate the effect of the (approximate) epsilon.

I started using the new comparison functions in critical places I found, i.e. where a failure to detect equality could result in a
meaningful difference in rendering, particularly the checks for fitting text and inline-block elements onto the current line, as the available width there may be the result of a shrink-to-fit width calculation.

Open question: Maybe it would be more appropriate to treat all negative values as unequal to zero, regardless of the difference, so that we don't accidentally end up with (very small) negative width or height values in some places.

[1] https://floating-point-gui.de/errors/comparison/
[2] https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/

Fixes #2560

### DIFF
--- a/src/FrameDecorator/Page.php
+++ b/src/FrameDecorator/Page.php
@@ -252,7 +252,8 @@ class Page extends AbstractFrameDecorator
             $style->padding_top
         ], $cbw);
 
-        return $childPos > $contentEdge && $contentEdge <= $this->bottom_page_edge;
+        return Helpers::lengthGreater($childPos, $contentEdge)
+            && Helpers::lengthLessOrEqual($contentEdge, $this->bottom_page_edge);
     }
 
     /**
@@ -574,7 +575,7 @@ class Page extends AbstractFrameDecorator
         }
 
         // Check if $frame flows off the page
-        if ($max_y <= $this->bottom_page_edge) {
+        if (Helpers::lengthLessOrEqual($max_y, $this->bottom_page_edge)) {
             // no: do nothing
             return false;
         }
--- a/src/FrameReflower/Text.php
+++ b/src/FrameReflower/Text.php
@@ -158,7 +158,7 @@ class Text extends AbstractFrameReflower
         ], $line_width);
         $frame_width = $text_width + $mbp_width;
 
-        if ($frame_width <= $available_width) {
+        if (Helpers::lengthLessOrEqual($frame_width, $available_width)) {
             return false;
         }
 
@@ -181,8 +181,9 @@ class Text extends AbstractFrameReflower
             $sep = $words[$i + 1] ?? "";
             $word = $sep === " " ? $words[$i] : $words[$i] . $sep;
             $word_width = $fontMetrics->getTextWidth($word, $font, $size, $word_spacing, $char_spacing);
+            $used_width = $width + $word_width + $mbp_width;
 
-            if ($width + $word_width + $mbp_width > $available_width) {
+            if (Helpers::lengthGreater($used_width, $available_width)) {
                 // If the previous split happened by soft hyphen, we have to
                 // append its width again because the last hyphen of a line
                 // won't be removed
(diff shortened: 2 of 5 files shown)


Mellthas:
Here is a simplified test case:

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
    size: 400pt 300pt;
    margin: 50pt;
}

table {
    border-collapse: collapse;
}

td {
    padding: 0;
    border: 1pt solid black;
}

span {
    display: inline-block;
    border: 1pt solid red;
}
</style>
</head>

<body>

<table>
    <tr>
        <td><b>Some content</b> Some content</td>
    </tr>
</table>

<p>
    <span><b>Some content</b> Some content</span>
</p>

</body>

</html>
```
</details>

[shrink-to-fit-before.pdf](https://github.com/dompdf/dompdf/files/7921418/shrink-to-fit-before.pdf)
[shrink-to-fit-after.pdf](https://github.com/dompdf/dompdf/files/7921417/shrink-to-fit-after.pdf)

So what is happening here? The width of the two text fragments within the cell (the inline-block case works out the same) is calculated as 68.988 for the bold text and 68.664 for the normal one, the sum of which is `137.652`. This is the calculated shrink-to-fit width. Then, when the text frames are actually laid out, `68.988` is first checked against the available width of `137.652`, and then `68.664` is checked against the remaining width of `137.652 - 68.988`; which should be `68.664` in theory, but because of rounding comes out as very slightly less (the difference is so small that `var_dump` doesn’t display it):

```php
var_dump(68.988 + 68.664);             // 137.652
var_dump(68.988 + 68.664 === 137.652); // true
var_dump(137.652 - 68.988);            // 68.664
var_dump(68.664 <= 137.652 - 68.988);  // false
```

bsweeney:
Heh, I actually have floating-point-gui.de open to read related to a related issue I was dealing with in some client-side code. Fixed the code in that case, still have to finish my research.

bsweeney (approved):

bsweeney:
It's too bad PHP doesn't have built-in precision number functionality. [BC Math functions](https://www.php.net/manual/en/book.bc.php) would have been an option if it were default functionality.

Spreeuw:
I can confirm this fixes #2755 too. Thanks!

Mellthas:
> It's too bad PHP doesn't have built-in precision number functionality. [BC Math functions](https://www.php.net/manual/en/book.bc.php) would have been an option if it were default functionality.

Although arbitrary precision would probably be overkill, considering the potential performance impact I would expect.

PR #2772, opened 2022-02-10
Fixed frame inline reflower when the frame has no child.

Daniel-KM:
There is a bug when a frame has no child for inline reflower.

### DIFF
--- a/src/FrameReflower/Inline.php
+++ b/src/FrameReflower/Inline.php
@@ -160,7 +160,8 @@ class Inline extends AbstractFrameReflower
         }
 
         // Assume the position of the first child
-        [$x, $y] = $frame->get_first_child()->get_position();
+        $firstChild = $frame->get_first_child();
+        [$x, $y] = $firstChild ? $firstChild->get_position() : [0, 0];
         $frame->set_position($x, $y);
 
         // Handle relative positioning


bsweeney:
I'm unable to trigger an error, despite the obvious potential null reference. Do you have a code sample?

markkimsal:
Empty get_first_child() guard is already checked around line 131.  This must've been triggered by a bad get_position() return value?

Mellthas:
> Empty get_first_child() guard is already checked around line 131. This must've been triggered by a bad get_position() return value?

With generated content involved, it is possible for the children to be removed during the reflow loop (lines 157–160), so that there are none anymore afterwards. I think I would like to solve the problem slightly differently, by returning early in such a case. I’ll need to take a closer look and post a sample showing the issue when I find some more time.

ilohpro:
I also have the same problem and i can confirm @markkimsal that during the check of the line 131, the first child isn't already null.
The change occurs during the line 163 operation but i couldn't find more precisly why on my side for the moment.

Mellthas:
Finally got back to this. See #2822 for a more thorough fix. I’ll post test cases there.

Mellthas:
Closing, as the issue should be addressed now. Please report back in case you can still trigger the error somehow with current master.

Issue #2608, opened 2021-10-18
Use composer to create packaged release ?


PowerKiKi:
Today DomPdf can either use composer's autoload mechanism, or use its custom autoload mechanism. I would like to suggest to drop the custom autoload mechanism. First of all because nowadays composer is quite ubiquitous.

But more importantly, because of the fact that DomPdf has this custom mechanic, it makes it mandatory for all (transitives) dependencies to also have such thing. This is visible in packaged version of DomPdf's `autoload.inc.php`:

```php
// Sabberworm
spl_autoload_register(function($class)
{
    if (strpos($class, 'Sabberworm') !== false) {
        $file = str_replace('\\', DIRECTORY_SEPARATOR, $class);
        $file = realpath(__DIR__ . '/lib/php-css-parser/lib/' . (empty($file) ? '' : DIRECTORY_SEPARATOR) . $file . '.php');
        if (file_exists($file)) {
            require_once $file;
            return true;
        }
    }
    return false;
});

// php-font-lib
require_once __DIR__ . '/lib/php-font-lib/src/FontLib/Autoloader.php';

//php-svg-lib
require_once __DIR__ . '/lib/php-svg-lib/src/autoload.php';
```

Removing entirely `src/autoload.php` and similar in dependencies would simplify maintenance (in at least 3 packages) and allow dependencies to manage their own dependencies independently of DomPdf.

As for the packaged version, if is deemed too important to drop it, then it can be re-created with automated third party tools, such as https://php-download.com/package/dompdf/dompdf. Even though my personal preference would be to not recommend pre-packaged version, because composer really is superior to "download and unzip".

I'd create a PR for that. What do you think ?

bsweeney:
Though not nearly as high as Composer installation, the packaged version still get quite a few downloads (https://api.github.com/repos/dompdf/dompdf/releases). So I think we'd probably do well to keep it around.

The autoload.inc.php seems irrelevant to the discussion, since dependency management wouldn't (typically) use the packaged release. So I think we can just focus on how Dompdf works with Composer and for source copied scenarios. It is true that the autoloader is excessive for when Composer is involved. Curious to see what changes you would make.

And hear any arguments against such a change.

PowerKiKi:
[...]

PR #2779, opened 2022-02-14
Bash script to compile release

markkimsal:
run from root of project to create releases/dompdf-X.X.X.tar.gz and zip files.

autoload.inc.php has been replaced with relative composer require.

satisfies #2608

### DIFF
--- a/scripts/autoload.inc.php
+++ b/scripts/autoload.inc.php
@@ -1,0 +1,2 @@
+<?php
+require (__DIR__ . "/vendor/autoload.php");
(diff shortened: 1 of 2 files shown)


PowerKiKi:
Before merging this, I would suggest to consider the alternative suggested in #2799. The difference is that the alternative will pull the release from packagist itself, instead of attempting to re-create it manually.

That ensure that the packaged release is as close as possible to what you get when you use composer normally. Minimizing differences across releases should minimize risks of discrepancies, and thus reduce maintenance.

PR #2783, opened 2022-02-17
Added support for jpeg data uri in $canvas->image()

davidsev:
Other image formats already work with data uri's, but jpeg's don't as cpdf->addJpegFromFile uses file_exists.
This change adds support for data uri's to addJpegFromFile.

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -6230,15 +6230,21 @@ EOT;
         // attempt to add a jpeg image straight from a file, using no GD commands
         // note that this function is unable to operate on a remote file.
 
-        if (!file_exists($img)) {
-            return;
+        if(substr($img, 0, 5) == 'data:') {
+            $filename = 'data-'.hash('md4',$img);
+        }
+        else {
+            if (!file_exists($img)) {
+                return;
+            }
+            $filename = $img;
         }
 
-        if ($this->image_iscached($img)) {
+        if ($this->image_iscached($filename)) {
             $data = null;
-            $imageWidth = $this->imagelist[$img]['w'];
-            $imageHeight = $this->imagelist[$img]['h'];
-            $channels = $this->imagelist[$img]['c'];
+            $imageWidth = $this->imagelist[$filename]['w'];
+            $imageHeight = $this->imagelist[$filename]['h'];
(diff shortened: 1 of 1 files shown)


davidsev:
That works for images in HTML, but it doesn't work for embedded php or callbacks.

For example, the following code doesn't work, but works fine with a png or bmp:
```
$dompdf->setCallbacks([[
    'event' => 'end_page_render',
	'f' => function ($info) {
		$canvas = $info['canvas'];
		$canvas->image('data:image/jpeg;base64,'.base64_encode(file_get_contents('tests/_files/jamaica.jpg')),20,80,100,100);
		$canvas->image('tests/_files/jamaica.jpg',20,200,100,100);
    }
]]);
```

The difference in behaviour between image formats suggests that this isn't intentional behaviour, and the wiki also implies that this should work.


Maybe a better solution would be to update the CPDF adapter to use the same image caching as the HTML uses?

bsweeney:
Thanks for the additional info. I'll take another look at the change.

bsweeney:
> the following code doesn't work, but works fine with a png or bmp.

You mean using a data-URI for a PNG or BMP adds the image to the PDF? That is entirely unintentional. I, honestly, had no idea that `file_get_contents` could read data-URIs! Which is completely silly since it is listed as a supported [wrapper](https://www.php.net/manual/en/wrappers.data.php) 😞.

> Maybe a better solution would be to update the CPDF adapter to use the same image caching as the HTML uses?

Not sure I would go quite that far. The Cpdf backend will be extracted from Dompdf into a separate project at some point.

lib/Cpdf.php
```diff
@@ -6230,15 +6230,21 @@ function addJpegFromFile($img, $x, $y, $w = 0, $h = 0)
         // attempt to add a jpeg image straight from a file, using no GD commands
         // note that this function is unable to operate on a remote file.
 
-        if (!file_exists($img)) {
-            return;
+        if(substr($img, 0, 5) == 'data:') {
+            $filename = 'data-'.hash('md4',$img);
+        }
```
bsweeney, line 6235: Now wondering if we should apply this same logic elsewhere where the filename is used as a caching key.

bsweeney (approved):

PR #2784, opened 2022-02-17
Features/font metrics cache as json

markkimsal:
Support both PHP and JSON formats for `dompdf_font_family_cache` file.


markkimsal:
I don't know why there are 129 commits.  Will try to re-create another PR.

bsweeney:
FYI this feature has been submitted in #2450

markkimsal:
Yes, but in that version they removed the PHP file instead of using it as a fallback.  Not sure how any of this would work on a running system, so I thought it would be better to support both formats at the same time for a while.

markkimsal:
can we rename the develop branch to something like `legacy/develop`?  because there's a comment on an open PR about using develop, then another discussion about getting rid of develop.

bsweeney:
I was planning to just remove develop at some point since we no longer use it. There's no pending work there and the history is fully represented in master.

As for 2450, it gets rid of the distributed font family and metrics cache files for the bundled fonts, but the generated versions are still read if present. If there's anything you think it missing go ahead and comment on that PR since it'll be included in an upcoming release.

markkimsal:
I looked at 2450, it's good.  I was confused by the master/develop situation and thought that PR was obsolete.  Anyway, I did this before I saw that one and thought, well why not PR push this up and get some discussion? If it gets closed it gets closed.

bsweeney:
I do appreciate the active interest in the project so keep it up 💯

PR #2799, opened 2022-03-11
Create packaged release with composer

### LINKED ISSUE #2608: Use composer to create packaged release ?

PowerKiKi:
This entirely replace our custom autoload mechanism. The added benefit
is reduced maintenance for DomPDF itself but also its dependencies. And
it opens up the possibility to add new dependency transparently, without
the need to manage their autoloading.

Fixes #2608
Closes #2779
Fixes https://github.com/dompdf/utils/issues/12
Closes https://github.com/dompdf/utils/pull/13


PowerKiKi:
`zipinfo -1 build/dompdf-1.2.0.zip` would give something similar to the following. The critical part being that `dompdf/autoload.inc.php` is 100% compatible across packages.


# Before

```
dompdf/CONTRIBUTING.md
dompdf/LICENSE.LGPL
dompdf/README.md
dompdf/SECURITY.md
dompdf/VERSION
dompdf/autoload.inc.php
dompdf/composer.json
dompdf/lib/*  all files redacted for brevity
dompdf/phpcs.xml
dompdf/src/*  all files redacted for brevity
```

# After

```
dompdf/CONTRIBUTING.md
dompdf/LICENSE.LGPL
dompdf/README.md
dompdf/SECURITY.md
dompdf/VERSION
dompdf/autoload.inc.php
dompdf/vendor/*  all files redacted for brevity
```

bsweeney (commented):
Do you have the time/inclination to make this a bit more OS-agnostic? `rm` and `cp` have equivalent PHP functions, but `which` does not. Maybe first check for a `COMPOSER` environment variable for identifying the location?

bsweeney:
I do currently use windows as my primary development environment. But I understand not wanting to build out complexity. I do have access to linux and mac so I can run the script there.

Mellthas:
Using something like Git Bash or WSL on Windows might also be an option.

PR #2800, opened 2022-03-13
Fix generated inline frames involving line breaks

### LINKED ISSUE #2794: Memory leak with link
I'm not sure what happens and why, but I get a memory leak when I use this HTML : 

```
<html>
        <head>
            <style>
                
                body {
                    font-family: Helvetica;
                }
                
                @page { 
                    size: A4;
                    margin: 1.5cm 1.5cm;
                } 
                a[href^=http]:after {
                    content:" <" attr(href) "> ";
                }
            </style>
        </head>
        <body> 
    
        <a href="https://solidarites-sante.gouv.fr/IMG/pdf/guide_auto-test.pdf">https://solidarites-sante.gouv.fr/IMG/pdf/guide_auto-test.pdf</a>
    
        </body>
    </html>
``` 

if you remove 
```
                a[href^=http]:after {
                    content:" <" attr(href) "> ";
                }
```
or if you  use 
```
      <a href="https://solidarites-sante.gouv.fr/IMG/pdf/guide_auto-test.pdf">Test</a>
```

it works.

You'll say it is stupid to add the href after the link if we already use it for the anchor text, and you'll be right. But in my case that's a user that made it like this... So it should work anyway.

This is not a big issue though, but I just wanted to let you know about it.

Mellthas:
Ensure that the changes to the split style are not immediately reset when the parent frame is split.

Fixes #2794

### DIFF
--- a/src/FrameDecorator/AbstractFrameDecorator.php
+++ b/src/FrameDecorator/AbstractFrameDecorator.php
@@ -246,8 +246,12 @@ abstract class AbstractFrameDecorator extends Frame
         if ($this->content_set
             && $this->get_node()->nodeName === "dompdf_generated"
         ) {
-            foreach ($this->get_children() as $child) {
-                $this->remove_child($child);
+            $content = $this->get_style()->content;
+
+            if ($content !== "normal" && $content !== "none") {
+                foreach ($this->get_children() as $child) {
+                    $this->remove_child($child);
+                }
             }
         }
     }
--- a/src/FrameDecorator/Inline.php
+++ b/src/FrameDecorator/Inline.php
@@ -89,10 +89,10 @@ class Inline extends AbstractFrameDecorator
         }
 
         //On continuation of inline element on next line,
-        //don't repeat non-vertically repeatable background images
+        //don't repeat non-horizontally repeatable background images
         //See e.g. in testcase image_variants, long descriptions
         if (($url = $split->get_style()->background_image) && $url !== "none"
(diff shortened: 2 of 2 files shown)


Mellthas:
The second commit fixes the logic for handling of non-repeated background images while at it. Here is a test case demonstrating the difference:

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>
<head>
<meta charset="UTF-8">
<style>
@page {
    size: 400pt 300pt;
    margin: 50pt;
}

.bg {
    background: url(http://placekitten.com/100/75) repeat-y;
    border: thin solid red;
}

.rx {
    background-repeat: repeat-x;
}
</style>
</head>
<body>

<span class="bg">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</span>
<br>
<span class="bg rx">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</span>

</body>
</html>
```
</details>

[repeat-before.pdf](https://github.com/dompdf/dompdf/files/8240145/repeat-before.pdf)
[repeat-after.pdf](https://github.com/dompdf/dompdf/files/8240144/repeat-after.pdf)

bsweeney (approved):

Mellthas:
Merged with a small fix on top.

PR #2807, opened 2022-03-16
Make `httpContext` available from options

### LINKED ISSUE #2682: Why `set_http_context` is not available from `Options()`

erikn69:
- Move `httpContext` from `Dompdf.php` to `Options.php` with the same functionality
- Support `array` on `setHttpContext()`(useful in laravel when using cache and arrays need to be serialized)

Now is posible to do:
```php
$options = new Options(['http_context' =>['ssl' => ['verify_peer'=>false]]]);
// or
$options = new Options();
$option->setHttpContext(['ssl' => ['verify_peer'=>false]]);
// or
$dompdf = new Dompdf();
$dompdf->setHttpContext(['ssl' => ['verify_peer'=>false]]);
```
Closes #2682

----
**Links:**
- https://www.php.net/manual/en/function.stream-context-create.php
- https://www.php.net/manual/en/context.php
- https://www.php.net/manual/en/context.ssl.php

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -156,14 +156,6 @@ class Dompdf
     private $protocol = "";
 
     /**
-     * HTTP context created with stream_context_create()
-     * Will be used for file_get_contents
-     *
-     * @var resource
-     */
-    private $httpContext;
-
-    /**
      * The system's locale
      *
      * @var string
@@ -400,7 +392,7 @@ class Dompdf
             $uri = $realfile;
         }
 
-        [$contents, $http_response_header] = Helpers::getFileContent($uri, $this->httpContext);
+        [$contents, $http_response_header] = Helpers::getFileContent($uri, $this->options->getHttpContext());
         if ($contents === null) {
             throw new Exception("File '$file' not found.");
         }
@@ -1244,12 +1236,12 @@ class Dompdf
     /**
      * Sets the HTTP context
      *
-     * @param resource $httpContext
+     * @param resource|array $httpContext
      * @return $this
      */
     public function setHttpContext($httpContext)
     {
-        $this->httpContext = $httpContext;
+        $this->options->setHttpContext($httpContext);
         return $this;
     }
 
@@ -1269,7 +1261,7 @@ class Dompdf
      */
     public function getHttpContext()
     {
-        return $this->httpContext;
+        return $this->options->getHttpContext();
     }
 
     /**
@@ -1363,6 +1355,11 @@ class Dompdf
      */
     public function setOptions(Options $options)
     {
+        // For backwards compatibility
+        if ($this->options && $this->options->getHttpContext() && !$options->getHttpContext()) {
+            $options->setHttpContext($this->options->getHttpContext());
+        }
+
         $this->options = $options;
         $fontMetrics = $this->getFontMetrics();
         if (isset($fontMetrics)) {
--- a/src/Options.php
+++ b/src/Options.php
@@ -276,6 +276,16 @@ class Options
     private $pdflibLicense = "";
 
     /**
+     * HTTP context created with stream_context_create()
+     * Will be used for file_get_contents
+     *
+     * @link https://www.php.net/manual/context.php
+     *
+     * @var resource
+     */
+    private $httpContext;
+
+    /**
      * @param array $attributes
      */
     public function __construct(array $attributes = null)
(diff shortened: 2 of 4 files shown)


src/Dompdf.php
```diff
@@ -1363,6 +1355,10 @@ public function getDom()
      */
     public function setOptions(Options $options)
     {
+        if ($this->options && $this->options->getHttpContext() && !$options->getHttpContext()) {
+            $options->setHttpContext($this->options->getHttpContext());
+        }
+
```
Mellthas, line 1361: I think I would leave this out, as none of the other options are merged like this.

src/Options.php
```diff
@@ -356,6 +366,8 @@ public function set($attributes, $value = null)
                 $this->setPdfBackend($value);
             } elseif ($key === 'pdflibLicense' || $key === 'pdflib_license') {
                 $this->setPdflibLicense($value);
+            } elseif ($key === 'setHttpContext' || $key === 'set_http_context' || $key === 'http_context') {
```
Mellthas, line 369: Should be `$key === 'httpContext' || $key === 'http_context'` for consistency.

src/Dompdf.php
erikn69: I think the same, buy i trying to avoid breaking changes, if you do this
https://github.com/dompdf/dompdf/blob/e76193db8a7f14de95f78888757e0c167929af3a/tests/DompdfTest.php#L38-L39
`$dompdf->setOptions(new Options());`  overwrites `setHttpContext`, i try to has the same functionality as before

tests/DompdfTest.php
```diff
@@ -49,6 +49,9 @@ public function testSetters()
         $this->assertInstanceOf(Options::class, $dompdf->getOptions());
         $this->assertEquals('test3', $dompdf->getProtocol());
         $this->assertInstanceOf(FrameTree::class, $dompdf->getTree());
+
+        $dompdf->setHttpContext(['ssl' => ['verify_peer'=>false]]);
+        $this->assertIsResource($dompdf->getHttpContext());
```
Mellthas, line 54: Would be nice to add tests for the new option to `OptionsTest`, too.

src/Options.php
erikn69: Fixed

src/Dompdf.php
Mellthas: Right, hadn’t thought of that. Then I would add a short comment explaining why the logic is there.

tests/DompdfTest.php
erikn69: Added

src/Dompdf.php
Mellthas: Maybe simply `// For backwards compatibility` would be enough. Just so it is easier to decide whether to remove the logic in a future major release (or deliberately keep it).

tests/DompdfTest.php
```diff
@@ -49,6 +49,10 @@ public function testSetters()
         $this->assertInstanceOf(Options::class, $dompdf->getOptions());
         $this->assertEquals('test3', $dompdf->getProtocol());
         $this->assertInstanceOf(FrameTree::class, $dompdf->getTree());
+
+        $dompdf = new Dompdf();
+        $dompdf->setHttpContext(['ssl' => ['verify_peer'=>false]]);
```
Mellthas, line 54: Small nit: Missing spaces around `=>`.

tests/OptionsTest.php
```diff
@@ -64,7 +64,8 @@ public function testSetters()
             'debugLayoutLines' => false,
             'debugLayoutBlocks' => false,
             'debugLayoutInline' => false,
-            'debugLayoutPaddingBox' => false
+            'debugLayoutPaddingBox' => false,
+            'httpContext' => ['ssl' => ['verify_peer'=>false]],
```
Mellthas, line 68: Same here.

Mellthas:
Apart from the remaining comments, I think the changes look fine.

tests/DompdfTest.php
erikn69: sorry, copy paste mistake

tests/OptionsTest.php
erikn69: Done

src/Dompdf.php
erikn69: added

Mellthas (approved):

PR #2814, opened 2022-03-18
Fix `Unsupported operand types: string * float` error in PHP 8.1

GenieTim:


bsweeney:
Do you have sample HTML+CSS we can use to confirm the error?

GenieTim:
Yes, sure, sorry. The following HTML leads to said error:

```html
<!DOCTYPE html>
<html lang="en">

<head>
  <meta charset="UTF-8">
  <meta http-equiv="X-UA-Compatible" content="IE=edge">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <title>Document</title>
  <style>
    .pure-table td {
      padding: 1em;
      font-size: inherit;
    }
  </style>
</head>

<body>
  <table class="pure-table">
    <tr>
      <td>4</td>
      <td>5</td>
      <td>6</td>
    </tr>
  </table>
</body>

</html>
```

Admittedly, my patch might be doing more than needed, while not doing what should be done (i.e. it only fixes the error/symptom rather than the actual origin of the problem).

<details>
<summary>Example code with call of dompdf</summary>


```php
<?php

require_once __DIR__ . "/../vendor/autoload.php";

use Dompdf\Dompdf;
use Dompdf\Options;

$options = new Options();
$options->setIsRemoteEnabled(true);
$options->setIsHtml5ParserEnabled(true);
$dompdf = new Dompdf($options);
$contxt = stream_context_create([
  'ssl' => [
    'verify_peer' => false,
    'verify_peer_name' => false,
    'allow_self_signed' => true,
  ],
  'curl' => [
    'curl_verify_ssl_host' => false,
    'curl_verify_ssl_peer' => false
  ]
]);
$dompdf->setHttpContext($contxt);

$html = <<<EOD
<!DOCTYPE html>
<html lang="en">

<head>
  <meta charset="UTF-8">
  <meta http-equiv="X-UA-Compatible" content="IE=edge">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <title>Document</title>
  <style>
    .pure-table td {
      padding: 1em;
      font-size: inherit;
    }
  </style>
</head>

<body>
  <table class="pure-table">
    <tr>
      <td>4</td>
      <td>5</td>
      <td>6</td>
    </tr>
  </table>
</body>

</html>

EOD;

$dompdf->loadHtml($html);
$dompdf->setPaper('A4', "portrait");
$dompdf->render();
echo $dompdf->output();

```
</details>

bsweeney:
Thanks for the sample that definitely does it.

I think this is maybe not how we want to solve the issue? The bug stems from the fact that style properties that inherit are now computed later in the process. At the same time, side widths (outline, margin, border, padding) are computed when the value is set. As a result, if both values are set within the same style declaration font-size is still "inherit" when padding is set.

Ultimately it doesn't matter since the values are re-computed later. But within the `single_length_in_pt` method the font-size is expected to be numeric (per the typed argument) so I think we should avoid casting the value. Because font-size is used internally when setting other properties it maybe deserves special handling so that it always has a valid computed value.

Mellthas:
See #2821 as alternative, which should fix the root of the problem.

PR #2819, opened 2022-03-19
Improve default font handling

bsweeney:


src/Css/Style.php
```diff
@@ -510,7 +512,7 @@ public function __construct(Stylesheet $stylesheet, $origin = Stylesheet::ORIG_A
             $d["elevation"] = "level";
             $d["empty_cells"] = "show";
             $d["float"] = "none";
-            $d["font_family"] = $stylesheet->get_dompdf()->getOptions()->getDefaultFont();
+            $d["font_family"] = empty($defaultFont) ? "serif" : $defaultFont;
```
Mellthas, line 515: Given that we had a bug due to using `empty` for checking for an empty string not too long ago (see #2566), I think I would avoid using the function here, even if a font with name `0` is highly unlikely in practice. Simplest approach is probably using `$defaultFont === ""` here and making sure that `$options->getDefaultFont()` always returns a string.

Mellthas, line 515: Given that we had a bug due to using `empty` for checking for an empty string not too long ago (see #2566), I think I would avoid using the function here, even if a font with name `0` is highly unlikely in practice. Simplest approach is probably using `$defaultFont === ""` here and making sure that `$options->getDefaultFont()` always returns a string.

src/FontMetrics.php
```diff
@@ -427,38 +427,44 @@ public function getFont($familyRaw, $subtypeRaw = "normal")
             return null;
         }
 
-        $family = "serif";
-
-        if (isset($this->fontLookup[$family][$subtype])) {
-            return $cache[$familyRaw][$subtypeRaw] = $this->fontLookup[$family][$subtype];
+        $fallback_families = [];
+        if (!empty($this->options->getDefaultFont())) {
```
Mellthas, line 431: Same here with `empty`.

src/Css/Style.php
bsweeney: A font named "0" ... that would never happen. Right? 😅 

I think I'm cool with that suggestions, in which case no changes are required in the Style class. I was primarily trying to avoid errors when the default font was set to `null`, which would break other logic within this class.

Instead we can modify Options to always set defaultFont to a string value. And if the user tries to set defaultFont to null or an empty string that we use the standard default ("serif")?

Mellthas: That’s sounds even better, yes.

bsweeney: Or, I guess, better to leave the current value instead of "serif" for an empty string.

bsweeney: 👍?

Mellthas: Spontaneously, I would lean towards interpreting these as the default value (so `"serif"`). Might also be better for backwards compatibility:
```
$options->setDefaultFont("sans-serif");
$options->setDefaultFont("");
```
would lead to sans-serif instead of serif.

bsweeney: I see what you mean. If the user sets null/empty then reset the value, rather than silently drop the value.

Mellthas: Yeah, feel that would be more natural.

Mellthas (approved):

PR #2820, opened 2022-03-19
Keep list-style-type's single character case

rudiedirkx:
Currently `<ol type="A">` will be saved with `list_style_type` = `a` instead of `A`, because of the `mb_strtolower` in `set_prop`.

The workaround is very easy: use `upper-alpha` instead of `A`. But still. `type="A"` is valid HTML.

### DIFF
--- a/tests/Css/StyleTest.php
+++ b/tests/Css/StyleTest.php
@@ -160,4 +160,25 @@ class StyleTest extends TestCase
         $style->set_prop("content", $value);
         $this->assertSame($expected, $style->content);
     }
+
+    public function valueCaseProvider(): array
+    {
+        return [
+            ["width", "Auto",           "width", "auto"],
+            ["list-style-type", "A",    "list_style_type", "A"],
+        ];
+    }
+
+    /**
+     * @dataProvider valueCaseProvider
+     */
+    public function testValueCase(string $cssProp, string $inputValue, string $phpProp, string $expectValue): void
+    {
+        $dompdf = new Dompdf();
+        $sheet = new Stylesheet($dompdf);
+        $style = new Style($sheet);
+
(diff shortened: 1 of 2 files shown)


Mellthas (changes requested):
Since I just recently removed the length limit in d3c39190e20e02bd046a841ee4b6a187e8bd1d1e, I think it would be simpler to add it back in a modified form by simply using `strlen($val) > 1` as additional condition here, without special case for `list-style-type`.

Doing the lower-casing at this point is really not ideal, but barring a more substantial rework, keeping the condition as simple as possible is preferable in my view.

rudiedirkx:
I agree with all points. And I've added a test for your `auto` and my `A`, because this might come back, and I don't know any other relevant cases, but they're easy to add now.

tests/Css/StyleTest.php
```diff
@@ -160,4 +160,25 @@ public function testContent(string $value, $expected): void
         $style->set_prop("content", $value);
         $this->assertSame($expected, $style->content);
     }
+
+    public function valueCaseProvider(): array
+    {
+        return [
+            ["width", "Auto",           "width", "auto"],
+            ["list-style-type", "A",    "list_style_type", "A"],
```
Mellthas, line 168: Adding some tests is nice! The underscore variant works for both getter and `set_prop`, would be just 3 parameters.

rudiedirkx: But `set_prop` is called with dashes, so that's how I'd test it. In case there's an exception for `list-style-type` or `font-family` or something later.

Mellthas: You have a point there. It’s fine for me either way.

Mellthas (approved):

PR #2823, opened 2022-03-20
Use Github Actions for tests

rudiedirkx:
Let Github run unit tests automatically.

I disabled Scrutinizer coverage upload, because I don't know if Dompdf does that.

(Taken from https://github.com/kristijanhusak/laravel-form-builder, thanks to @roelofr)


.github/workflows/test.yml
```diff
@@ -0,0 +1,54 @@
...
+  COMPOSER_MEMORY_LIMIT: -1
+
+jobs:
+  test:
+    name: "Build"
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 12
+      matrix:
+        php: ['7.4', '8.0', '8.1']
```
bsweeney, line 17: Current PHP requirement is 7.1 and higher. Though those are no longer actively developed, as long as we haven't raised the minimum requirement we should run on those versions as well.

bsweeney (changes requested):
I'm all on board for this. One small request.

.github/workflows/test.yml
rudiedirkx: You're right. I copied those versions from my fav gh action file. I'll include 7.1+.

rudiedirkx:
Only you can trigger this PR's gh actions actually running:

> First-time contributors need a maintainer to approve running workflows

I think you can do that on https://github.com/dompdf/dompdf/actions

Might be smart to do that before merging =)

Mellthas:
I think it would also be nice to have a job for checking code style via `phpcs`, though I suppose this can also easily be added later.

Mellthas:
The Travis config should be obsolete then, shouldn’t it?

rudiedirkx:
> I think it would also be nice to have a job for checking code style via `phpcs`, though I suppose this can also easily be added later.

I don't know how to do that one. You can add it, but testing it isn't easy, because it has to go through gh actions. Adding a step is easy though.

> The Travis config should be obsolete then, shouldn’t it?

I've never seen a Travis run. Is that triggered automatically somewhere?

Mellthas:
> I don't know how to do that one. You can add it, but testing it isn't easy, because it has to go through gh actions. Adding a step is easy though.

Then don’t bother, unless you want to look into it. Was mostly meant as a reminder.

> I've never seen a Travis run. Is that triggered automatically somewhere?

I think it isn’t run anymore since travis-ci.org closed down/was moved to travis-ci.com. @bsweeney might know more.

bsweeney:
I didn't set up the TravisCI integration so I don't have much to add. We obviously haven't been maintaining it so I'm happy to just remove the relevant configuration. Happy to do that after the merge.

rudiedirkx:
I've removed Travis, and added a code sniffer step. It seems to work.

.github/workflows/test.yml
```diff
@@ -0,0 +1,57 @@
...
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: composer-${{ runner.os }}-${{ matrix.php }}-${{ matrix.package-release }}-${{ hashFiles('**/composer.json') }}
+          restore-keys: |
+            composer-${{ runner.os }}-${{ matrix.php }}-${{ matrix.package-release }}-${{ env.cache-name }}-
+            composer-${{ runner.os }}-${{ matrix.php }}-${{ matrix.package-release }}-
+            composer-${{ runner.os }}-${{ matrix.php }}-
+            composer-${{ runner.os }}-
+
+      - name: Install composer dependencies
+        run: composer install --no-progress --no-interaction --prefer-${{ matrix.package-release }}
```
roelofr, line 45: This behaviour is, while correct, often replaced by the [ramsey/composer-install](https://github.com/ramsey/composer-install) action. Therefore, this would be better:

```suggestion
      - name: Install composer dependencies
        uses: ramsey/composer-install@v1
        with:
          dependency-versions: ${{ matrix.package-release }}
```

But, since the `package-release` is set to `dist` and nothing else, you may just remove that last part:

```yaml
      - name: Install composer dependencies
        uses: ramsey/composer-install@v1
```

Also remove the `package-release` entry in the matrix.

PR #2825, opened 2022-03-21
Drop invalid border styles

bsweeney:


src/Css/Style.php
```diff
@@ -2073,6 +2073,10 @@ protected function _set_style_side_type($style, $side, $type, $val)
 
                 $this->_props_computed[$prop] = $has_line_style ? $val_computed : 0;
             }
+        } elseif (($style === "border" || $style === "outline") && $type === "style") {
+            if (in_array($val, Style::$BORDER_STYLES)) {
```
Mellthas, line 2077: I prefer passing `true` by default to `in_array`, as to avoid any unexpected comparison behavior (and to be consistent with `===`). Though I guess it doesn’t really matter in this case.

src/Css/Style.php
```diff
@@ -2073,6 +2073,10 @@ protected function _set_style_side_type($style, $side, $type, $val)
...
+                $this->_props_computed[$prop] = $val;
+            }
```
Mellthas, line 2079: For consistency, would set `$this->_props_computed[$prop]` to `null` otherwise, as this might be called as part of one of the public setter methods.

src/Css/Style.php
bsweeney: Probably not but I'll make this change anyway. If nothing else it's more consistent usage.

src/Css/Style.php
bsweeney: Good catch I missed that it wasn't being set otherwise.

Mellthas (approved):

PR #2822, opened 2022-03-20
Stop inline reflow early when reset

Mellthas:
Fixes #2763
Alternative to #2772

PR #2850, opened 2022-04-09
Fix crash in inline reflower when child reflow moves all children

### LINKED ISSUE #2763: Call to a member function get_position() on null
I'm using ` "dompdf/dompdf": "dev-master",` and i'm getting this error

`Call to a member function get_position() on null`

in `vendor/dompdf/dompdf/src/FrameReflower/Inline.php`  on line `163`

```
        // Assume the position of the first child
        [$x, $y] = $frame->get_first_child()->get_position();
```

this is my code to generate the pdf

```
$options = new Options();
        $options->set('isRemoteEnabled', TRUE); //If this setting is set to true, DOMPDF will access remote sites for images and CSS files as required.
        $dompdf = new DOMPDF($options);
        
        // images are hosted on a server with a self-signed security certificate or other certificate problems, creating and setting a stream context will be necessary.
        $context = stream_context_create([ 
            'ssl' => [ 
                'verify_peer' => FALSE, 
                'verify_peer_name' => FALSE,
                'allow_self_signed'=> TRUE 
            ] 
        ]);
        $dompdf->setHttpContext($context);

        $dompdf->loadHtml($view); 

        $dompdf->setPaper('A4', 'portrait');

        $dompdf->render();

        // Parameters
        $x          = 505;
        $y          = 820;
        $text       = "Page {PAGE_NUM} of {PAGE_COUNT}";     
        $font       = $dompdf->getFontMetrics()->get_font('Helvetica', 'normal');   
        $size       = 10;    
        $color      = [0,0,0];
        $word_space = 0.0;
        $char_space = 0.0;
        $angle      = 0.0;

        $dompdf->getCanvas()->page_text(
            $x, $y, $text, $font, $size, $color, $word_space, $char_space, $angle
        );

        $dompdf->stream($filename, [ "Attachment" => (YII_ENV) ? 0 : 1]); //set to 1 to auto download

        Yii::$app->end();
```

any idea why? my code works sometimes too. not sure what's making it break on some of my html (`$view`)

Mellthas:
The fix in #2822 catered for cases where the frame itself is moved and reset. When a page break occurs before the first child, which can happen when an inline frame contains a block frame, the frame is left empty after child reflow. Make sure to not crash in that case.

The frame is left without position, but as empty inline frames are not drawn currently, this should be fine for now.

Fixes #2763

### DIFF
--- a/src/FrameReflower/Inline.php
+++ b/src/FrameReflower/Inline.php
@@ -165,6 +165,10 @@ class Inline extends AbstractFrameReflower
             }
         }
 
+        if (!$frame->get_first_child()) {
+            return;
+        }
+
         // Assume the position of the first child
         [$x, $y] = $frame->get_first_child()->get_position();
         $frame->set_position($x, $y);


icelake0:
This change fixes an issue I am experiencing on a production server
Kindly help look into merging this PR if possible.

Mellthas:
@bsweeney I have added the PR to the 2.0.0 milestone, like the linked issue. But maybe it makes sense to do another small bugfix release before that?

icelake0:
> @bsweeney I have added the PR to the 2.0.0 milestone, like the linked issue. But maybe it makes sense to do another small bugfix release before that?

I agree with @bsweeney on this, My team is using another [package](https://github.com/barryvdh/laravel-dompdf/blob/v1.0.0/composer.json) that depends on v:^1 of this .
A bug fix release will go a long way to help users of both packages close out on issues like this

bsweeney:
I think you're right I'll make it happen.

PR #2859, opened 2022-04-19
Improvements to resource handling

bsweeney:


src/Options.php
```diff
@@ -1008,4 +1100,41 @@ public function getHttpContext()
     {
         return $this->httpContext;
     }
+
+    public function validateLocalUri(string $uri) {
```
Mellthas, line 1104: New line before `{` missing.

src/Options.php
```diff
@@ -1008,4 +1100,41 @@ public function getHttpContext()
...
+        }
+
+        if (!$realfile) {
+            return [false, "File not found."];
+        }
+
+        return [true, null];
+    }
+
+    public function validatePharUri(string $uri) {
```
Mellthas, line 1128: New line before `{` missing.

src/Options.php
```diff
@@ -1008,4 +1100,41 @@ public function getHttpContext()
...
+        $file = substr(substr($uri, 0, strpos($uri, ".phar") + 5), 7);
+        return $this->validateLocalUri("$file");
+    }
+
+    public function validateRemoteUri(string $uri) {
```
Mellthas, line 1133: New line before `{` missing.

src/Options.php
```diff
@@ -1008,4 +1100,41 @@ public function getHttpContext()
...
+        if (!$realfile) {
+            return [false, "File not found."];
+        }
+
+        return [true, null];
+    }
+
+    public function validatePharUri(string $uri) {
+        $file = substr(substr($uri, 0, strpos($uri, ".phar") + 5), 7);
+        return $this->validateLocalUri("$file");
```
Mellthas, line 1130: `"$file"` -> `$file`

tests/OptionsTest.php
```diff
@@ -95,4 +95,55 @@ public function testSetters()
...
+        $options->setAllowedProtocols("http://");
+        $allowedProtocols = $options->getAllowedProtocols();
+        $this->assertIsArray($allowedProtocols);
+        $this->assertEquals(1, count($allowedProtocols));
+        $this->assertArrayHasKey("http://", $allowedProtocols);
+        $this->assertIsArray($allowedProtocols["http://"]);
+        $this->assertArrayHasKey("rules", $allowedProtocols["http://"]);
+        $this->assertIsArray($allowedProtocols["http://"]["rules"]);
+        $this->assertEquals(1, count($allowedProtocols["http://"]["rules"]));
+        $this->assertEquals(array($options, "validateRemoteUri"), $allowedProtocols["http://"]["rules"][0]);
```
Mellthas, line 111: Using the old array syntax `array()` here.

src/Image/Cache.php
```diff
@@ -64,116 +57,79 @@ class Cache
      */
     static function resolve_url($url, $protocol, $host, $base_path, Dompdf $dompdf)
     {
-        self::$_dompdf = $dompdf;
+        $options = $dompdf->getOptions();
```
Mellthas, line 60: Maybe use the opportunity to pass the options directly instead of the Dompdf instance here? Would make the dependency clearer.

src/Image/Cache.php
```diff
@@ -64,116 +57,79 @@ class Cache
...
-
-        $remote = ($protocol && $protocol !== "file://") || ($parsed_url['protocol'] !== "");
-
-        $data_uri = strpos($parsed_url['protocol'], "data:") === 0;
-        $full_url = null;
-        $enable_remote = $dompdf->getOptions()->getIsRemoteEnabled();
-        $tempfile = false;
+        
+        $protocol = strtolower($parsed_url["protocol"]);
+        $is_data_uri = strpos($parsed_url['protocol'], "data:") === 0;
```
Mellthas, line 71: Would use double quotes for `'protocol'` here, too.

Mellthas (changes requested):
General observations while testing:
* Local paths with spaces don’t seem to work anymore, at least when used with `img`.
* Using a local path outside chroot: Warning is _Image not readable or empty_ instead of _Permission denied on $path. The file could not be found under the paths specified by Options::chroot_.

src/Options.php
```diff
@@ -499,6 +520,77 @@ public function setChroot($chroot, $delimiter = ',')
...
+                if (is_string($protocol)) {
+                    $protocols[$protocol] = [];
+                    if (is_array($config)) {
+                        $protocols[$protocol] = $config;
+                    }
+                } elseif (is_string($config)) {
+                    $protocols[$config] = [];
+                }
+            }
+        }
```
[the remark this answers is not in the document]
bsweeney: This makes sense. I think my head was stuck in accommodating usage similar to other Options, but this is new functionality so there's no reason to do something like that.

src/Options.php
```diff
@@ -310,6 +325,8 @@ public function __construct(array $attributes = null)
             ]
         ]);
 
+        $this->setAllowedProtocols(",file://,http://,https://");
```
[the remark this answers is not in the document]
bsweeney: ha ha don't you regularly deal with people writing weird logic?

src/Image/Cache.php
bsweeney: I was trying to decide between that change and API compatibility and settled on compatibility. This is a new major release, though, so a breaking change would be OK. Not that (m)any people would be calling this directly.

src/Options.php
bsweeney: Now why did I even do that? I think I might have been combining vars in the call to validateLocalUri in an earlier revision.

bsweeney:
Thanks for the review. I was hoping you'd find some time for it.

src/Image/Cache.php
Mellthas: Or rather `$parsed_url['protocol']` -> `$protocol`

Mellthas:
> Local paths with spaces don’t seem to work anymore, at least when used with `img`.

Looks like this is already broken on `master`, probably due to the new HTML parser. Spaces seem to be always encoded as `%20` now when read from the `src` attribute.

src/Image/Cache.php
```diff
@@ -64,116 +57,79 @@ class Cache
...
-                    }
+            if (!isset($resolved_url)) {
+                $tmp_dir = $options->getTempDir();
+                if (($resolved_url = @tempnam($tmp_dir, "ca_dompdf_img_")) === false) {
+                    throw new ImageException("Unable to create temporary image in " . $tmp_dir, E_WARNING);
                 }
-            } // Not remote, local image
-            else {
-                $resolved_url = Helpers::build_url($protocol, $host, $base_path, $url);
+                self::addTempImage($full_url, $resolved_url);
```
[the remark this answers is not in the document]
bsweeney: Since we use tempnam to generate `$resolved_url` the temporary file exists at this point. If we throw an exception before we add an entry to the cache the temp file will be orphaned. Adding the resolved URL to the directory of temporary images ensures the temp file gets cleaned up when the cache is cleared.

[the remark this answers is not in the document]
Mellthas: Then using `$tempfile = $resolved_url;` as it was before seems simpler. Lines 168–170 already handle cleaning the temp file in case of an exception.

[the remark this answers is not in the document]
bsweeney: when you're right you're right

src/Options.php
```diff
@@ -298,6 +298,18 @@ public function __construct(array $attributes = null)
         $this->setFontCache($this->getFontDir());
         $this->setLogOutputFile($this->getTempDir() . "/log.htm");
 
+        $ver = "";
+        $versionFile = realpath(__DIR__ . "/../VERSION");
+        if (file_exists($versionFile) && ($version = trim(file_get_contents($versionFile))) !== false && $version !== '$Format:<%h>$') {
```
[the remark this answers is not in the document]
bsweeney: Good catch on the trim misuse. I went ahead and changed the same logic in Dompdf.php

src/Image/Cache.php
```diff
@@ -64,116 +57,79 @@ class Cache
...
+        
+        $protocol = strtolower($parsed_url["protocol"]);
+        $is_data_uri = strpos($parsed_url['protocol'], "data:") === 0;
 
         try {
-
-            // Remote not allowed and is not DataURI
-            if (!$enable_remote && $remote && !$data_uri) {
-                throw new ImageException("Remote file access is disabled.", E_WARNING);
+            if (isset(self::$_cache[$full_url])) {
```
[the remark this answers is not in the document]
bsweeney: Yes I that's a good analysis of the intended logic.

bsweeney:
Small update based on your feedback. I need to look into the issue with local paths that contain spaces. It does appear to be html5-php doing the conversion.

bsweeney:
The issue with space encoding could be overly aggressive URI encoding in html5-php. In a quick reading of the spec the [URL parsing](https://url.spec.whatwg.org/#url-parsing) state machine should encode space characters as part of the fragment percent-encode set or the query percent-encode set, neither of which should apply in the case of a file's path.

A quick test shows that the following ASCII characters are encoded:
- $
- ^
- |
- {
- }
- \
- [
- ]
- {space}

A bit more research is probably in order, but should we plan on running local paths through urldecode?

Mellthas:
It looks like it is actually `DOMDocument::saveHTML` doing the aggressive encoding. `DOMDocument::saveXML` doesn’t do this, but adds an XML declaration. Using `HTML5::saveHTML` seems to work as intended.

Mellthas:
> Should we plan on running local paths through urldecode?

From some limited browser testing, local paths should work with percent encoded characters (so e.g. using `%20` instead of a space). So using `rawurldecode` for local paths might be the right thing to do. But since this also doesn’t work with previous dompdf versions, I don’t think this needs to block the current work.

src/Image/Cache.php
```diff
@@ -62,125 +55,115 @@ class Cache
      * @throws ImageException
      * @return array             An array with two elements: The local path to the image and the image extension
      */
-    static function resolve_url($url, $protocol, $host, $base_path, Dompdf $dompdf)
+    static function resolve_url($url, $protocol, $host, $base_path, Options $options)
```
Mellthas, line 58: The doc block is still referencing the old `$dompdf` parameter. Could also remove the `@throws ImageException` while at it, as the exception is never actually thrown.

src/Image/Cache.php
```diff
@@ -62,125 +55,115 @@ class Cache
...
+                    function ($parser, $name, $attributes) use ($dompdf, $parsed_url, $full_url) {
+                        if ($name === "image") {
+                            $attributes = array_change_key_case($attributes, CASE_LOWER);
+                            $url = $attributes["xlink:href"] ?? $attributes["href"];
+                            if (!empty($url)) {
+                                $inner_full_url = Helpers::build_url($parsed_url["protocol"], $parsed_url["host"], $parsed_url["path"], $url);
+                                if ($inner_full_url === $full_url) {
+                                    throw new ImageException("SVG self-reference is not allowed", E_WARNING);
+                                }
+                                [$resolved_url, $type, $message] = self::resolve_url($url, $parsed_url["protocol"], $parsed_url["host"], $parsed_url["path"], $dompdf);
```
Mellthas, line 143: `$dompdf` -> `$options`

src/Image/Cache.php
```diff
@@ -62,125 +55,115 @@ class Cache
...
-                    //Don't put replacement image into cache - otherwise it will be deleted on cache cleanup.
-                    //Only execute on successful caching of remote image.
-                    if ($enable_remote && $remote || $data_uri) {
-                        self::$_cache[$full_url] = $resolved_url;
+            if ($type === "svg") {
+                $parser = xml_parser_create("utf-8");
+                xml_parser_set_option($parser, XML_OPTION_CASE_FOLDING, false);
+                xml_set_element_handler(
+                    $parser,
+                    function ($parser, $name, $attributes) use ($dompdf, $parsed_url, $full_url) {
```
Mellthas, line 134: `$dompdf` -> `$options`

Mellthas (changes requested):

src/Image/Cache.php
bsweeney: how did I miss that smh

bsweeney:
I checked earlier versions and the space issue does indeed appear to be long-standing. So, I'm fine delaying a fix until we have more time to consider our options.

Mellthas:
One more issue: Disabling the `file://` protocol via `allowedProtocols` doesn’t work if local paths are accepted, as `Helpers::build_url` normalizes such URLs to local paths. But indeed, disabling local paths breaks dompdf, as it can’t even load the default style sheet then. Maybe the `""` and `file://` protocols should always be allowed?

bsweeney:
I need to think on that one. Maybe we treat "" as "file://" for protocol checks and not have "" in allowedProtocols. Disabling file:// shouldn't totally break things, I think I need to tweak the logic to allow resource retrieval from within the lib directory. Previous logic had a similar hard-coded allowance for chroot checks.

Mellthas:
> Disabling file:// shouldn't totally break things, I think I need to tweak the logic to allow resource retrieval from within the lib directory. Previous logic had a similar hard-coded allowance for chroot checks.

That sounds more sensible, yes.

bsweeney:
OK, a few changes here. The issue preventing loading the base stylesheet is fixed. And instead of the "" protocol I leaned into using the "file://" protocol. I need to do a bit more testing around file:// references but so far I haven't run into any issues.

src/Adapter/CPDF.php
```diff
@@ -738,7 +738,6 @@ public function circle($x, $y, $r1, $color, $width = null, $style = null, $fill
      * @param string $image_url
      * @param string $type
      *
-     * @throws Exception
      * @return string The url of the newly converted image
```
Mellthas, line 741: Return type is now `string|null`.

src/Css/Stylesheet.php
```diff
@@ -1672,7 +1650,7 @@ public function getDefaultStylesheet()
     {
         $options = $this->_dompdf->getOptions();
         $rootDir = realpath($options->getRootDir());
-        return $rootDir . self::DEFAULT_STYLESHEET;
+        return Helpers::build_url("file://", "", $rootDir, $rootDir . self::DEFAULT_STYLESHEET);
```
Mellthas, line 1649: Just appending `"file://"` should be enough here, no need for `build_url`.

Mellthas (changes requested):

Mellthas:
Observation regarding the use of the`file://` protocol: Links to existing local files are no longer added to the PDF when using the CPDF adapter, the protocol in front seems to prevent that. Though I couldn’t find a way to specify such links so that they would actually work with the old code (link is in the PDF, but the path does not work).

bsweeney:
> It seems it is just Firefox not understanding the `file://` protocol for links (when viewing PDF), seems to work fine in other PDF viewers.

Hm ... that's unfortunate. I was thinking of removing the file:// prior to creating the link but when it was OK in the testing I did do I thought maybe it was OK as-is. I'll do some more testing.

src/Css/Stylesheet.php
Mellthas: Okay, now I understand. Is building `$full_url` even necessary in any case?

bsweeney:
Looking into Firefox a bit more and it looks like behavior with PDFs is to _not_ render link actions if the URI points to the local filesystem ([ref](https://github.com/mozilla/pdf.js/issues/9637#issuecomment-393482383)), with or without the "file://" at the front. With this in mind I think the change in URI handling with this PR is OK since the behavior is viewer-dependent.

I do feel like some work could be put in to better build out file-based URIs, but I didn't want to touch that logic too much. A project for a future time.

src/Css/Stylesheet.php
bsweeney: It's true that `explode_url` already determines the protocol is "file://" if no protocol is found. Which is really all we need here.

I guess the one use case to consider is a user manually loading a stylesheet outside the HTML. `build_url` will ensure that the path is normalized if, for example, the user were to provide a relative filesystem reference. Though even that use case looks excessive since chroot checks are made after running local files through realpath.

src/Dompdf.php
```diff
@@ -352,43 +345,25 @@ public function loadHtmlFile($file, $encoding = null)
...
-                    $chrootValid = true;
-                    break;
-                }
-            }
-            if ($chrootValid !== true) {
-                throw new Exception("Permission denied on $file. The file could not be found under the paths specified by Options::chroot.");
-            }
-
-            $ext = strtolower(pathinfo($realfile, PATHINFO_EXTENSION));
+        if (strpos($uri, "://") === false) {
```
Mellthas, line 355: Should probably be `$protocol === "file://"` now?

src/Helpers.php
```diff
@@ -57,28 +57,34 @@ public static function pre_r($mixed, $return = false)
...
+            $protocol = "phar://";
         }
 
         $ret = "";
-        if ($protocol !== "file://") {
-            $ret = $protocol;
-        }
 
-        if (!in_array(mb_strtolower($protocol), ["http://", "https://", "ftp://", "ftps://"], true)) {
+        $is_local_path = in_array(strtolower($protocol), ["file://", "phar://"], true);
```
Mellthas, line 85: `$protocol` is already lower-cased at the start.

src/Helpers.php
```diff
@@ -868,22 +873,23 @@ public static function imagecreatefrombmp($filename, $context = null)
...
+     * @param int $maxlen
      * @return string[]
      */
     public static function getFileContent($uri, $context = null, $offset = 0, $maxlen = null)
     {
         $content = null;
         $headers = null;
         [$protocol] = Helpers::explode_url($uri);
-        $is_local_path = ($protocol === "" || $protocol === "file://");
+        $is_local_path = in_array(strtolower($protocol), ["", "file://", "phar://"], true);
```
Mellthas, line 886: Checking for `""` here should be unnecessary now.

src/Css/Stylesheet.php
bsweeney: Though we parse the default rootDir through `realpath` it's possible the user might modify the option. The user-specified value is not currently parsed through `realpath`. So while not optimal this is a fairly straightforward method to get the actual file path.

The way some of these default values are handled, and the ability for the user to customize, could be reworked in a future release.

src/Helpers.php
bsweeney: Since it's a public method I left that check in there for people calling the method directly from their own code. Someone actually using this method that way seems unlikely, but with a large enough user base ...

src/Helpers.php
bsweeney: True though in looking at that method I realized some of the comparison logic was not sufficiently case-insensitive, so a few minor changes are in order.

src/Dompdf.php
bsweeney: oh my ... yes!

src/Css/Stylesheet.php
Mellthas: The `rootDir` option is passed through `realpath` here, directly above, no matter if the default or user-specified:

```php
$rootDir = realpath($options->getRootDir());
```

Or do you mean something else?

src/Helpers.php
Mellthas: Good point.

src/Css/Stylesheet.php
bsweeney: No, you're right. I was looking at the Options class when I was making that comment and didn't follow the code all the way through.

I still think the current logic works. The Options class doesn't run the user-specified value through realpath. If the user modifies `rootDir` but specifies a relative path the current logic ensures that the default stylesheet path is fully normalized.

We don't need to build the URL in load_css_file since internal use of the load_css_file method is after already building the file URL. Anyone attempting to manually load CSS would just need to ensure that their URL conforms to expectations (has a protocol, or they've enabled URIs that don't specify a protocol).

Mellthas: True, I guess its fine.

Mellthas (approved):

Issue #1848, opened 2018-11-29
Relative paths to files in html become absolute


Ujifman:
I'm using this lib to make set of pdfs from set of html. 
So I have about 20 html files with links to each other, and I have to convert them into set of pdfs with working links to pdfs. Links in html are relative. 
```
<html xmlns="http://www.w3.org/1999/xhtml" lang="en-GB" xml:lang="en-GB">
	<body xmlns="http://www.w3.org/1999/xhtml" onload="init()">
		<a xmlns="http://www.w3.org/1999/xhtml" href="another.pdf">LINK</a>
	</body>
</html>
```
I change them to link on pdfs, and then convert htmls to pdfs. After conversion I have absolute links in pdfs, so set of pdfs is not moveable. The problem is in `build_url` method of `Helpers.php`. 
```
if ($url[0] !== '/' && (strtoupper(substr(PHP_OS, 0, 3)) !== 'WIN' || (mb_strlen($url) > 1 && $url[0] !== '\\' && $url[1] !== ':'))) {
                // For rel path and local acess we ignore the host, and run the path through realpath()
                $ret .= realpath($base_path) . '/';
            }
```
I don't fully understand reason to use `realpath`, and make path absolute. Why it is made like this?

bsweeney:
This is probably something worth looking into to change. Note that I think we'd also have to change the object type from a URI action to something else. There may be a few ways we could do this, such as a GoToR action.

Another thing this highlights ... that's not a valid URI since it doesn't include a scheme. I wonder if we could use a file scheme and a relative path (e.g. file:another.pdf or something like that).

PR #2887, opened 2022-05-23
don't change relative file url

ta-tikoma:
Ability to create relative links to files.

Example: 
I want to build archive with pdf file and other files. In pdf file I have links on other files.
I know archive structure and want to make correct links.

### DIFF
--- a/src/Helpers.php
+++ b/src/Helpers.php
@@ -70,6 +70,11 @@ class Helpers
 
         if (strpos($url, "file://") === 0) {
             $url = substr($url, 7);
+
+            if ($url[0] === '.') {
+                return $url;
+            }
+
             $protocol = "";
         }
 


ta-tikoma:
@bsweeney thanks for your review.
When this library create pdf with links on file (with protocol: ```file://``) it don't know where pdf will be used and which files structure will be. Only application creator know it. Other recourses (images, external stylesheets) does't use protocol ```file:://```.

bsweeney:
I understand the issue, see #1848. The change as written addresses one particular use case, but other use cases are valid. Some users _do_ reference images and stylesheets from the local filesystem.

bsweeney:
Current change may be acceptable. I'll review in the context of usage within the library.

ta-tikoma:
@bsweeney thanks for your time

bsweeney:
I ended up needing to go a different route because of how this method is used internally (see c9f1b02f). Instead of modifying the build_url method I updated the link logic to fall back to the specified HREF if build_url is unable to process the URI. The effect is the same, links to non-existent files (on the local filesystem) are maintained as-is.

PR #2894, opened 2022-05-25
Refactor setting used values

Mellthas:
This implements the second suggestion from https://github.com/dompdf/dompdf/pull/2722#issuecomment-1005690489.

During layout, properties are resolved to their used values. These values don’t need to be parsed and computed anymore, so they can be set more efficiently than during the style-parsing stage. There are two ways how used values are set currently: Either directly on the frame’s `style` or on the frame’s `original_style`. The difference is effectively that values set on the original style are preserved after a frame reset (page or line break), whereas values set on the actual style are reset instead (because `style` is reset to a clone of `original_style`).

For efficiency and clarity, replace `original_style` with the use of two different property setters:

* `$style->$prop = $value` is equivalent to setting the property on `original_style` and `style` at the same time before
* `$style->set_used($prop, $value)` is equivalent to setting the property on `style` only before

Both methods set the used value to the given value, but the former also sets the specified and computed value, so that a style reset keeps the value. I played around with different signatures, e.g. making `$style->$prop = $value` perform as `$style->set_used($prop, $value)` and having a dedicated method for the first case instead; this would result in less changes throughout the code base, but the current state feels more natural and easier to understand. Suggestions welcome.

Improves overall rendering performance and memory usage slightly, as only one style object is managed per frame.

Summary of the method changes:
* `Style::set_prop()` explicitly takes string values only now. The given value is parsed and computed. Used during style parsing
* `Style::__set()` sets the specified and used value of a property. The used value will not be cleared on style reset. Essentially similar to `Style::set_prop()` while skipping value parsing and computation
* `Style::set_used()` sets the used value of a property only. The used value will be cleared on style reset
* `Style::get_specified()` retrieves the specified value of a property. Replaces usage of `original_style` to get the specified value


bsweeney (approved):
I like where you're going with this.

Mellthas:
Thanks!

PR #2913, opened 2022-06-09
Fix IDE warning about `getAttribute` method

### LINKED ISSUE #2693: DOMNode does not have getAttribute method
In the Dompdf class (line number 605) We're calling the getAttribute method. The usage here is not quite correct since the method is part of the DOMElement class and not the DOMNode class. And the collection we're pulling the element from is a DOMNode collection. An error is not thrown during runtime since the node we're accessing is actually a DOMElement. But the specific logic we use can be better written.

_Originally posted by @bsweeney in https://github.com/dompdf/dompdf/discussions/2692#discussioncomment-1786838_

Mellthas:
Fixes #2693

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -558,9 +558,11 @@ class Dompdf
         $acceptedmedia[] = $this->options->getDefaultMediaType();
 
         // <base href="" />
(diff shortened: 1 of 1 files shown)


bsweeney (approved):
I was totally looking at this and your solution is almost exactly the same as mine.

PR #2915, opened 2022-06-09
Suggestions for callback handling

Mellthas:
Here are 3 suggestions for improving callback handling. As number 1 and 3 would be breaking changes, the 2.0 release might be a good time to do these. Though I am not really sure whether number 3 is worth the break, as it is purely for convenience and consistency; see the adapted test cases for how it would look in practice.

### 1. Make `page_script` take effect immediately
Existing issues this aims to resolve:
* Calling `output` multiple times after render will stack up additional invocations of all `page_*` method calls.
* Calling the canvas `page_*` methods before render does not work when the canvas is replaced on render start. This happens when the default paper size and orientation do not match the document.

I think that this behavior is also easier to understand. The `page_*` methods now need to be called after render or, if used in embedded PHP code, at the end of the document.

### 2. Add an `end_document` callback event

As replacement for the old `page_script` method behavior. This allows the page script to be set up before render, while making sure that it is actually called after render (and not on output).

This should allow pretty much every basic use case to be handled with dompdf callbacks only.

### 3. Modify dompdf callback signature

Make the callbacks receive three parameters `$frame`, `$canvas`, and `$fontMetrics` instead of an info array; for ease of use and to be consistent with the `end_document` callback and `page_script`.

### DIFF
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -146,13 +146,6 @@ class CPDF implements Canvas
     protected $_page_count;
 
     /**
-     * Callbacks to process on every page after rendering is complete
-     *
-     * @var callable[]
-     */
-    protected $pageCallbacks;
-
-    /**
      * Array of pages for accessing after rendering is initially complete
      *
      * @var array
@@ -201,7 +194,6 @@ class CPDF implements Canvas
         $this->_height = $size[3] - $size[1];
 
         $this->_page_number = $this->_page_count = 1;
-        $this->pageCallbacks = [];
 
         $this->_pages = [$this->_pdf->getFirstPageId()];
     }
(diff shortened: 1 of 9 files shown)


Mellthas:
Addendum to number 2: The event could in the future be restricted to only apply the page script to the pages actually rendered, with the page numbers adapted appropriately. This could make it easier to render multiple documents onto the same canvas.

bsweeney:
I think I'm OK with these changes, but let me review a bit more. We'll want to create a migration guide to reference when questions come up.

PaolaRuby:
Hi, is https://github.com/dompdf/dompdf/issues/2919 posible?

Mellthas:
> We'll want to create a migration guide to reference when questions come up.

Definitely. Regarding the `page_script` change, we should check the code samples on the wiki; they should be easy to adapt so that they work both with old and new versions.

Mellthas:
I am going to update the Usage page on the wiki.

Mellthas:
> I am going to update the Usage page on the wiki.

Done. I can also add a section to the migration guide, or do you want to do that @bsweeney?

bsweeney:
Go ahead, I hadn't really prepared anything for it yet.

Mellthas:
Alright, will do that tomorrow.

Mellthas:
Added a section to the [Migration Guide](https://github.com/dompdf/dompdf/wiki/Migration-Guide).

PR #2921, opened 2022-06-14
Various doc block cleanups for `Canvas` and `Cpdf`

Mellthas:
Includes dropping the `$np` parameter from the `Cpdf::polygon` method. Similar to the GD `imagepolygon` function, where the corresponding parameter has been deprecated, the parameter is dubious and expected to be `count($p) / 2` usually. Passing more will try to access undefined array keys and passing less is essentially a built-in `array_slice`; which can easily be applied manually if needed.

https://www.php.net/manual/en/function.imagepolygon.php

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -3871,11 +3871,11 @@ EOT;
     /**
      * draw a line from one set of coordinates to another
      *
-     * @param $x1
-     * @param $y1
-     * @param $x2
-     * @param $y2
-     * @param bool $stroke
+     * @param float $x1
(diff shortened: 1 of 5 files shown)


bsweeney:
Was going to suggest we also update the copy in php-svg-lib, but it looks like that library doesn't use this method for drawing polygons so this change should not cause a compatibility issue there. With the next major release we should go ahead and move forward with moving the Cpdf library to it's own project so we're not maintaining two copies.

bsweeney (approved):

PR #2949, opened 2022-07-10
Fix rendering performance with large number of nested inline elements

Mellthas:
The algorithm for determining the final size of inline frames had exponential complexity in the number of nested inline frames due to calling itself recursively twice for every level.

Addresses #639


Mellthas:
The sample https://github.com/dompdf/dompdf/issues/639#issuecomment-1179180722 now renders as fast as one would expect.

bsweeney (approved):

PR #2979, opened 2022-08-06
Add support for the calc CSS function

samuferenc:
Basic calc CSS function
It's can handle addition (+), subtraction (-), multiplication (*), division (/), and parentheses

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -1016,6 +1016,11 @@ class Style
 
         $font_size = $font_size ?? $this->__get("font_size");
 
+        $calcPattern = "/^calc\((.*)?\)$/i";
+        if (preg_match($calcPattern, $l, $calcMatches)) {
+            return $this->evaluate_func_calc($this->parse_calc($calcMatches[1]), $ref_size, $font_size);
+        }
+
         $key = "$l/$ref_size/$font_size";
 
         if (\array_key_exists($key, $cache)) {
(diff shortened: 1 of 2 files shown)


tests/Css/StyleTest.php
```diff
@@ -39,6 +39,29 @@ public function testLengthInPt(string $length, ?float $ref_size, $expected): voi
         $this->assertSame($expected, $result);
     }
 
+    public function calcLengthInPtProvider(): array
```
Mellthas, line 42: Test cases with inner parentheses would be nice to have.

tests/Css/StyleTest.php
```diff
@@ -39,6 +39,29 @@ public function testLengthInPt(string $length, ?float $ref_size, $expected): voi
...
+          ["calc(50% - 1pt)", 200, 99.0],
+          ["calc(100)", null, 100.0],
+          ["calc(100% / 3)", 100, 33.333333333333336],
+        ];
+    }
+
+    /**
+     * @dataProvider calcLengthInPtProvider
+     */
+    public function testCalcLengthInPt(string $length, ?float $ref_size, $expected): void
```
Mellthas, line 55: Since the method is identical to `testCalcLengthInPt`, I would merge both, plus the providers; just separate the `calc` tests with a blank line in the data provider for clarity.

Mellthas (changes requested):
Great work here! I have added comments for details that I think should be addressed. Can you also rebase the branch?

One more issue: A declaration of `width: calc(100% - 100pt);` will fail now, because of the assumptions taken in `compute_length_percentage_positive`. I can handle the needed modifications, though, either after merge, or I will submit a PR to your branch later.

bsweeney:
FYI I'm going to bump this up to the next release (which will likely be either 2.1.0 or 3.0 though it is currently tagged as 2.0.4).

src/Css/Style.php
```diff
@@ -3608,6 +3612,85 @@ protected function parse_image_resolution(string $val): ?string
...
+     * @return array
+     */
+    public function calcParse(string $expr): array
+    {
+        $precedence = ['*' => 3, '/' => 3, '+' => 2, '-' => 2];
+        $ops = ['*', '/', '+', '-'];
+
+        $opStack = [];
+        $queue = [];
+        $parts = explode(' ', trim(str_replace(['(', ')', '  '], [' ( ', ' ) ', ' '], $expr)));
```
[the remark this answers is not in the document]
bsweeney: The logic was updated to normalize the expression by ensuring whitespace around all operators and collapsing multiple spaces.

src/Css/Style.php
```diff
@@ -999,6 +999,10 @@ protected function single_length_in_pt(string $l, float $ref_size = 0, ?float $f
 
         $font_size = $font_size ?? $this->__get("font_size");
 
+        if (strncmp($l, 'calc(', 5) == 0) {
+            return $this->_compute_calc($this->calcParse(substr($l, 5, -1)), $ref_size, $font_size);
```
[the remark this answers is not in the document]
bsweeney: `single_length_in_pt` was left in it's current location, but if the evaluation fails at any point null is returned. Failure can be caused by:
- no values to parse
- lack of operator for two non-operator values (e.g. `calc(1 2 3)`).
- failed length calculation

src/Css/Style.php
```diff
@@ -999,6 +999,10 @@ protected function single_length_in_pt(string $l, float $ref_size = 0, ?float $f
 
         $font_size = $font_size ?? $this->__get("font_size");
 
+        if (strncmp($l, 'calc(', 5) == 0) {
```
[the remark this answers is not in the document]
bsweeney: Switched to a generic regex to capture the calc statement.

src/Css/Style.php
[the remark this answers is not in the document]
bsweeney: There is a lack of validation in the parsing method, but since the logic was written for this specific use case (not for general use) I think I'm OK with that. Errors in the `calc` structure will be caught when attempting to evaluate. We shouldn't see much impact on rendering unless the CSS is excessively `calc`'d.

src/Css/Style.php
```diff
@@ -1083,6 +1088,109 @@ protected function single_length_in_pt(string $l, float $ref_size = 0, ?float $f
...
+        if ($expr === '') {
+            return [];
+        }
+
+        $parts = explode(' ', $expr);
+
+        foreach ($parts as $part) {
+            if ($part === '(') {
+                $opStack[] = $part;
+            } elseif ($part == ')') {
```
Mellthas, line 1119: Strict comparison here for consistency.

src/Css/Style.php
```diff
@@ -1083,6 +1088,109 @@ protected function single_length_in_pt(string $l, float $ref_size = 0, ?float $f
...
+                while (count($opStack) > 0 && end($opStack) !== '(') {
+                    $queue[] = array_pop($opStack);
+                }
+                if (end($opStack) == '(') {
```
Mellthas, line 1123: Same here.

src/Css/Style.php
```diff
@@ -1083,6 +1088,109 @@ protected function single_length_in_pt(string $l, float $ref_size = 0, ?float $f
...
+                    array_pop($opStack);
+                }
+            } elseif (array_key_exists($part, $precedence)) {
+                while (count($opStack) > 0 && end($opStack) != '(' && $precedence[end($opStack)] >= $precedence[$part]) {
```
Mellthas, line 1127: Same here.

src/Css/Style.php
```diff
@@ -1083,6 +1088,109 @@ protected function single_length_in_pt(string $l, float $ref_size = 0, ?float $f
...
+        if (count($rpn) === 0) {
+            return null;
+        }
+
+        $ops = ['*', '/', '+', '-'];
+
+        $stack = [];
+
+        foreach ($rpn as $part) {
+            if (in_array($part, $ops)) {
```
Mellthas, line 1161: Add `true` as third parameter for strict comparison.

src/Css/Style.php
```diff
@@ -1083,6 +1088,109 @@ protected function single_length_in_pt(string $l, float $ref_size = 0, ?float $f
...
+
+        return $queue;
+    }
+
+    /**
+     * Reverse Polish Notation
+     * @param array $rpn
+     * @param float $ref_size
+     * @param float|null $font_size
+     * @return float
```
Mellthas, line 1148: Return value is `float|null` now.

src/Css/Style.php
```diff
@@ -1083,6 +1088,109 @@ protected function single_length_in_pt(string $l, float $ref_size = 0, ?float $f
...
+    private function parse_calc(string $expr): array
+    {
+        if (substr_count($expr, '(') !== substr_count($expr, ')')) {
+            return [];
+        }
+
+        $precedence = ['*' => 3, '/' => 3, '+' => 2, '-' => 2];
+
+        $opStack = [];
+        $queue = [];
```
Mellthas, line 1105: Would move those to after the expression check in lines 1110–1112 for clarity.

Mellthas (changes requested):
Okay, only some minor nits left, then I think we are good to go here. I will post a follow-up PR addressing the issue I mentioned above.

Can you also squash your commits (and force push) when you are done, please?

src/Css/Style.php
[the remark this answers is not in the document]
Mellthas: Should be fine, yes, since both methods are private and used in conjunction.

Mellthas:
Thank you, and sorry for the long wait. Opened #3298 as the promised follow-up.

PR #2992, opened 2022-08-24
Rerolling old PR for 2098

dealancer:
See https://github.com/dompdf/dompdf/issues/2098

### DIFF
--- a/src/Helpers.php
+++ b/src/Helpers.php
@@ -105,7 +105,7 @@ class Helpers
             //drive: followed by a relative path would be a drive specific default folder.
             //not known in php app code, treat as abs path
             //($url[1] !== ':' || ($url[2]!=='\\' && $url[2]!=='/'))
-            if ($url[0] !== '/' && (strtoupper(substr(PHP_OS, 0, 3)) !== 'WIN' || (mb_strlen($url) > 1 && $url[0] !== '\\' && $url[1] !== ':'))) {
+            if (mb_strtolower($protocol) == "file://" || $url[0] !== '/' && (strtoupper(substr(PHP_OS, 0, 3)) !== 'WIN' || (mb_strlen($url) > 1 && $url[0] !== '\\' && $url[1] !== ':'))) {
                 // For rel path and local access we ignore the host, and run the path through realpath()
                 $ret .= realpath($base_path) . '/';
             }


bsweeney:
Same as #2099? As I [commented](https://github.com/dompdf/dompdf/pull/2099#issuecomment-745753639) there:

> I don't think this is the right change to make. It looks like what you're trying to do is force Dompdf to adhere to the document root of your application. But that's not the intention of the base parameters. They're intended to act similar to the base href of an HTML document.
> 
> The current logic already supports file-protocol for the base protocol or URL. Current logic prepends the base path if the URL isn't absolute, which is the expected behavior. With this change when the base protocol is "file://" Dompdf would always prepend the base path. We only want to prepend the base path if the url is relative.

[I have a solution in mind](https://github.com/dompdf/dompdf/issues/2098#issuecomment-1111087666) though I'm not yet sure when it'll be implemented.

dealancer:
@bsweeney, thanks for explanation, it makes sense. looks like I am doing something wrong and I appreciate your help in advance.

I need to replace relative links to CSS files to absolute path with a file:// schema.

The original HTML looks like:

```
<link rel="stylesheet" media="all" href="/modules/contrib/entity_print/css/entity-print.css?rh52a6">
```

The goal is to replace it to 

```
<link rel="stylesheet" media="all" href="/var/www/docroot/modules/contrib/entity_print/css/entity-print.css">
```

Would the existing codebase allow us do so? I can call theses methods:

```
$this->dompd->setProtocol('file://');
$this->dompdf->setBaseHost($base_host);
$this->dompdfOptions->setChroot($chroot);
```

What `$base_host` and `$chroot` should be? I am not sure what `$chroot` does and if we ever need to set it.

The problem may be in `?rh52a6` added to the end of the file, but it looks like a separate issue.

bsweeney:
Regardless of the resource context (web or local filesystem) a path that starts with a "/" indicates that the resource is not relative to the rendered file but to the root of the context. Dompdf isn't designed to transform the URL in the way you want. The base URL components (protocol, host, path) just provide hints on where to look for resources when the URL is relative (i.e., doesn't start with a "/").

The "?rh52a6" would definitely be problematic when you're accessing the stylesheet through the local filesystem. That's indicative that you're using a web layout. The typical mechanism for rendering the PDF using local resources is to use an alternate layout rather than the layout used for serving up web content.

PR #3006, opened 2022-09-14
Issue 3005: Fix null value in FrameReflower/Text

KudrautsauSiarhei:


Mellthas (approved):
Using an early return here is a bit clearer I think; the `$block` check below can be removed now. There is also a similar potential crash in the `ListBullet` class. I’ll handle these points in a followup PR.

Issue #3019, opened 2022-10-05
Uninitialized string offset 22


jochemgruter:
Got the error `Uninitialized string offset 22 ` on file vendor/dompdf/dompdf/src/Css/Stylesheet.php:772

When i add the following code above `switch ($tok[$j]) { ` then the problem is solved.
                    
      if (!isset($tok[$j])) {
              break;
      }

jochemgruter:
[app.css.txt](https://github.com/dompdf/dompdf/files/9718885/app.css.txt)

This is the css file I use, to reproduce.

bsweeney:
Looks like there's an issue parsing CSS that uses variable-like syntax in the class name. The following from your stylesheet triggers the error.

```
.w-\[var\(--sidebar-width\)\]{width:var(--sidebar-width)}
```

That doesn't appear to be an actual CSS variable but a CSS variable used as a class name? I dunno it just looks weird.

bsweeney:
FYI, Dompdf doesn't yet support CSS variables (#1872), which your stylesheet appears to use heavily. There's a lot to parse so I didn't look to see if there's sufficient fallback styling.

PR #3021, opened 2022-10-05
Fix error Uninitialized string offset 22

jochemgruter:
Related to issue #3019

### DIFF
--- a/src/Css/Stylesheet.php
+++ b/src/Css/Stylesheet.php
@@ -769,6 +769,10 @@ class Stylesheet
                         $attr .= $tok[$j++];
                     }
 
+                    if (!isset($tok[$j])){
+                        break;
+                    }
+
                     switch ($tok[$j]) {
 
                         case "~":


src/Css/Stylesheet.php
```diff
@@ -769,6 +769,10 @@ private function _css_selector_to_xpath(string $selector, bool $first_pass = fal
                         $attr .= $tok[$j++];
                     }
 
+                    if (!isset($tok[$j])){
+                        break;
```
Mellthas, line 773: Note that breaking here changes the logic, the existing `$query` remains; so e.g. the invalid selector `.w[href` would be treated as `.w` now instead of `.w[href]` as before. To just remove the warning, retaining the current logic, one could use `$tok[$j] = null;` instead.

It would be better to drop the entire selector instead. I am working on some small updates here to do just that.

Mellthas: See #3057

Mellthas:
Resolved via #3057 now.

PR #3034, opened 2022-10-24
Handle `null` block in reflower classes

Mellthas:
Follow-up to #3006, see https://github.com/dompdf/dompdf/pull/3006#pullrequestreview-1153368957. Make sure to not throw when no block is passed to the reflow methods.

### DIFF
--- a/src/FrameReflower/ListBullet.php
+++ b/src/FrameReflower/ListBullet.php
@@ -31,6 +31,10 @@ class ListBullet extends AbstractFrameReflower
      */
     function reflow(BlockFrameDecorator $block = null)
     {
+        if ($block === null) {
+            return;
+        }
+
         /** @var ListBulletFrameDecorator */
         $frame = $this->_frame;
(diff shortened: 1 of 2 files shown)


Mellthas:
Test case for a crash in `FrameReflower\ListBullet` that is fixed by this PR:

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
body {
    display: inline !important;
}

span {
    display: -dompdf-list-bullet;
}
</style>
</head>

<body><span>list-bullet</span></body>

</html>
```

bsweeney:
Suggesting this instead of 3006? I agree that early return is preferable to conditional handling later.

bsweeney (approved):

Mellthas:
> Suggesting this instead of 3006?

As far as the end result is concerned, yes. I have merged #3006 first now to retain the attribution, as the early return is functionally equivalent.

PR #3037, opened 2022-10-24
Fix logic for skipping white space between block-level elements

### LINKED ISSUE #2991: span style margin-left not working
`<span style="margin-left:50px;"> </span>`  was working in earlier version but not working now.

Mellthas:
The previous logic was too eager. Account for cases where the text has box decoration applied via a parent inline element.

Fixes #2991

### DIFF
--- a/src/FrameReflower/Text.php
+++ b/src/FrameReflower/Text.php
@@ -278,12 +278,10 @@ class Text extends AbstractFrameReflower
             $text = ltrim($text, " ");
         }
 
-        // Exclude wrapped white space. This handles white space between block
-        // elements in case white space is collapsed
         if ($text === "") {
             $frame->set_text("");
             $style->set_used("width", 0.0);
-            return null;
+            return false;
         }
 
         // Determine the next line break
@@ -408,6 +406,12 @@ class Text extends AbstractFrameReflower
         }
 
         $frame->position();
+
+        // Skip wrapped white space between block-level elements in case white
+        // space is collapsed
+        if ($frame->get_text() === "" && $frame->get_margin_width() === 0.0) {
+            return;
+        }
 
         $line = $block->add_frame_to_line($frame);
         $trimmed = trim($frame->get_text());


Mellthas:
Here is a test document demonstrating several different cases:

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
    size: 400pt 300pt;
    margin: 50pt;
}

.box {
    border: 1pt solid orange;
}

.break {
    page-break-before: always;
}

.ws {
    background: red;
}

.deco-ws {
    background: lightblue;
    border-left: 3pt solid blue;
}
</style>
</head>

<body>
    <div class="box">A<span class="ws"> </span>B</div>
    <div class="box"><span class="ws"> </span>C</div>
    <div class="box">D<span class="deco-ws"> </span>E</div>
    <div class="box"><span class="deco-ws"> </span>F</div>

    <div class="box break">A</div><span class="ws"> </span><div class="box">B</div>
    <span class="ws"> </span>
    <div class="box">C</div>
    <span class="ws"> </span>
    <span class="ws"> </span>
    <span class="ws"> </span>
    <div class="box">D</div>
    <span class="deco-ws"> </span>
    <div class="box">E</div>
</body>

</html>
```
</details>

[white-space-before.pdf](https://github.com/dompdf/dompdf/files/9855942/white-space-before.pdf)
[white-space-after.pdf](https://github.com/dompdf/dompdf/files/9855941/white-space-after.pdf)

bsweeney:
This is a definite improvement, though we're still optimizing elements out of the flow when they have no content. For example, removing the space from inside the deco-ws class:

```
<div class="box">D<span class="deco-ws"></span>E</div>
```

Mellthas:
Yes, that is due to the fact that empty inline frames are neither drawn, nor properly considered during layout:

https://github.com/dompdf/dompdf/blob/d4dccd308f713b00d47615fcc8108c35fc6cb28c/src/Renderer/Inline.php#L21-L23

Since there are no text-frame children, the logic here does not apply. The interaction between text and inline frames is still pretty sub-optimal.

My intention here is to fix the regression in rendering of the test case compared to version 1.1.1 and earlier. The (non-) rendering of empty inline frame hasn’t changed since then.

bsweeney (approved):

PR #3038, opened 2022-10-24
Render white-space-only text

Mellthas:
This ensures sure that text decoration is applied.

### DIFF
--- a/src/Renderer/Text.php
+++ b/src/Renderer/Text.php
@@ -45,7 +45,7 @@ class Text extends AbstractRenderer
         $style = $frame->get_style();
         $text = $frame->get_text();
 
-        if (trim($text) === "") {
+        if ($text === "") {
             return;
         }
 


Mellthas:
See the following test case, which would render with a gap in the underline before:

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
    size: 400pt 300pt;
    margin: 50pt;
}

html {
    font-size: 30pt;
}

.underline {
    text-decoration: underline;
    background-color: lightblue;
}
</style>
</head>

<body>
    <div><span class="underline">Some</span><span class="underline"> </span><span class="underline">underlined text</span></div>
    <div class="underline">Some<span> </span>underlined text</div>
</body>

</html>
```
</details>

[underline-before.pdf](https://github.com/dompdf/dompdf/files/9855984/underline-before.pdf)
[underline-after.pdf](https://github.com/dompdf/dompdf/files/9855983/underline-after.pdf)

bsweeney (approved):

PR #3041, opened 2022-10-25
Update Dompdf.php

renatogap:
Included paper orientation that was not receveid. For example: when setting "landscape" it just take the default.

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -706,7 +706,7 @@ class Dompdf
         // Set paper size if defined via CSS
         if (is_array($basePageStyle->size)) {
             [$width, $height] = $basePageStyle->size;
-            $this->setPaper([0, 0, $width, $height]);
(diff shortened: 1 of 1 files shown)


bsweeney (changes requested):
I think this isn't quite right. The orientation is used to [swap the height/width](https://github.com/dompdf/dompdf/blob/v2.0.1/src/Adapter/CPDF.php#L167-L169) when sizing the page in the canvas adapter. However, when parsing the `@page` size declaration the [orientation has already been taken into account](https://github.com/dompdf/dompdf/blob/v2.0.1/src/Css/Style.php#L3428-L3432).

If you have a sample that demonstrates the issue you're attempting to address that would help clarify the problem.

Mellthas:
This is indeed working as intended, as @bsweeney explained. Maybe it might be good to add a small comment to avoid any future confusion, as the difference in how orientation is treated is not ideal. If you think something should work differently here, please file a new issue.

Issue #3055, opened 2022-11-05
PHP 8.1 deprecated  strtolower()


Schneidoa:
I receive the following 2 warnings.

PHP: 8.1.10
DOMPDF:  v2.0.1

`strtolower(): Passing null to parameter #1 ($string) of type string is deprecated  Line: 992 File: vendor\dompdf\dompdf\src\Dompdf.php`

`strtolower(): Passing null to parameter #1 ($string) of type string is deprecated  Line: 167 File: vendor\dompdf\dompdf\src\Adapter\CPDF.php`

bsweeney:
It looks like somewhere along the way the paper orientation is being set to `null`. Can you provide a sample of the PHP usage of Dompdf as well as the relevant HTML+CSS?

Schneidoa:
Thank you for the quick reply and the hint,

I think I found the problem in my case. The orientation at setPaper() is null, but it works with php 8.0



```         $output = $this->smarty->fetch($this->templateFile);

            $dompdf = new Dompdf();
            $dompdf->setPaper('A4', $this->model->getTemplateVar('FORMAT'));
            $dompdf->loadHtml($output);
            $dompdf->render();
            $dompdf->stream('document.pdf');
```

The value of $output:
```
<style type="text/css">
   <!--
      @page project {
      	size: A4 portrait;
      	margin: 2cm;
      }
      
      .projectPage {
      	page: project;
      	page-break-after: always;
      	font-size: 10px;
      	font-family: "Lucida Console", "Courier New", Arial, sans-serif;
      }
      -->
</style>
<div class="projectPage">
05.11.2022
   <div style="text-align: center; font-size: 14px;">P R O J E K T B L A T T</div>
   <div style="text-align: center;">Test - Projekt bla</div>
   Projektstatus: E
   <table width="90%" style="clear: both">
      <tr>
         <td><b>Projektnr.</b></td>
         <td>: 22/008</td>
         <td><b>Termin</b></td>
         <td>: </td>
      </tr>
      <tr>
         <td><b>Sparte</b></td>
         <td>: 213</td>
         <td><b>Beginn</b></td>
         <td>: </td>
      </tr>
      <tr>
         <td><b>Ort</b></td>
         <td>: Teststadt</td>
         <td><b>Ende</b></td>
         <td>: </td>
      </tr>
      <tr>
         <td><b>Maßnahme</b></td>
         <td>: bla</td>
         <td><b>Vertrag</b></td>
         <td>: </td>
[...]

PR #3060, opened 2022-11-07
Add type declaration to paper-orientation parameters

Mellthas:
Addresses #3055


bsweeney:
Too bad we're not yet at PHP 8.0 as the minimum so we could use a [union type](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.composite.union) on the paper argument.

bsweeney (approved):

PR #3074, opened 2022-11-30
test on PHP 8.2 + imagick + gmagick

IonBazan:
Adds PHP 8.2 to the build matrix and tests against `imagick` and `gmagick` extension.

I used `include` syntax so there is only one job for these extensions which should be enough.

`gmagick` cannot be installed on 8.2 so sticked to 8.1.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,17 +14,25 @@ jobs:
     strategy:
       max-parallel: 12
       matrix:
-        php: ['7.1', '7.2', '7.3', '7.4', '8.0', '8.1']
+        php: ['7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2']
         package-release: [dist]
+        extensions: ['gd']
+        include:
+          - php: '8.1'
+            package-release: 'dist'
+            extensions: 'gmagick'
+          - php: '8.2'
+            package-release: 'dist'
+            extensions: 'imagick'
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


bsweeney (changes requested):
Can we test against variations in the extension configuration? Ideally we would run the test with and without the imagick extension.

IonBazan:
@bsweeney this is what's done now - we have PHP 7.1-8.2 with `gd` only, PHP 8.2 with `imagick` and PHP 8.1 with `gmagick`:
image


If you want, we can make it run in all combinations, but that would produce 20 CI jobs which I don't think is necessary as it's enough to check them with a single PHP version.

bsweeney:
Ah, sorry, my bad. Still getting my feet underneath me with regards to github actions. I meant to check myself on that an forgot.

bsweeney (approved):

bsweeney:
As far as checking against more versions I could go either way on that question, but it does seem excessive. Though eventually that may be the scenario we find ourselves in as more PHP versions are supported. Worth considering if we continually roll that check forward, maybe just use it on the last minor release (e.g., 7.4 and 8.2)?

IonBazan:
https://github.com/vitoc/gmagick seems a bit abandoned to me but let's keep it in the CI as this library still supports it. We can drop it after graceful deprecation and sunset cycle but it all depends on how many people use it with dompdf.

In the meantime, I think it's worth using this pipeline against #2510 to see if it still passes when Imagick is present (it did, when I run it locally).

PR #3075, opened 2022-12-02
Fix AbstractFrameDecorator::reset_counter

attisan:
When using counters on strongly nested html dom, resetting the counter using `counter-reset` "sometimes" won't work.

### DIFF
--- a/src/FrameDecorator/AbstractFrameDecorator.php
+++ b/src/FrameDecorator/AbstractFrameDecorator.php
@@ -776,7 +776,13 @@ abstract class AbstractFrameDecorator extends Frame
      */
     public function reset_counter(string $id = self::DEFAULT_COUNTER, int $value = 0): void
     {
-        $this->get_parent()->_counters[$id] = $value;
+        $counter_frame = $this->lookup_counter_frame($id);
+
+        if ($counter_frame) {
+            $counter_frame->_counters[$id] = $value;
+        } else {
+            $this->get_parent()->_counters[$id] = $value;
+        }
     }
 
     /**


bsweeney:
Using consistent logic with `increment_counter` makes sense. Do you have sample HTML?

Mellthas:
Your change unfortunately breaks nested counters completely. See the following sample, the items are supposed to be numbered

* 1
* 1.1
* 1.1.1

not _1, 2, 3_.

```html
<!DOCTYPE html>
<html>
<head>
<meta charset="UTF-8">
<style>
ul {
    counter-reset: li;
}

li {
    counter-increment: li;
}

li::before {
    content: counters(li, ".") " ";
}
</style>
</head>
<body>
    <ul>
        <li><span>Item 1</span>
            <ul>
                <li><span>Item 2</span>
                    <ul>
                        <li><span>Item 3</span></li>
                    </ul>
                </li>
            </ul>
        </li>
    </ul>
</body>
</html>
```

Mellthas:
While reviewing this, I realized that automatic counter reset (`counter-increment` or `content: counter()` without explicit `counter-reset`) is not handled properly, though. This should be addressed by #3163.

The other thing I came across is a change in how `counter-reset` is supposed to work when used on siblings. Note how the following sample produces different counter values in recent Firefox versions compared to other browsers and dompdf:

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>
<head>
<meta charset="UTF-8">
<style>
body {
    counter-reset: c;
}

li {
    counter-increment: c;
}

li::before {
    content: counters(c, ".") " ";
}

.reset {
    counter-reset: c;
}
</style>
</head>
<body>
    <ul>
        <li>Item 1</li>
        <li class="reset">Item 2</li>
        <li>Item 3</li>
        <li class="reset">Item 4</li>
        <li>Item 5</li>
    </ul>

    <ul>
        <li>Item 6</li>
        <li class="reset">Item 7</li>
        <li>Item 8</li>
    </ul>
</body>
</html>
```
</details>


Maybe this is what you are encountering?

Other than that, the current implementation seems to be fine wrt. `counter-reset`. Since you didn’t provide any further information, we can only guess what you are trying to fix. If you are still encountering unexpected behavior with #3163 applied, please file a new issue and provide a sample document if possible.

PR #3100, opened 2023-01-05
Fixed bug where svg polylines get automatically closed

mck89:
This fixes:

- https://github.com/dompdf/php-svg-lib/issues/101
- https://github.com/dompdf/php-svg-lib/issues/99

The problem is that the cpdf version included in dompdf does not contain the `$close` parameter like the version included in php-svg-lib: https://github.com/dompdf/php-svg-lib/blob/master/src/Svg/Surface/CPdf.php#L4177-L4190

Before:
[before.pdf](https://github.com/dompdf/dompdf/files/10352171/before.pdf)

After:
[after.pdf](https://github.com/dompdf/dompdf/files/10352172/after.pdf)

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -4220,9 +4220,9 @@ EOT;
         $this->addContent(sprintf("\n%.3F %.3F %.3F %.3F re", $x1, $y1, $width, $height));
     }
 
-    function stroke()
+    function stroke(bool $close = false)
     {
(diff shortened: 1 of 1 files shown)


bsweeney:
Thank you for that. I totally forgot to make sure the two are synchronized 😞.

In a near-term release I'm going to move Cpdf into it's own project so that we don't have to maintain the code base in two locations.

PR #3112, opened 2023-01-19
tinyfix Uninitialized string offset for minmax() attr

vobar:
In stack laravel/vite - after bulid application assets i try use app-xxxxxxxx.css file and get error "Uninitialized string offset"
in my case - minmax() function break the PDF processing, this fix can avoid this and can use this in future for minmax aplying in package

sorry for my english

### DIFF
--- a/src/Css/Stylesheet.php
+++ b/src/Css/Stylesheet.php
@@ -713,7 +713,7 @@ class Stylesheet
                         $query .= "/*";
                     }
 
-                    $attr_delimiters = ["=", "]", "~", "|", "$", "^", "*"];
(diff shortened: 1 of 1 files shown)


bsweeney:
I am unable to reproduce. Can you provide a sample document that exhibits the issue?

Mellthas:
We have fixed several instances of uninitialized string offset warnings with dompdf 2.0.2, see issue #3019 for reference. Since that version was released after this pull request, you were probably encountering the warning with an older version? 

Anyways, the proposed fix looks odd, it could only really fix such an error by accident: The `minmax` function is completely unrelated to attribute-selector parsing, which you are modifying here.

I’m going to close this, but if you are still encountering the warning with the latest release, please file a new issue.

PR #3115, opened 2023-01-23
Try to create config directories if non existant

Blaklis:
The patch will allow dompdf to try creating tmpDir, fontDir and fontCache if they do not exist on the filesystem - and will trigger a warning if they can't be created.

Even if the Options.php file mention that pathes should exist, a number of people will set the options and won't even see that the folders weren't created.

Trying to create them by default will : 
- Avoid bugs / lack of performances when directories are created
- Will warn the user that his configurations are not suitable if necessary.

Later, it might also be possible to change the trigger_error to an exception, but right now, it would be a BC break.

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -848,6 +848,9 @@ class Options
      */
     public function setFontCache($fontCache)
     {
+        if (!is_dir($fontCache) && !mkdir($fontCache)) {
+            trigger_error("Unable to create fontCache - a valid value should be specified.", E_USER_WARNING);
+        }
         $this->fontCache = $fontCache;
         return $this;
     }
@@ -866,6 +869,9 @@ class Options
      */
     public function setFontDir($fontDir)
     {
+        if (!is_dir($fontDir) && !mkdir($fontDir)) {
+            trigger_error("Unable to create fontDir - a valid value should be specified.", E_USER_WARNING);
+        }
         $this->fontDir = $fontDir;
         return $this;
     }
@@ -1053,6 +1059,9 @@ class Options
      */
     public function setTempDir($tempDir)
     {
+        if (!is_dir($tempDir) && !mkdir($tempDir)) {
+            trigger_error("Unable to create tempDir - a valid value should be specified.", E_USER_WARNING);
+        }
         $this->tempDir = $tempDir;
         return $this;
     }


bsweeney:
This seems reasonable.

src/Options.php
```diff
@@ -848,6 +848,9 @@ public function getDpi()
      */
     public function setFontCache($fontCache)
     {
+        if (!is_dir($fontCache) && !mkdir($fontCache)) {
```
bsweeney, line 851: Should we make the directory recursively (e.g., `mkdir($dir, 0777, true)`)?

bsweeney (changes requested):
Thinking on this again, I wonder if we just require that the user create the necessary directories prior to setting them in Dompdf. The main reason being that if we really want to do this we should probably allow the user to specify the permissions, which complicates things a bit. We could, instead of making the directory, just throw an exception if the specified directory is invalid so that the user knows they forgot to do something.

I'm still open to this change, just trying to think through potential unintended consequences.

Blaklis:
> Thinking on this again, I wonder if we just require that the user create the necessary directories prior to setting them in Dompdf. The main reason being that if we really want to do this we should probably allow the user to specify the permissions, which complicates things a bit. We could, instead of making the directory, just throw an exception if the specified directory is invalid so that the user knows they forgot to do something.
> 
> I'm still open to this change, just trying to think through potential unintended consequences.

Yep, I agree - however, I didn't want to have a BC break by preventing people to run if the folder doesn't exist. Might it be a simple warning message for the next version, and a break on the next one?

src/Options.php
Blaklis: Yep, that would be definitely better, even if the 777 permission might be problematic somehow

src/Options.php
```diff
@@ -848,6 +848,9 @@ public function getDpi()
...
+            trigger_error("Unable to create fontCache - a valid value should be specified.", E_USER_WARNING);
```
bsweeney, line 852: Thinking Dompdf should return without setting the option if it's unable to create the directory.

bsweeney: Sometimes it's hard to balance between just making it work and asking developers to do their job 😅. You're probably right that a lot of users will miss the warnings. Is your suggestion to not make the directory but add the warning, then in a future major release add the mkdir functionality?

PR #3155, opened 2023-03-15
Added automatic character range detection that determines font

robvanderlee:


bsweeney (changes requested):
I think these suggested changes should help with performance. The changes I made are:
- Always perform font mapping. This removes duplicate processing of text in those cases where the full block of text was not supported by the font. This imparts a slight performance hit in the case where the full text is supported by the font, but it's negligible (with my limited testing).
- Only process as much text as necessary when mapping the font. Since in this particular use case we only want the first block of mapped text there's no need to map beyond that.
- Commented out the building the string for each mapped block. That adds overhead for something we don't currently use. I think it could still be worthwhile, just not in this particular use case. Maybe we can optionally generated that content based on an additional argument?

Worth noting that the additional frame count resulting from the font mapping still impacts performance, but less so than before.

It may be possible to squeeze a bit more performance out of this by processing the string one character at a time instead of splitting it at the start. However, I think this would only improve performance on extremely long strings, which seems like an outlier.

src/FrameReflower/Text.php
```diff
@@ -393,6 +396,21 @@ function reflow(BlockFrameDecorator $block = null)
...
+                $frame->split_text($breakpoints[0]-1 + mb_strlen($charMapping[$breakpoints[0]]["char"]));
+                $text = $this->pre_process_text($frame->get_text());
+            }
+
+            $mapped_font = key_exists(0, $breakpoints) ? $charMapping[$breakpoints[0]]["font"] : false;
+            if ($mapped_font !== false) {
+                $style->set_used("font_family", $mapped_font);
+            }
+        }
+
```
bsweeney, line 413:
```suggestion
        // map text to fonts based on supported Unicode range
        $families = array_map(function ($value) { return trim($value, " '\""); }, explode(",", $style->get_specified("font_family")));
        $charMapping = $this->getFontMetrics()->getTextFamilies($families, $text, 1);
        if (isset($charMapping[0])) {
            if ($charMapping[0]["length"] !== 0) {
                $frame->split_text($charMapping[0]["length"]);
                $text = $this->pre_process_text($frame->get_text());
            }
            $mapped_font = $charMapping[0]["font"];
            if ($mapped_font !== null) {
                $style->set_used("font_family", $mapped_font);
            }
        }

```

src/FontMetrics.php
```diff
@@ -326,6 +339,133 @@ public function getTextWidth(string $text, $font, float $size, float $wordSpacin
...
+            if ($_subtable["platformID"] == 0 || $_subtable["platformID"] == 3 && $_subtable["platformSpecificID"] == 1) {
+                $subtable = $_subtable;
+                break;
+            }
+        }
+
+        $this->unicodeCharMapTables[$font] = $subtable;
+        return $this->unicodeCharMapTables[$font];
+    }
+
```
bsweeney, line 415: Add the following method in support of the changes to getTextFamilies:

```suggestion

    private function fontSupportsGlyph(int $codepoint, string $font): bool
    {
        if (!key_exists($font, $this->unicodeCharMapTables) || $codepoint === 0xFFFF) {
            return false;
        }

        $fontRange = $this->unicodeCharMapTables[$font];
        $glyphIndexArray = $fontRange["glyphIndexArray"];

        return array_key_exists($codepoint, $glyphIndexArray);
    }

```

src/FontMetrics.php
```diff
@@ -326,6 +339,133 @@ public function getTextWidth(string $text, $font, float $size, float $wordSpacin
...
+                unset($chars[$index]);
+            }
+
+            if (key_exists($currentIndex, $chars) && !is_array($chars[$currentIndex])) {
+                $chars[$currentIndex] = ['font' => false, 'char' => $char];
+            }
+        }
+
+        return $chars;
+    }
```
bsweeney, line 462:
```suggestion
    /**
     * Maps string
     * @param Style  $style
     * @param string $text
     * @return array
     */
    public function getTextFamilies(array $fontFamilies, string $text, int $count = -1): array
    {
        $char_mapping = [];
        $fonts = [];

        foreach ($fontFamilies as $family) {
            $fonts[$family] = $font = $this->getFont($family);

            if (!key_exists($font, $this->unicodeCharMapTables)) {
                $this->loadunicodeCharMapTables($font);
            }
        }

        $char_array = mb_str_split($text, 1, "UTF-8");
        $start_index = 0;
        $char_index = -1;
        while (isset($char_array[++$char_index])) {
            $char = $char_array[$char_index];
            if (function_exists("mb_ord")) { //PHP 7.2+
                $char_code = mb_ord($char, "UTF-8");
            } else {
                $char_code = $this->getCanvas()->get_cpdf()->utf8toCodePointsArray($char);
            }

            $mapped_font = null;
            foreach ($fonts as $family => $font) {
                if ($this->fontSupportsGlyph($char_code, $font)) {
                    $mapped_font = $font;
                    break;
                }
            }

            if ($mapped_font === null) {
                break;
            }

            if (!isset($char_mapping[$start_index])) {
                $char_mapping[$start_index] = ["font" => $mapped_font, "length" => 0, "char" => null];
            }

            if ($mapped_font !== $char_mapping[$start_index]["font"]) {
                $char_mapping[$start_index]["length"] = $char_index - $start_index;
                if ($count > 0 && count($char_mapping) === $count) {
                    break;
                }
                $start_index = $char_index;
                $char_mapping[$start_index] = ["font" => $mapped_font, "length" => 0, "char" => null];
            }
        }

        // build the string for each mapping
        // foreach ($char_mapping as $start_index => &$info) {
        //     $info["char"] = mb_substr($text, $start_index, $info["length"]);
        // }

        return $char_mapping;
    }
```

bsweeney (changes requested):
Updated suggestions. A lot of the other functions would not longer be needed, though once we go 'round a few more times we can figure out what can be extracted.

Mellthas:
If you think this is ready for review, I can take a look at the changes.

bsweeney:
Thanks ... I'm doing some last minute testing to see if I can squeeze out a bit more performance. But I think the basic logic is ready for review so have at it.

bsweeney:
@robvanderlee FYI I rewrote the history to make this change a bit easier to follow.

There will always be a performance impact with this change since we're a) parsing each character of a text frame, and b) generating more frames (when required by font matching). I don't have a clean system for performance testing, but what testing I did do shows minimal impact on most documents. The greatest impact would be in the target scenario where the user just wants to specify a sequence of fonts and let Dompdf pick the right one. But even there the impact is not significant.

The base document looks like this:
```
<!doctype html>
<html>
    <head>
        <link rel="preconnect" href="https://fonts.googleapis.com">
        <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
        <link href="https://fonts.googleapis.com/css2?family=Sawarabi+Gothic&display=swap" rel="stylesheet">
        <style>
            @font-face {
                font-family: 'Great Vibes';
                font-style: normal;
                font-weight: 400;
                src: url('https://www.1001fonts.com/download/font/great-vibes.regular.ttf') format('truetype');
            }
        </style>
        <style>
            @font-face {
                font-family: 'Extraoin Zhurdlyou Anarchy';
                src: url('https://www.1001fonts.com/download/font/extraoin-zhurdlyou.anarchy.ttf') format('truetype');
                font-style: normal;
                font-weight: 400;
            }
        </style>
        <style>
            body {
                font-family: 'Great Vibes', 'Extraoin Zhurdlyou Anarchy', 'Sawarabi Gothic', serif;
            }
            .japanese {
                font-family: 'Sawarabi Gothic', serif;
            }
            .cyrillic {
                font-family: 'Extraoin Zhurdlyou Anarchy', serif;
                font-size: 32px;
                font-weight: 400;
            }
        </style>
    </head>
    <body>
        <!-- text goes here -->
    </body>
</html>
```

First test, 100 paragraphs of lorem ipsum:

| PHP | v2.0.3/Cpdf | vNext/Cpdf | v2.0.3/PDFLib | vNext/PDFLib | v2.0.3/GD | vNext/GD |
|-----|-------------|------------|---------------|--------------|-----------|----------|
| 7.1 | 641         | 704        | 420           | 671          | 11108     | 11178    |
| 8.2 | 595         | 608        | 319           | 582          | 6952      | 6969     |

Next test, a document with lots of manual font styling to render the text with the appropriate font. The document was 100 of the following paragraphs. This is not ideal since there is no text variability, but there aren't many lorem ipsum generators that work with multiple languages in the body content. Let alone segment the text appropriately.

```
<p><span class="japanese">能登</span> - <span class="cyrillic">Каждый</span> - <span class="japanese">半島 能登</span> - <span class="cyrillic">Каждый</span> - <span class="japanese">半島 能登</span> - This is great vibes font rendered text <span class="cyrillic">Каждый</span> - <span class="japanese">半島 能登</span> - <span class="cyrillic">Каждый</span> - <span class="japanese">半島 能登</span> - <span class="cyrillic">Каждый</span> - <span class="japanese">半島 能登</span></p>
```

| PHP | v2.0.3/Cpdf | vNext/Cpdf | v2.0.3/PDFLib | vNext/PDFLib | v2.0.3/GD | vNext/GD |
|-----|-------------|------------|---------------|--------------|-----------|----------|
| 7.1 | 1124        | 1133       | 473           | 1276         | 4822      | 4876     |
| 8.2 | 810         | 822        | 429           | 905          | 3047      | 3083     |

Final test, the same as above without the any manual text styling.

```
<p>能登 - Каждый - 半島 能登 - Каждый - 半島 能登 - This is great vibes font rendered text Каждый - 半島 能登 - Каждый - 半島 能登 - Каждый - 半島</p>
```

| PHP | v2.0.3/Cpdf | vNext/Cpdf | v2.0.3/PDFLib | vNext/PDFLib | v2.0.3/GD | vNext/GD |
|-----|-------------|------------|---------------|--------------|-----------|----------|
| 7.1 | 375         | 835        | 341           | 884          | 1615      | 5265     |
| 8.2 | 341         | 706        | 315           | 698          | 1160      | 3178     |

Mellthas:
As a quick note, the render tests fail because bold and italic font styles are not applied anymore.

bsweeney:
Thank for finding that ... wat?!?

bsweeney:
Unit tests seem happy and most of the "core" test I have up in the debug helper look OK on first pass. I am seeing a slight difference in the rendering of some document (e.g., https://eclecticgeek.com/dompdf/core_tests/encoding_utf-8_w3.html) that I need to investigate.

src/Helpers.php
```diff
@@ -590,24 +590,129 @@ public static function record_warnings($errno, $errstr, $errfile, $errline)
...
+     * Shim for use on systems running PHP < 7.2
+     *
+     * @param string $c
+     * @param string $encoding
+     * @return int|false
      */
-    public static function unichr($c)
+    public static function uniord(string $c, string $encoding = null)
     {
+        static $cache = [];
```
Mellthas, line 603: I wondered if the cache does much here really … So doing some rudimentary benchmarking of different variants (1 million iterations calling the functions with a fixed codepoint), I get results like the following on my machine (numbers are comparable on re-runs):

`mb_ord` variant | time
----------------------- | ------
`mb_ord` direct | 37.32 ms
`Helpers::uniord` | 79.99 ms
`Helpers::uniord` using `\array_key_exists` | 63.56 ms
`Helpers::uniord` without cache | 101.53 ms
`Helpers::uniord` just wrapping `mb_ord` | 59.54 ms

`mb_chr` variant | time
----------------------- | ------
`mb_chr` direct | 40.18 ms
`Helpers::unichr` | 77.17 ms
`Helpers::unichr` using `\array_key_exists` | 57.20 ms
`Helpers::unichr` without cache | 99.42 ms
`Helpers::unichr` just wrapping `mb_chr` | 67.29 ms

Note, the second `Helpers::unichr` variant just adds a leading backslash to `array_key_exists($c, $cache)` (see https://tideways.com/profiler/blog/compiler-optimized-php-functions for background).

So it seems, yes the cache speeds the operation up, but extra function calls and checks quickly eat up any benefits. Using the native functions directly seems to be significantly faster still.

Mellthas: Did a re-run using random codepoints; seems the cache can start to slow things down if it gets too large, e.g. using `rand(0, 2000)` as input for `Helpers::unichr` (haven’t thought about whether that is realistic at all):

`mb_chr` variant | time
---------------------- | ------
`mb_chr` direct | 86.87 ms
`Helpers::unichr` | 201.40 ms
`Helpers::unichr` using `\array_key_exists` | 186.53 ms
`Helpers::unichr` without cache | 144.91 ms
`Helpers::unichr` just wrapping `mb_chr` | 107.66 ms

I wonder if these differences would even make any noticeable impact when looking at the overall render time. So my hunch would be to just avoid any caching here, not worth the effort.

Mellthas (changes requested):

Mellthas:
Will look at the rest of the code later this week.

src/Helpers.php
Mellthas: Note: All tests done on PHP 8.2.

src/Helpers.php
```diff
@@ -590,24 +590,129 @@ public static function record_warnings($errno, $errstr, $errfile, $errline)
     }
 
     /**
-     * @param $c
-     * @return bool|string
+     * Get Unicode code point of character
+     *
+     * Shim for use on systems running PHP < 7.2
```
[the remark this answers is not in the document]
bsweeney: I've been pondering the idea. Supporting all the way back to PHP 7.1 definitely puts a crimp in things. I was looking at the [packagist stats](https://packagist.org/packages/dompdf/dompdf/php-stats) and the low bar for current installations seems to be PHP 7.4. I think we can safely move to that version with the next release.

src/Helpers.php
bsweeney: I think one of the issues I was fighting against in my performance testing was that I had XDebug enabled and forgot the fairly significant impact it can have on performance. I did all the optimization work prior to generating actual performance numbers over multiple runs. Whoops. It definitely helped find some things that could impact performance, but probably not in a significant way except on slower systems.

With the discussion of bumping the minimum version I think it's safe to say removing this shim entirely is worthwhile.

Mellthas: > I think one of the issues I was fighting against in my performance testing was that I had XDebug enabled and forgot the fairly significant impact it can have on performance.

Oh, yeah, XDebug definitely hit me, too. Switched to PHP 7.1 for comparison and wondered why everything was suddenly ten times as fast …

bsweeney:
> Unit tests seem happy and most of the "core" test I have up in the debug helper look OK on first pass. I am seeing a slight difference in the rendering of some document (e.g., https://eclecticgeek.com/dompdf/core_tests/encoding_utf-8_w3.html) that I need to investigate.

Found a two issues.
- Non-printable characters weren't excluded from the font mapping. This was causing unnecessary frame splitting which was particularly problematic for preformatted (`white-space: pre;`) text.
- Frame splitting performed around font mapping exposed an issue where unsupported characters in a font were not being included in text width calculations when using Cpdf as the backend. This was causing frame overlap when a frame consisted entirely of unsupported characters for the used font.

src/Helpers.php
bsweeney: FYI the worst-case scenario for the caching logic appears to be https://eclecticgeek.com/dompdf/core_tests/encoding_utf-8_all.html. It's taking minutes to render the document rather than seconds. Caching hurts in this case because most characters are not reused so a lot of work is done to maintain a cache that is unnecessary. So I think you're right, caching overall is not helpful.

src/Adapter/PDFLib.php
```diff
@@ -734,9 +734,7 @@ protected function _load_font($font, $encoding = null, $options = "")
         // Check if the font is a native PDF font
         // Embed non-native fonts
         $test = strtolower($baseFont);
-        if (in_array($test, DOMPDF::$nativeFonts)) {
-            $font = basename($font);
-        } else {
+        if (!in_array($test, DOMPDF::$nativeFonts)) {
```
Mellthas, line 737: While at it, `$isNativeFont` is determined directly before, I think that could be reused here.

src/FontMetrics.php
```diff
@@ -326,6 +339,84 @@ public function getTextWidth(string $text, $font, float $size, float $wordSpacin
...
+     * the actual text of the substring.
+     *
+     * @param string $text            The text to map
+     * @param array  $fontFamilies    List of font families to map against
+     * @param string $subtype         The font subtype (italic, bold, etc.)
+     * @param int    $count           The number of matches to return
+     * @param bool   $returnSubstring Should the actual matched text be returned
+     * @return array
+     */
+    public function mapFontsToText(string $text, array $fontFamilies, string $subtype = "normal", int $count = -1, bool $returnSubstring = false): array
```
Mellthas, line 358: `mapTextToFonts`?

src/FontMetrics.php
```diff
@@ -65,6 +66,18 @@ class FontMetrics
      */
     private $options;
 
+    /**
+     * Character mapping table for fonts
+     * @var array
+     */
+    private $unicodeCharMapTables = [];
```
Mellthas, line 73: Both properties are unused.

src/FontMetrics.php
```diff
@@ -326,6 +339,84 @@ public function getTextWidth(string $text, $font, float $size, float $wordSpacin
...
+            // build the string for each mapping
+            foreach ($char_mapping as $start_index => &$info) {
+                $info["text"] = mb_substr($text, $start_index, $info["length"]);
+            }
+        }
+
+        return $char_mapping;
+    }
+
+    public function isCoreFamily(string $family): bool
```
Mellthas, line 415: Unused.

src/FrameReflower/Text.php
```diff
@@ -386,13 +389,35 @@ function reflow(BlockFrameDecorator $block = null)
...
+        $text = $this->pre_process_text($frame->get_text());
+
+        // map text to fonts based on supported Unicode range
+        $families = array_map(
+            function ($value) {
+                return trim($value, " '\"");
+            },
+            explode(",", $style->get_specified("font_family"))
+        );
+        $charMapping = $font_metrics->mapFontsToText($text, $families, $font_metrics->getType($style->font_weight . ' ' . $style->font_style), 1);
```
Mellthas, line 404: This has not changed, but still wondering here: As `FontMetrics::getType()` is always used with a parameter like `"$weight $style"` and then goes on to split that input again, why does the function not just take 2 parameters?

src/FrameReflower/Text.php
```diff
@@ -386,13 +389,35 @@ function reflow(BlockFrameDecorator $block = null)
 
         // Determine the text height
```
Mellthas, line 390: The comment is separated from the code it refers to now.

src/FrameReflower/Text.php
```diff
@@ -386,13 +389,35 @@ function reflow(BlockFrameDecorator $block = null)
...
+            function ($value) {
+                return trim($value, " '\"");
+            },
+            explode(",", $style->get_specified("font_family"))
+        );
+        $charMapping = $font_metrics->mapFontsToText($text, $families, $font_metrics->getType($style->font_weight . ' ' . $style->font_style), 1);
+        if (isset($charMapping[0])) {
+            if ($charMapping[0]["length"] !== 0) {
+                $frame->split_text($charMapping[0]["length"]);
+                $text = $this->pre_process_text($frame->get_text());
```
Mellthas, line 408: `$this->pre_process_text()` should no be needed here, already handled above.

lib/Cpdf.php
```diff
@@ -5164,7 +5164,7 @@ public function getTextWidth(float $size, string $text, float $wordSpacing = 0.0
             $this->selectFont($this->defaultFont);
         }
 
-        $text = str_replace(["\r", "\n"], "", $text);
+        $text = preg_replace('/[\x00-\x1F\x7F\xA0]/u', '', $text);
```
Mellthas, line 5167: I think `\xA0` slipped in there accidentally. Would be nice to have a short comment that this is about non-printable characters here.

lib/Cpdf.php
```diff
@@ -5186,14 +5186,18 @@ public function getTextWidth(float $size, string $text, float $wordSpacing = 0.0
 
                 if (isset($current_font['C'][$char])) {
                     $char_width = $current_font['C'][$char];
+                } elseif (isset($current_font['C'][0xFFFD])) {
+                    $char_width = $current_font['C'][0xFFFD];
```
Mellthas, line 5190: Similarly, a short comment mentioning which character this is would be nice.

Mellthas (changes requested):

Mellthas:
The font-resolve logic is now essentially duplicated: We have the existing logic in the `Style` class and the new `mapFontsToText`, which also takes the current text into account. Some ideas on how to resolve this:

* Change the `font-family` logic in the `Style` class to compute the declaration to an array of the font names instead of the first available font. Maybe the final font paths, so non-existing fonts would be dropped already.
* The `Text` reflower would then set the final font name as the used value of the property, as it does now. Whether the property has been resolved could then be determined based on whether it returns an array or a string.
* The question would then be how to handle the other places where `font-family` is checked on non-text frames, e.g. when determining the height of inline frames. Might be right to use the first available font there, as it is essentially done now, but I haven’t verified that.

But maybe this does not need to be addressed right now, and can be tackled later.

src/FrameReflower/Text.php
```diff
@@ -386,13 +389,35 @@ function reflow(BlockFrameDecorator $block = null)
...
+            }
+            $mapped_font = $charMapping[0]["font"];
+            if ($mapped_font !== null) {
```
Mellthas, line 411: Could fall back to the first available font here directly, so it does not need to be determined again below (in line 417 via the `Style` getter). Maybe `mapFontsToText` should directly fall back to the first font instead of `null`?

src/FontMetrics.php
```diff
@@ -326,6 +339,84 @@ public function getTextWidth(string $text, $font, float $size, float $wordSpacin
...
+     */
+    public function mapFontsToText(string $text, array $fontFamilies, string $subtype = "normal", int $count = -1, bool $returnSubstring = false): array
+    {
+        $char_mapping = [];
+        $fonts = [];
+
+        foreach ($fontFamilies as $family) {
+            $font = $this->getFont($family, $subtype);
+            if ($font !== null) {
+                $fonts[$family] = $font;
```
Mellthas, line 366: The array keys (`$family`) are never used, can simply use a list here.

Mellthas:
One more thing: Shrink-to-fit width determination (for table columns, float, etc.) is now incorrect when text is split via the new logic, since `get_min_max_width` is called before reflow. That should be solvable by applying the split early if needed, but would need a way to determine whether the final font has been determined for a specific frame to avoid checking the text twice.

robvanderlee:
@bsweeney, @Mellthas great to see the improvements! I'll be sure to run some testing on my part and see If i can contribute to the review and/or make some suggested changes. I've had a busy couple weeks behind me but I'll make some time in this week to give a proper look.

src/Helpers.php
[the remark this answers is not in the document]
bsweeney: Work related to increasing the minimum supported version of PHP can happen in the [php74 branch](https://github.com/dompdf/dompdf/tree/php74).

src/FrameReflower/Text.php
bsweeney: The frame text hasn't yet been set to the processed version. I think we need to move the `set_text`, which is currently below this logic, before this block so that the split occurs in the correct location after processing. Even so, the split frames will have their text run through pre-processing again, which is not ideal.

src/FrameReflower/Text.php
bsweeney: Then again, it's not used in very many places so maybe I will go ahead and make that change.

lib/Cpdf.php
bsweeney: indeed

src/Adapter/CPDF.php
```diff
@@ -786,6 +786,39 @@ public function add_link($url, $x, $y, $width, $height)
         }
     }
 
+    public function font_supports_text(string $font, string $text)
```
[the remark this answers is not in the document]
bsweeney: I did some quick testing using some (poorly) mocked up logic over 10000 iterations. For each test I checked to see if the codepoint existed in an array. The runs were:
1. generate a character code at random and run the check
2. pick a random character code from an array, put it in an array with a singular item, run the check in a foreach
3. pick a random character from an array, put it in a string, mb_strsplit => array_unique => array_map(mb_ord), run the check in a foreach

The last test attempting to be much like what's going on with this change. You can definitely [see the performance impact](https://3v4l.org/k6gbU#v8.2.5), though the time taken is still small enough that maybe it's OK. Still I'll see if I can work out some further improvement to the logic.

src/FontMetrics.php
bsweeney: Thanks, I missed these when I was cleaning up from previous iterations of the logic.

src/FontMetrics.php
bsweeney: I think I was using it at some point, but that logic must've been dropped.

src/Adapter/CPDF.php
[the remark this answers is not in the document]
Mellthas: In my mind this is less about optimization than about keeping it simple while solving the problem at hand. The current solution has a higher code complexity _and_ it incurs a performance cost, I think that is not worth it when all it gains is a maybe slightly more convenient interface for a hypothetical user; the current variant could always be implemented in addition, if there is demand or the need for it later. I see the `Canvas` interface as primarily serving the rendering process, and only secondarily as a convenient interface for manipulating the PDF directly – most users will never touch it directly, or only use the most basic methods.

That said, maybe it is possible to move some code around so that the current signature would make more sense.

src/FrameReflower/Text.php
Mellthas: I think you are right, yes.

src/FontMetrics.php
Mellthas: It is definitely funny to have both kind of contradicting each other. Though the more I think about it, the more confused I get right now :-D.

bsweeney: I think I'll go with your suggestion.

src/FrameReflower/Text.php
bsweeney: Actually, following through usage and I think I'll put this off for that future change. The way that return value flows through to other method calls indicates that some more thought should be put into it than I want to do now.

src/FrameReflower/Text.php
bsweeney: First available specified may not be a valid font-family. We could default to the first valid in the mapping logic. Though then that method does not indicate when text is completely unsupported by the specified fonts. Which I guess is a question of generalization vs. our particular use case.

As you can tell I lean towards generalization where I can. I think this shouldn't be an issue in most cases where:
1. the user is specifying valid fonts
2. the text is supported by one of the specified fonts

And even if we do have to fall back to the Style class logic, so long as there aren't a lot of splits that require falling back the performance impact should be minimized.

src/Adapter/CPDF.php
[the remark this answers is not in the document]
bsweeney: I testing variations of the method where the accepted second argument is one of the following:
- a string
- a single character
- a single character code

and did not see significant performance difference on my own system. I'll have to take a second look, but maybe because the FontMetrics method has already split the string up the overhead in the Canvas method is minimal (even though it's doing some of the same work).

I'll have to take another look to confirm but it seems like we're unlikely to squeeze much more performance out of this logic.

src/Adapter/PDFLib.php
```diff
@@ -726,17 +726,14 @@ protected function _load_font($font, $encoding = null, $options = "")
         // Fix for PDFLibs case-sensitive font names
         $baseFont = basename($font);
         $isNativeFont = false;
-        if (isset(self::$nativeFontsTpPDFLib[$baseFont])) {
-            $font = self::$nativeFontsTpPDFLib[$baseFont];
+        $lcBaseFont = strtolower($basefont);
```
Mellthas, line 729: Variable `$baseFont` is misspelled here.

src/FrameReflower/Text.php
```diff
@@ -409,12 +433,12 @@ function reflow(BlockFrameDecorator $block = null)
 
         // Skip wrapped white space between block-level elements in case white
         // space is collapsed
-        if ($frame->get_text() === "" && $frame->get_margin_width() === 0.0) {
+        if ($text === "" && $frame->get_margin_width() === 0.0) {
```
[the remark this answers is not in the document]
bsweeney: I noticed the failing tests but hadn't yet had a chance to look into it. Thanks for finding that.

src/Adapter/PDFLib.php
bsweeney: really need to get my environment fixed up (or my eyes)

src/FrameReflower/Text.php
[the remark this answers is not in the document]
bsweeney: Thanks! Much better now.

src/FrameReflower/Text.php
bsweeney: Yeah, it's quite the edge case. I'll think on this more while I think on some of the other things you brought up.

bsweeney: I'm going to table further improvements for now.

bsweeney:
> One more thing: Shrink-to-fit width determination (for table columns, float, etc.) is now incorrect when text is split via the new logic, since `get_min_max_width` is called before reflow. That should be solvable by applying the split early if needed, but would need a way to determine whether the final font has been determined for a specific frame to avoid checking the text twice.

There appear to be two issues related to this?
1. Incorrect width determination since the used font may have differing widths than the first valid.
2. When a frame is split the new frame is pushed to the next line.

I think you're right that performing the font mapping earlier in the process is probably necessary. I'll need to look at this some more.

src/Adapter/CPDF.php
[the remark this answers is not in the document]
bsweeney: Going to leave as-is for now. We can re-visit the optimization at a later date.

bsweeney:
Update to address the shrink-to-fit issue by moving the font mapping logic to the decorator so that it can be called as necessary. Duplicated calls are avoided by recording the mapped font in a private class. I do think it would work better to handle recognition of mapping being complete by inspecting the used value ... somehow. I didn't see an obvious _good_ way to do this so I'll leave that as an optimization for another time.

bsweeney (approved):

PR #3163, opened 2023-03-20
Fix automatic counter reset

Mellthas:
From the spec [1]:

> If 'counter-increment' or 'content' on an element or pseudo-element
> refers to a counter that is not in the scope of any 'counter-reset',
> implementations should behave as though a 'counter-reset' had reset
> the counter to 0 on that element or pseudo-element.

Test cases are included.

`auto reset nested` is a simplified version of the test from https://eclecticgeek.com/dompdf/core_tests/css_counter.html. Note, how the second and third counters were not reset properly on the last two items before there (_4|14|-15_ and _5|14|-15_ instead of _4|0|0_ and _5|0|0_).

[1] https://www.w3.org/TR/CSS21/generate.html#scope

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -2066,63 +2066,6 @@ class Style
     }
 
     /**
-     * @param string $value
-     * @param int    $default
-     *
-     * @return array|string
-     */
-    protected function parse_counter_prop(string $value, int $default)
-    {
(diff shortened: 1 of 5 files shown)


bsweeney:
That generated content test is 💯

bsweeney (approved):

PR #3176, opened 2023-04-02
Improve handling of case-insensitivity during style parsing

Mellthas:
* **Improve handling of case-insensitivity during style parsing**
Handle case-insensitivity individually per property: Keywords need to be handled case-insensitively, but custom identifiers are case-sensitive; whether something is a keyword or a custom identifier depends on the property in question (e.g. `THIN` is to be treated equivalently to `thin` when parsed as border style, but as a counter name, `THIN` is distinct from `thin`).
* **Remove support for HTML-compatibility `list-style-type` values in CSS**
These are neither in the specs, nor supported by current browsers. The values are mapped to regular counter-style names in the `AttributeTranslator` class now. To clarify: `type="A"` is still supported in HTML, and translated to `list-style-type: upper-alpha;`, but support for `list-style-type: A;` is removed (treated as `list-style-type: decimal;` now because the name is unrecognized).
* **Move `content` property parsing to the `Style` class**
  * Compute the `content` and `quotes` properties to drop invalid  declarations, and make the parsing stricter
  * Do not lower-case the counter name and string arguments for the  `counter` and `counters` functions
  * Only strip outer quotes when parsing CSS strings (e.g.  `content: "'Quoted'"`; would result in the string `Quoted` instead of  `'Quoted'`)
  * Properly inherit the `auto` value for the `quotes` property (will  only make a difference when auto quotes consider the current language)
  * As long as we are dealing with ASCII-only string parts, we don't  need to use the `mb_*` functions
* **Improve `list-style` shorthand parsing**
* **Improve `background-position` parsing**

https://www.w3.org/TR/css-content-3/#content-property
https://www.w3.org/TR/CSS2/generate.html#list-style
https://www.w3.org/TR/css-counter-styles-3/#predefined-counters

Regarding the HTML-compatibility values for `list-style-type`: We could also retain support for these values in CSS as a legacy feature if desired.

### DIFF
--- a/src/Adapter/PDFLib.php
+++ b/src/Adapter/PDFLib.php
@@ -333,7 +333,6 @@ class PDFLib implements Canvas
      * @param int $object the ID of a previously opened object
      *
      * @throws Exception
-     * @return void
      */
     public function reopen_object($object)
     {
--- a/src/Css/AttributeTranslator.php
+++ b/src/Css/AttributeTranslator.php
@@ -167,14 +167,14 @@ class AttributeTranslator
         'ol' => [
             'compact' => 'margin: 0.5em 0;',
             'start' => 'counter-reset: -dompdf-default-counter %d;',
-            'type' => 'list-style-type: %s;',
+            'type' => '_set_list_style_type',
         ],
         'ul' => [
             'compact' => 'margin: 0.5em 0;',
-            'type' => 'list-style-type: %s;',
+            'type' => '_set_list_style_type',
         ],
         'li' => [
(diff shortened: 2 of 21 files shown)


bsweeney:
> Regarding the HTML-compatibility values for list-style-type: We could also retain support for these values in CSS as a legacy feature if desired.

I think we can safely drop support for the non-standard values.

Mellthas:
Added a commit computing `font-weight`, which slightly simplifies the font-family resolution in the `Style` class.

Mellthas:
Thank you!

PR #3185, opened 2023-04-17
Use uppercase Imagick format

Starfox64:
The canonical way to represent ImageMagick formats is uppercase. ImageMagick generally does not care and will still work with lowercase formats except when applying coders policies from policy.xml as these are case sensitive. In this instance, lowercase formats would fail the policy despite being authorized.

Here is the policy suggested by ImageMagick that previously failed :
```xml
<policy domain="delegate" rights="none" pattern="*" />
<policy domain="filter" rights="none" pattern="*" />
<policy domain="coder" rights="none" pattern="*" />
<policy domain="coder" rights="read|write" pattern="{GIF,JPEG,PNG,WEBP,BMP}" />
```

It fails because PNG is listed in uppercase on the last line. The current workaround is to add the lowercase version like so :
```xml
<policy domain="coder" rights="read|write" pattern="{GIF,JPEG,PNG,png,WEBP,BMP}" />
```

I'm submitting the PR because I believe some hosts come with a restrictive policy and will not allow you to customize this policy. This would allow using this library on such hosts and to provide a better experience to ImageMagick users.

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -5706,7 +5706,7 @@ EOT;
             }
 
             $imagick = new \Imagick($file);
-            $imagick->setFormat('png');
+            $imagick->setFormat('PNG');
 
             // Get opacity channel (negative of alpha channel)
             if ($imagick->getImageAlphaChannel()) {
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -5716,7 +5716,14 @@ EOT;
                 if (\Imagick::getVersion()['versionNumber'] < 1800) {
                     $alpha_channel->negateImage(true);
                 }
-                $alpha_channel->writeImage($tempfile_alpha);
+
+                try {
+                    $alpha_channel->writeImage($tempfile_alpha);
+                } catch (\ImagickException $th) {
+                    // Backwards compatible retry attempt in case the IMagick policy is still configured in lowercase
+                    $alpha_channel->setFormat('png');
+                    $alpha_channel->writeImage($tempfile_alpha);
+                }
 
                 // Cast to 8bit+palette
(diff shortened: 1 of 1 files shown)


bsweeney:
Should we try uppercase and, if it fails, try lowercase?

Starfox64:
For reference, this is the exception that is thrown on policy violation:
`ImagickException`
`attempt to perform an operation not allowed by the security policy `FORMAT_NAME_HERE' @ error/constitute.c/IsCoderAuthorized/421`

bsweeney:
I don't have a system where I can test this easily at the moment so your help investigating further would be great.

If we use the uppercase version of the string does it pass the policy restriction if the policy is configured with only the lowercase string?

```
<policy domain="coder" rights="read|write" pattern="{png}" />
```

Uppercase seems to be the standard so I think making the change to the code is prudent. If specifying the format in uppercase always works then we're all good. But if not, I'd like to see if we can further harden the logic against the exception in this scenario. To that end I have two thoughts.

First, can use the [queryFormats](https://www.php.net/manual/en/imagick.queryformats.php) method to get the correct format casing? Or does that always returns uppercase?

Second, is the exception thrown when Dompdf attempts to set the format or when it attempts the actual conversion? Namely I'm interested in if we could do something like the following:

```
try {
    $imagick->setFormat('PNG');
} catch (ImagickException ex) {
    $imagick->setFormat('png');
}
```

so that we can try the lowercase format when we get that exception in case that's how the policy is configured.

Starfox64:
I did some checking on my end and the error occurs during `writeImage`, I was able to confirm that this change allows everything to continue working even if the policy was previously set to the lowercase version.

PR #3220, opened 2023-06-03
Make exception of corrupt tables fetchable

tobiasbambullis:
Our customers can enter html by their own. When they forget to close a tr element for example, there is an uncaught exception because get_cellmap is executed on null. To fetch this we add a simple exception.


bsweeney:
I haven't yet merged, but the same (well, similar) change was going to be included in the next release. See 2b8fddee772e1a3fd0be3eb37e7c300460cbe48e.

Mellthas:
The changes have been merged into `master` in the meantime.

PR #3255, opened 2023-08-07
Implement css vars

tmolitor-stud-tu:
I'm not sure if this complys wih your coding guidlines and it might not be the way you'd like it be solved, but maybe it's still useful :)

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -1248,7 +1248,8 @@ class Style
             $prop = self::$_props_alias[$prop];
         }
 
-        if (!isset(self::$_defaults[$prop])) {
+        // props beginning with "--" (changed above to "__") are css variables --> always allow them
+        if (!isset(self::$_defaults[$prop]) && !substr($prop, 0, 2) == '__') {
             global $_dompdf_warnings;
             $_dompdf_warnings[] = "'$prop' is not a recognized CSS property.";
             return;
@@ -1367,7 +1368,8 @@ class Style
             $prop = self::$_props_alias[$prop];
         }
 
-        if (!isset(self::$_defaults[$prop])) {
+        // props beginning with "--" (changed in set_prop() to "__") are css variables --> always allow them
+        if (!isset(self::$_defaults[$prop]) && !substr($prop, 0, 2) == '__') {
             throw new Exception("'$prop' is not a recognized CSS property.");
         }
 
@@ -1508,6 +1510,12 @@ class Style
      */
     protected function compute_prop(string $prop, $val)
     {
+        // handle css var() functions as already computed (late binding)
+        // TODO: extend to calc() and other css functons?
+        if(is_string($val) && substr($val, 0, 4) == 'var(') {
+            return $val;
+        }
+
         // During style merge, the parent style is not available yet, so
         // temporarily use the initial value for `inherit` properties. The
         // keyword is properly resolved during inheritance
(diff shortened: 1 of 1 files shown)


bsweeney (changes requested):
I look forward to reviewing this change more fully. In the meantime a few small requests:

- Across the board you should to wrap the body of a conditional statement, loop, etc. in braces (`{ }`) to avoid potential mistake or misunderstanding during future edits.
- Do you have test cases?

tmolitor-stud-tu:
I've fixed the curly braces, testcases are coming soon :)

tmolitor-stud-tu:
A test case would be the following html code (I'm not sure how to write proper dompdf testcases to test this programmatically):

```
<!DOCTYPE html>
<html>
	<head>
		<style>
			.inner {
				background-color: var(--custom-color);
			}
			.middle1 {
				background-color: var(--custom-color);
			}
			div {
				height: 4em;
				width: 4em;
			}
			.outer {
				height: 16em;
				width: 16em;
				--custom-color: green;
				background-color: blue;
			}
			.middle2 {
				background-color: yellow;
				height: 9em;
				width: 9em;
				--custom-color: red;
			}
		</style>
	</head>
	<body>
		<div class="outer">
			<div class="middle1"></div>
			<div class="middle2">
				<div class="inner"></div>
			</div>
		</div>
	</body>
</html>
```
See [this jsFiddle](https://jsfiddle.net/k40cqfpy/)
And [the PDF output](https://github.com/dompdf/dompdf/files/12552603/dompdf_cssvar.pdf)

wundii:
i think, that including the css vars is a good idea, unfortunately this does not work with bootstrap 5.*, with defined variables in the :root{--bs-green:#198754} element. warning: Undefined array key "__bs_green"

bsweeney:
@tmolitor-stud-tu thoughts on #3314? I think that PR takes the idea a bit further to accommodate, for example, inheritance.

bsweeney:
@tmolitor-stud-tu any thoughts on #3314?

src/Css/Style.php
```diff
@@ -1551,6 +1559,23 @@ protected function computed(string $prop)
...
+                if(is_string($part) && substr($part, 0, 4) == 'var(') {
+                    $css_var = str_replace("-", "_", substr(trim($part), 4, -1));
+                    $part = $this->get_specified($css_var) ?? (isset($this->parent_style) ? $this->parent_style->get_specified($css_var) : null);
+		}
+            }
+            if(!is_array($this->_props_computed[$prop])) {
+                $this->_props_computed[$prop] = $value[0];
+            }
+        }
+
```
bsweeney, line 1578: This should probably be handled in the compute_prop method so that the value can then be parsed by any property-specific compute methods.

src/Css/Style.php
```diff
@@ -1508,6 +1510,12 @@ public function __get(string $prop)
      */
     protected function compute_prop(string $prop, $val)
     {
+        // handle css var() functions as already computed (late binding)
+        // TODO: extend to calc() and other css functons?
+        if(is_string($val) && substr($val, 0, 4) == 'var(') {
+            return $val;
```
bsweeney, line 1516: I think think we want to return here. How would we accommodate something like:

```
<style>
    :root {
        --half: 50%;
    }
    div {
        width: var(--half);
    }
</style>
<div>dompdf</div>
```

src/Css/Style.php
bsweeney: Is this logic sufficient to accommodate something like:

```
<style>
    :root {
        --half: 50%;
    }
    div {
        width: var(--half);
    }
</style>
<div>dompdf</div>
```

bsweeney (changes requested):
Some initial thoughts for discussion. I haven't thoroughly evaluated this in relation to the spec nor in relation to how we generally parse CSS properties.

bsweeney:
FYI at this point I'm inclined to go with 3314 since some of the logic is built out a bit more. Maybe you can collaborate on that PR?

tmolitor-stud-tu:
> FYI at this point I'm inclined to go with 3314 since some of the logic is built out a bit more. Maybe you can collaborate on that PR?

That's okay with me :) I'm just glad support for css-vars is finally getting into dompdf. And if my pr did just the initial kickoff for this, that's fine by be, too :)

bsweeney:
I think the other PR is just about ready. Just waiting on any final reviews from involved devs.

PR #3269, opened 2023-09-01
Add experimental PDF/A compliance mode

### LINKED ISSUE #1106: PDF/A
I would like to generate PDF in PDF/A Model?

Starfox64:
This PR adds experimental PDF/A compliance mode and closes #1106 

I'm calling this experimental for multiple reasons:
- Currently only supports the CPDF backend
- Only takes care of adding the required metadata
- Doesn't force font embedding, it's on the user to embed their fonts
- I can only test so many documents, the user should generally verify actual compliance with a tool like veraPDF (profile PDF/A-3b) and report issues
- Isn't as robust GhostScript

This still should work for most people on non esoteric documents.


lib/Cpdf.php
```diff
@@ -2975,6 +3062,135 @@ protected function o_embedded_file($id, $action, $options = null): ?string
...
 
+    /**
+     * Enable PDF/A compliance mode
+     */
+    public function enablePdfACompliance()
+    {
+        $this->pdfa = true;
+
+        $this->o_catalog($this->catalogId, 'outputIntents', [
+            'iccProfileData' => file_get_contents(__DIR__ . '/res/sRGB2014.icc'),
```
williamdes, line 3073: can this path be a variable on a line above ?
In Debian packaging I will have to change it to point to a path located in another package

Starfox64: Just to make sure I understand, you'd like it to be like this?

```php
$iccProfilePath = __DIR__ . '/res/sRGB2014.icc';

$this->o_catalog($this->catalogId, 'outputIntents', [
            'iccProfileData' => file_get_contents($iccProfilePath ),
```

Or are you asking for the path to be exposed in configuration?

williamdes: That would be even easier to patch or link

For tcpdf this is what is done: https://sources.debian.org/src/tcpdf/6.7.4%2Bdfsg-1/debian/php-tcpdf.links/#L1

lib/Cpdf.php
```diff
@@ -2975,6 +3062,137 @@ protected function o_embedded_file($id, $action, $options = null): ?string
...
+     */
+    public function enablePdfACompliance()
+    {
+        $this->pdfa = true;
+
+        $iccProfilePath = __DIR__ . '/res/sRGB2014.icc';
+
+        $this->o_catalog($this->catalogId, 'outputIntents', [
+            'iccProfileData' => file_get_contents($iccProfilePath),
+            'iccProfileName' => 'sRGB2014.icc',
```
williamdes, line 3076: should this be basename
```suggestion
            'iccProfileName' => basename($iccProfilePath),
```

lib/res/sRGB2014.icc
williamdes, line 1: Can you say where you got this file?
What licence it is?
Is this the same file as https://sources.debian.org/src/tcpdf/6.7.4%2Bdfsg-1/include/sRGB.icc/?

Starfox64: Source: https://www.color.org/srgbprofiles.xalter
License: https://www.color.org/profiles2.xalter#license

williamdes: Thank you!
Let's let @bsweeney let us know if adding a licence file should be done like in https://github.com/dompdf/php-font-lib/tree/master/tests/resources/fonts/ahem

williamdes:
@Starfox64 do you think it would solve #3269 ?

Starfox64:
That's the wrong Issue ID I think.

williamdes:
Indeed, this one: https://github.com/dompdf/dompdf/issues/3442

Could you rebase your work ?
I was about to try it but composer wants 2.1+

Starfox64:
I've done a merge recently and it looks like I'm ahead of master.

With regards to #3442 I think this might be fixed by this line https://github.com/dompdf/dompdf/pull/3269/files#diff-f365bcb24d5c081b6bfde4aa59a8eda68824ea9658c362098554fc659dcec1d0R3426

It's funny to see that we are literally trying to do the same thing btw, nothing spells out fun more than trying to embed XML in PDF metadata.

williamdes:
> I've done a merge recently and it looks like I'm ahead of master.
> 
> With regards to #3442 I think this might be fixed by this line https://github.com/dompdf/dompdf/pull/3269/files#diff-f365bcb24d5c081b6bfde4aa59a8eda68824ea9658c362098554fc659dcec1d0R3426
> 
> It's funny to see that we are literally trying to do the same thing btw, nothing spells out fun more than trying to embed XML in PDF metadata.

Okayy, I got it wrong: `barryvdh/laravel-dompdf[v2.1.0, ..., v2.1.1] require dompdf/dompdf ^2.0.3`
So I will "downbase" your work and test it for 2.0

williamdes:
https://github.com/williamdes/dompdf/tree/pdfa-2.0

and https://github.com/williamdes/dompdf/tree/feat/pdfa is a rebased version without merges

bsweeney:
FYI I'm re-targeting this to 3.1 (from 3.0.1) since it's a new feature not a patch. I know there are a lot of issues slate for 3.0.x releases but I suspect most of those will be pushed out once I have a chance to review them.

williamdes (approved):
Can you add `Fixes: #3442` to this PR ?
It passes nearly all compliance tests except
```
[Specification: ISO 19005-3:2012, Clause: 6.2.11.4.1, Test number: 1](https://github.com/veraPDF/veraPDF-validation-profiles/wiki/PDFA-Parts-2-and-3-rules#rule-621141-1)	
The font programs for all fonts used for rendering within a conforming file shall be embedded within that file, as defined in ISO 32000-1:2008, 9.9.	Failed
```

williamdes:
Would you mind reviewing this first batch @bsweeney ?
This is quite critical for me as I need to progress into PDF/A + Factur-x

lib/res/sRGB2014.icc
bsweeney: I think this makes sense. Naming ... something like sRGB2014.icc.LICENSE (ugly)? Or should we put the ICC profile in a sub-directory with the standard-named LICENSE file?

```
The copyright owner and terms of use of an ICC profile are normally identified in the Creator field in the profile header and in the Copyright tag. Where ICC is the copyright owner, the following license terms apply:

"This profile is made available by the International Color Consortium, and may be copied, distributed, embedded, made, used, and sold without restriction. Altered versions of this profile shall have the original identification and copyright information removed and shall not be misrepresented as the original profile."

(reference https://www.color.org/profiles2.xalter#license)
```

lib/Cpdf.php
```diff
@@ -76,6 +76,11 @@ class Cpdf
      */
     public $signatureMaxLen = 5000;
 
+    /**
+     * @var bool Is PDF/A compliance mode enabled
+     */
+    public $pdfa = false;
```
bsweeney, line 82: Do we want to generically support PDF/A since various levels of compliance are defined? Should we just support a specific version which evolves as we update the library.

Also, instead of a flag for PDF/A support should we define constants (e.g., DOMPDF::PDF17, DOMPDF::PDFA1A) for various specifications and store the desired spec in a property, using that to toggle functionality?

wblessen: I think, pdfa is a good start, as it is mandatory 1st of january here.

I tested this solution with  adding "DejaVu Sans" as defaultFont, but verapdf it is still throwing this error:

The font programs for all fonts used for rendering within a conforming file shall be embedded within that file, as defined in ISO 32000-1:2008, 9.9	Failed
1 occurrences	[Hide](https://demo.verapdf.org/#)
PDFont	
Subtype == "Type3" || Subtype == "Type0" || renderingMode == 3 || fontFile_size == 1	
root/document[0]/pages[0](6 0 obj PDPage)/contentStream[0](7 0 obj PDContentStream)/operators[102]/font[0](Helvetica-Bold)

bsweeney:
FYI, I force pushed a rebase for this branch so the merge commit wasn't necessary. I'll be taking a look at what else we need and comment and/or commit additional changes.

williamdes:
> FYI, I force pushed a rebase for this branch so the merge commit wasn't necessary. I'll be taking a look at what else we need and comment and/or commit additional changes.

Thank you, let us know if some help is needed

lib/Cpdf.php
bsweeney: Upon further consideration I think we can stick with a simple flag for now. While I like the idea of being able to indicate the level of support using a constant it's unrealistic to expect we'll add varying levels of support anytime soon. We can revisit the implementation in the future if we decide to support other variants of PDF/A.

bsweeney:
@wblessen it's possible to achieve PDF/A compliance with the changes here. I'll add a wiki article outlining the requirements and provide sample code.

To address your specific issue regarding fonts, you should set your default font by passing it in as part of the instantiation options and your document should not specify any fonts that would break compliance. For example, if you style your document with a generic font family then Dompdf will translate that into a core PDF font. So, for example, `font-family: sans-serif` is translated to mean "use the core PDF font Helvetica" which is not an embeddable font.

Something like the following should generate a compliant document:

```
$dompdf = new Dompdf([
    "defaultFont" => "DejaVu Sans",
    "isPdfAEnabled" => true
]);
$html = <<<EOF
<html>
<head>
    <style>
        * { font-family: DejaVu Sans; }
    </style>
</head>
<body>
    <p>This is PDF/A-3b compliant.</p>
</body>
EOF;
$dompdf->loadHtml($html);
```

I'm still thinking through how to ensure text styled with a generic family uses an embeddable font. Most likely it will require remapping the font families prior to render. I'll include that information in the wiki once I have a solution.

williamdes:
For reference I am posting some documentation I found on mpdf
https://mpdf.github.io/what-else-can-i-do/pdf-a3-xmp-rdf.html

Seems to have been implemented in https://github.com/mpdf/mpdf/pull/558/
and rdf data in https://github.com/mpdf/mpdf/commit/4978b70eea405aed75633bd05ecbe3eab5b0b947

Maybe we can copy the example and make it for dompdf

bsweeney (commented):
While I do see some room for improvement, I think it's best to put some of these potential changes off to a future release so we can move forward with getting this out there for those who need the functionality now.

I do have some improvements so hold tight just a little longer.

lib/Cpdf.php
```diff
@@ -3198,6 +3418,12 @@ function output($debug = false)
 
         $xref = [];
         $content = '%PDF-' . self::PDF_VERSION;
+
+        if ($this->pdfa) {
+            // Force binary mode with 4 random bytes above 127
+            $content .= "\n%" . chr(rand(128, 256)) . chr(rand(128, 256)) . chr(rand(128, 256)) . chr(rand(128, 256));
```
bsweeney, line 3426: FYI, the max value for the `rand` function should be 255 since, per documentation, the `chr` codepoint should be "An integer between 0 and 255." The function `$bytevalue %= 256;`'s the value, which results in a codepoint of 0 when 256 is passed in.

Noting for reference, will fix in an upcoming push.

bsweeney:
I ran the code through my local test suite and found a few minor compatibility problems. I also enabled PDF/A when using the PDFLib backend (which did not require many changes).

Only thing I haven't been able to address so far is compliance when using CMYK colors/images. I'll look at that issue a bit more, but I suspect addressing CMYK support may have to wait. I think the correct resolution would be to use a consistent color profile in the document, which would require a bit of effort. Plus, the CMYK profiles from ICC are fairly large compared to Dompdf itself (2 - 8 MB for the ones I looked at).

lib/Cpdf.php
```diff
@@ -2975,6 +3076,136 @@ protected function o_embedded_file($id, $action, $options = null): ?string
...
+
+    /**
+     * Generate the Metadata XMP XML for PDF/A
+     *
+     * @return string
+     */
+    function getXmpMetadata()
+    {
+        $md = <<<EOT
+<?xpacket begin="ï»¿" id="W5M0MpCehiHzreSzNTczkc9d"?>
```
williamdes, line 3102: If someone changes the file encoding it may break
Can this seqence be converted to "\xEF\xBB\xBF"
I am unsure my hex is good

williamdes (approved):

lib/Cpdf.php
bsweeney: I believe that's right. Per the XMP spec the embedded XML "must be encoded as UTF-8." The document also handily indicates the BOM byte sequence to use for that encoding (0xEF 0xBB 0xBF).

wblessen:
> FYI, it's possible to achieve PDF/A compliance with the changes here. 

Hello Brian,  this is very good news.

> I'll add a page to the wiki outlining the requirements and provide sample code.

https://github.com/dompdf/dompdf/wiki/PDFA-Support

> To address your specific issue regarding fonts, you should set your default font by passing it in as part of the instantiation options and your document should not specify any fonts that would break compliance. For example, if you style your document with a generic font family then Dompdf will translate that into a core PDF font. So, for example, `font-family: sans-serif` is translated to mean "use the core PDF font Helvetica" which is not an embeddable font.

I think its perfectly fine, to add a hint to the wiki, 
that generic font families are not possible with PDF-A mode. 

You could even throw an Exception, IMO

bsweeney:
Indeed, PDFLib will throw exceptions for non-conforming elements. Soft failure does seem to be unsupportive of the end-user goal without at least a notification that the generated PDF will not be compliant. Perhaps it does make sense to throw an exception if a non-conforming element (e.g., CMYK element or non-embeddable font) is used.

Cruiser13:
Awesome, thanks a lot!

dimablack:
Does this library support Tagged PDF?

bsweeney:
Dompdf does not yet support tagged PDF. I'm looking to start working on accessibility features soon after the next release.

lib/res/sRGB2014.icc
williamdes: I am packaging this today, and the 2014 file is distributed on the non-free package: https://packages.debian.org/trixie/all/icc-profiles/filelist

Rather than the free file from https://packages.debian.org/forky/all/icc-profiles-free/filelist

I take the bet that the free one will work fine too

PR #3270, opened 2023-09-05
php 8.2 Fatal Error fix

wegas22:


src/Dompdf.php
```diff
@@ -432,6 +432,11 @@ public function loadHtml($str, $encoding = null)
         }
 
         if (in_array(strtoupper($encoding), array('UTF-8','UTF8')) === false) {
+
+						if($encoding == null) {
+							$encoding = 'UTF-8';
+						}
```
bsweeney, line 438: Not sure why `$encoding` would be null at this point. Just above here if it's null it's set to a value. Do you have sample usage that breaks here?

bsweeney (changes requested):

wegas22:
http://stankoqm.beget.tech/index.php?option=com_ajax&plugin=jpdfgenerator&group=system&format=raw

Закоментировал правку указывает на эту строчку посмотрите)

src/Dompdf.php
wegas22: http://stankoqm.beget.tech/index.php?option=com_ajax&plugin=jpdfgenerator&group=system&format=raw

Закоментировал правку указывает на эту строчку посмотрите)

bsweeney: What is the value of `$encoding` at this point? Like I said, it shouldn't be null since a check is performed earlier. And also, `strtoupper` on a null will also result in an error.

Are you passing an empty string to `loadHtml` for the encoding argument?

wegas22: ahh i get it

bsweeney: If it is an empty string issue then that's still a problem. We should probably use `if (empty($encoding))` instead of `if ($encoding === null)`.

Mellthas: > If it is an empty string issue then that's still a problem. We should probably use `if (empty($encoding))` instead of if `($encoding === null)`.

I think that makes sense, addressing as part of #3309.

PR #3275, opened 2023-09-12
Make all test providers static

williamdes:
Ref: https://sources.debian.org/src/php-dompdf/2.0.3+dfsg-2/debian/patches/0004-Adapt-for-phpunit-10.patch/

This has no backwards compatibility breaks.
But makes phpunit 10 happy

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       max-parallel: 12
       matrix:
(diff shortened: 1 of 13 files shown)


bsweeney:
Do you want to go ahead and make phpunit 10 allowed in the composer configuration?

williamdes:
Thank you !
I will update this into Debian on the next release

bsweeney:
Thank _you_ for ensuring our testing environment keeps up with current requirement.

PR #3300, opened 2023-10-23
Fix missing exception imports

Mellthas:
Otherwise the missing-table exceptions will not throw as intended ;-).


bsweeney:
Yikes, thanks for catching that 😮

PR #3304, opened 2023-10-26
Misc fixes for table rendering and hyperlink handling

Mellthas:
* **Small adjustments to table rendering**
	* Do not render borders for table row groups in the separated model. They should be ignored per spec [1]
	* Render outlines for table rows
	* Support table rows as named destinations and hyperlinks
	* Draw outlines for table cells instead of row groups when debugging layout

* **Renderer fixes for named destinations and hyperlinks**
	* Fix anchor elements with `name` attribute with display other than `inline`
	* Make hyperlinks with display `table-cell` and `table-row-group` work
	* List bullets cannot be named destinations

* **Image: Improve `alt` text handling**
	* Render image message below `alt` text
	* Exactly fit box if displaying `alt` text
	* Display `alt` text if it is `"0"`

[1] https://www.w3.org/TR/CSS21/tables.html#separated-borders

### DIFF
--- a/src/Renderer/Image.php
+++ b/src/Renderer/Image.php
@@ -68,17 +68,6 @@ class Image extends Block
             }
         }
 
-        if ($msg = $frame->get_image_msg()) {
-            $parts = preg_split("/\s*\n\s*/", $msg);
-            $font = $style->font_family;
-            $height = 10;
-            $offset = $alt !== "" ? $h : 0;
-
-            foreach ($parts as $i => $part) {
-                $this->_canvas->text($x, $y + $offset + $i * $height, $part, $font, $height * 0.8, [0.5, 0.5, 0.5]);
-            }
-        }
-
         $this->addNamedDest($node);
         $this->debugBlockLayout($frame, "blue");
     }
--- a/src/Renderer/Image.php
+++ b/src/Renderer/Image.php
@@ -36,11 +36,9 @@ class Image extends Block
         [$x, $y, $w, $h] = $content_box;
 
         $src = $frame->get_image_url();
-        $alt = null;
+        $alt = "";
 
-        if (Cache::is_broken($src) &&
-            $alt = $frame->get_node()->getAttribute("alt")
-        ) {
+        if (Cache::is_broken($src) && ($alt = $frame->get_node()->getAttribute("alt")) !== "") {
             $font = $style->font_family;
             $size = $style->font_size;
             $word_spacing = $style->word_spacing;
@@ -73,10 +71,10 @@ class Image extends Block
             $parts = preg_split("/\s*\n\s*/", $msg);
             $font = $style->font_family;
             $height = 10;
-            $_y = $alt ? $y + $h - count($parts) * $height : $y;
+            $offset = $alt !== "" ? $h : 0;
 
-            foreach ($parts as $i => $_part) {
-                $this->_canvas->text($x, $_y + $i * $height, $_part, $font, $height * 0.8, [0.5, 0.5, 0.5]);
+            foreach ($parts as $i => $part) {
+                $this->_canvas->text($x, $y + $offset + $i * $height, $part, $font, $height * 0.8, [0.5, 0.5, 0.5]);
             }
         }
 
--- a/src/Renderer/Image.php
+++ b/src/Renderer/Image.php
@@ -23,6 +23,7 @@ class Image extends Block
     function render(Frame $frame)
     {
         $style = $frame->get_style();
+        $node = $frame->get_node();
         $border_box = $frame->get_border_box();
 
         $this->_set_opacity($frame->get_opacity($style->opacity));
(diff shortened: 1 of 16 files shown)


src/Renderer/Image.php
```diff
@@ -73,18 +72,14 @@ function render(Frame $frame)
             $parts = preg_split("/\s*\n\s*/", $msg);
             $font = $style->font_family;
             $height = 10;
-            $_y = $alt ? $y + $h - count($parts) * $height : $y;
+            $offset = $alt !== "" ? $h : 0;
 
-            foreach ($parts as $i => $_part) {
-                $this->_canvas->text($x, $_y + $i * $height, $_part, $font, $height * 0.8, [0.5, 0.5, 0.5]);
+            foreach ($parts as $i => $part) {
+                $this->_canvas->text($x, $y + $offset + $i * $height, $part, $font, $height * 0.8, [0.5, 0.5, 0.5]);
```
bsweeney, line 78: I wonder if it's worth maintaining this bit of logic. Dompdf used to return specific messaging when an image couldn't be loaded, but all we do now is [return a generic message](https://github.com/dompdf/dompdf/blob/master/src/Image/Cache.php#L173).

src/Renderer/TableCell.php
```diff
@@ -62,12 +62,9 @@ function render(Frame $frame)
...
         }
 
-        $id = $frame->get_node()->getAttribute("id");
-        if (strlen($id) > 0) {
-            $this->_canvas->add_named_dest($id);
-        }
-
-        // $this->debugBlockLayout($frame, "red", false);
+        $this->addNamedDest($node);
+        $this->addHyperlink($node, $border_box);
```
bsweeney, line 66: To handle a scenario where the anchor element is styled `display: table-cell`?

bsweeney (approved):

src/Renderer/TableCell.php
Mellthas: Yes, a bit of an edge case, but it works in browsers, too.  Just noticed that it makes sense on table rows and row groups, as well.

Mellthas:
Sample document for hyperlinks styled as table-internal elements:

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
.table {
    display: table;
}

.tbody {
    display: table-row-group;
}

.tr {
    display: table-row;
}

.td {
    display: table-cell;
}
</style>
</head>

<body>
    <div class="table">
        <div class="tr">
            <a href="https://example.com" class="td">Link</a>
            <a href="https://example.com" class="td">Link</a>
            <a href="https://example.com" class="td">Link</a>
        </div>
        <a href="https://example.com" class="tr">
            <div class="td">Link</div>
            <div class="td">Link</div>
            <div class="td">Link</div>
        </a>
        <a href="https://example.com" class="tbody">
            <div class="tr">
                <div class="td">Link</div>
                <div class="td">Link</div>
                <div class="td">Link</div>
            </div>
            <div class="tr">
                <div class="td">Link</div>
                <div class="td">Link</div>
                <div class="td">Link</div>
            </div>
        </a>
    </div>
</body>

</html>
```

src/Renderer/TableCell.php
bsweeney: The things you can do...

src/Renderer/Image.php
Mellthas: True, displaying the message doesn’t seem that useful anymore. I pushed an extra commit just removing the logic. What do you think?

Mellthas: Sample document:

<details>
<summary>HTML</summary>

```html
<!DOCTYPE html>
<html>

<head>
<meta charset="UTF-8">
<style>
@page {
    size: 400pt 300pt;
    margin: 50pt;
}

p {
    margin: 0 0 10pt;
}

.i1 {
    width: 100%;
    height: 0;
    outline: 2pt solid green;
}

.i2 {
    width: 50%;
    height: 100pt;
    outline: 2pt solid green;
}

.i3 {
    width: 100%;
    height: 0;
    outline: 2pt solid green;
}
</style>
</head>

<body>
    <p><img src="invalid.png" class="i1" alt=""></p>
    <p><img src="invalid.png" class="i2" alt=""></p>
    <p><img src="invalid.png" class="i3" alt="Alternate text"></p>
    <p><img src="invalid.png" class="i3" alt="0"></p>
    <p><img src="invalid.png" class="i3" alt="Lorem ipsum dolor sit amet, consectetur adipisici elit, sed eiusmod tempor incidunt ut labore et dolore magna aliqua."></p>
</body>

</html>
```
</details>

Before: [before.pdf](https://github.com/dompdf/dompdf/files/13190774/before.pdf)
After: [after.pdf](https://github.com/dompdf/dompdf/files/13190773/after.pdf)
2.0.3 for comparison: [203.pdf](https://github.com/dompdf/dompdf/files/13190771/203.pdf)

bsweeney: Yeah, I think excluding it makes sense. It's a much cleaner look, particularly if alt text is provided.

Issue #1872, opened 2019-01-04
CSS variables


caliburn82:
I get problems when using CSS variables on margin or padding.

```css
:root {
    --spacing: 20px;
    --size: 300px;
    --grey: #888888;
}

.container {
    background: var(--grey-d);
    font-size: var(--spacing);
    margin: var(--spacing) 0; /* Errors: A non-numeric value encountered */
    padding: var(--spacing); /* Error: Box property calculation requires containing block width */
    width: var(--size);
}
```
I've been informed that dompdf doesn't support css variables, but would it be possible to add this to future revisions?

In the short term however, could error handling be added when variables are encounterd?

ElRochito:
Any news about this ?

bsweeney:
no news

jezzdk:
+1000000

Midorina:
This should be prioritized.

joelruizcabrera:
Is there anything new on this?

kgp43:
Would be nice to see this implemented indeed.

tmolitor-stud-tu:
I've tried to implement this. So far it works well for all my usecases, but it might have bugs, see my `ccs_vars` branch over here: https://github.com/tmolitor-stud-tu/dompdf/tree/ccs_vars

charliestrafe:
Is this issue fixed in the latest stable release? CSS variables still aren't working for me:
`:root {
    --navy: #1A2535;
}
body {
    color: var(--navy);
}
`

Thanks!

bsweeney:
The feature is not yet released. It will come out with Dompdf 3.0.0, which should come out soon.

PR #3314, opened 2023-11-02
Another take on implementing CSS variables (custom properties)

Neograph734:
I have always thought this would be difficult because of the time dompdf/dompdf#1872 is already open, but thanks tmolitor-stud-tu for showing how easy it actually is in dompdf/dompdf#3255. 

The approach there was a great start, but it was missing the possibility to use fallback values as a second parameter of var(). Furthermore I have shuffled some assignments around and wrote tests. Please share your thoughts.

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -1520,7 +1520,7 @@ class Style
 
                 // If this property is depending on another value, clear the
                 // computed value so that it can be used later.
-                if (self::$_dependent_props[$prop]) {
+                if (!empty(self::$_dependent_props[$prop])) {
                     unset($this->_props_computed[$prop]);
                     unset($this->_props_used[$prop]);
                 }
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -1586,7 +1586,7 @@ class Style
     public function set_prop(string $prop, $val, bool $important = false, bool $clear_dependencies = true): void
     {
         // Skip some checks for CSS custom properties.
-        if (!(\strpos($prop, '--') === 0)) {
+        if (!$this->is_custom_property($prop)) {
 
             $prop = str_replace("-", "_", $prop);
 
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -1390,7 +1390,7 @@ class Style
      */
     public function set_prop(string $prop, $val, bool $important = false, bool $clear_dependencies = true): void
     {
-		// Store CSS custom property as-is and leave.
+        // Store CSS custom property as-is and leave.
         if (substr($prop, 0, 2) === "--") {
             $this->_props[$prop] = trim($val);
             return;
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -1231,7 +1231,7 @@ class Style
         $variable = is_array($matches) ? $matches[1] : $matches;
 
         // Split property name and an optional fallback value.
-        [$custom_prop, $fallback] = explode(',', $variable, 2) + [null, null];
+        [$custom_prop, $fallback] = explode(',', $variable, 2) + ['', ''];
         $fallback = trim($fallback);
 
         // Try to retrieve the custom property value, or use the fallback value
@@ -1269,28 +1269,18 @@ class Style
         if ($parent) {
             // For properties that inherit by default: When the cascade did
             // not result in a value, inherit the parent value. Inheritance
-            // is handled via the specific sub-properties for shorthands
+            // is handled via the specific sub-properties for shorthands. Custom
+            // properties (variables) are selected by the -- prefix.
             foreach ($parent->_props as $prop => $val) {
-                if (!isset($this->_props[$prop]) && isset(self::$_inherited[$prop])) {
+                if (!isset($this->_props[$prop]) && (
+                    isset(self::$_inherited[$prop])
+                    || substr($prop, 0, 2) === "--"
+                )) {
                     $parent_val = $parent->computed($prop);
 
                     $this->_props[$prop] = $parent_val;
                     $this->_props_computed[$prop] = $parent_val;
                     $this->_props_used[$prop] = null;
-                }
-            }
-
-            // Iterate over the parent properties and inherit all the custom
-            // properties that are not defined on the own element.
-            foreach ($parent->_props as $parent_prop => $parent_val) {
-                if (substr($parent_prop, 0, 2) === "--"
-                    && !isset($this->_props[$parent_prop])
-                ) {
-                    $parent_val = $parent->computed($parent_prop);
-
-                    $this->_props[$parent_prop] = $parent_val;
-                    $this->_props_computed[$parent_prop] = $parent_val;
-                    $this->_props_used[$parent_prop] = null;
                 }
             }
         }
@@ -1493,8 +1483,7 @@ class Style
             // resolve this when the value is needed. Just set the property now.
             if (is_string($val) && strpos($val, 'var(') !== false) {
                 $this->_props[$prop] = $val;
-                self::$_dependent_props[$prop] = true;
-                //return;
+                return;
             }
 
             $computed = $this->compute_prop($prop, $val);
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -633,6 +633,14 @@ class Style
     protected $_props = [];
 
     /**
+     * Used to track which CSS property were set directly versus
+     * those set via shorthand property
+     *
+     * @var array<string, true>
+     */
+    protected $_props_specified = [];
+
+    /**
      * Computed values of the CSS properties.
      *
      * @var array<string, mixed>
@@ -653,6 +661,22 @@ class Style
      * @var array<string, true>
      */
     protected $non_final_used = [];
+
+    /**
+     * Used to track CSS property assignment entry/exit. Used to watch
+     * for circular dependencies.
+     *
+     * @var array<int, string>
+     */
+    protected $_prop_stack = [];
+
+    /**
+     * Used to track CSS variable resolution entry/exit. Used to watch
+     * for circular dependencies
+     *
+     * @var array<int, string>
+     */
+    protected $_var_stack = [];
 
     /**
      * Style of the parent element in document tree.
@@ -1230,14 +1254,18 @@ class Style
     private function parse_var($matches) {
         $variable = is_array($matches) ? $matches[1] : $matches;
 
+        if (\in_array($variable, $this->_var_stack, true)) {
+            return null;
+        }
+        array_push($this->_var_stack, $variable);
+
         // Split property name and an optional fallback value.
         [$custom_prop, $fallback] = explode(',', $variable, 2) + ['', ''];
         $fallback = trim($fallback);
 
         // Try to retrieve the custom property value, or use the fallback value
-        // if the value could not be resolved. Suppress 'undefined array key'
-        // errors from undefined properties.
-        $value = @$this->computed($custom_prop) ?? $fallback;
+        // if the value could not be resolved.
+        $value = $this->computed($custom_prop) ?? $fallback;
 
         // If the resolved value also has vars in it, resolve again.
         $pattern = self::CSS_VAR;
@@ -1246,6 +1274,7 @@ class Style
             [$this, "parse_var"],
             $value);
 
+        array_pop($this->_var_stack);
         return $value ?: null;
     }
 
@@ -1272,21 +1301,28 @@ class Style
             // is handled via the specific sub-properties for shorthands. Custom
             // properties (variables) are selected by the -- prefix.
             foreach ($parent->_props as $prop => $val) {
-                if (!isset($this->_props[$prop]) && (
-                    isset(self::$_inherited[$prop])
-                    || substr($prop, 0, 2) === "--"
-                )) {
+                if (
+                    !isset($this->_props[$prop])
+                    && (
+                        isset(self::$_inherited[$prop])
+                        || substr($prop, 0, 2) === "--"
+                    )
+                ) {
                     $parent_val = $parent->computed($prop);
 
-                    $this->_props[$prop] = $parent_val;
-                    $this->_props_computed[$prop] = $parent_val;
-                    $this->_props_used[$prop] = null;
+                    if (substr($prop, 0, 2) === "--") {
+                        $this->set_prop($prop, $parent_val);
+                    } else {
+                        $this->_props[$prop] = $parent_val;
+                        $this->_props_computed[$prop] = $parent_val;
+                        $this->_props_used[$prop] = null;
+                    }
                 }
             }
         }
 
         foreach ($this->_props as $prop => $val) {
-            if ($val === "inherit") {
+            if ($val === "inherit" && substr($prop, 0, 2) !== "--") {
                 if ($parent && isset($parent->_props[$prop])) {
                     $parent_val = $parent->computed($prop);
 
@@ -1322,7 +1358,11 @@ class Style
                 $this->_important_props[$prop] = true;
             }
 
-            $this->_props[$prop] = $val;
+            if (substr($prop, 0, 2) === "--") {
+                $this->set_prop($prop, $val, $important);
+            } else {
+                $this->_props[$prop] = $val;
+            }
 
             // Copy an existing computed value only for non-dependent
             // properties; otherwise it may be invalid for the current style
@@ -1334,6 +1374,10 @@ class Style
             } else {
                 unset($this->_props_computed[$prop]);
                 unset($this->_props_used[$prop]);
+            }
+
+            if (\array_key_exists($prop, $style->_props_specified)) {
+                $this->_props_specified[$prop] = true;
             }
         }
     }
@@ -1406,6 +1450,7 @@ class Style
                 return;
             }
         }
+        $this->_props_specified[$prop] = true;
 
         // Trim declarations unconditionally, but only lower-case for comparison
         // with the general keywords. Properties must handle case-insensitive
@@ -1429,6 +1474,29 @@ class Style
             } else {
                 $method = "_set_$prop";
 
+                // Resolve the CSS custom property value(s).
+                $pattern = self::CSS_VAR;
+
+                // Always set the specified value for properties that use CSS variables
+                // so that an invalid initial value does not prevent re-computation later.
+                $this->_props[$prop] = $val;
+
+                //TODO: we shouldn't need to parse this twice
+                preg_match_all("/$pattern/", $val, $matches, PREG_SET_ORDER);
+                foreach ($matches as $match) {
+                    if ($this->parse_var($match) === null) {
+                        // unset specified as for specific prop under expectation it will be overridden
+                        foreach (self::$_props_shorthand[$prop] as $sub_prop) {
+                            unset($this->_props_specified[$sub_prop]);
+                        }
+                        return;
+                    }
+                }
+                $val = preg_replace_callback(
+                    "/$pattern/",
+                    [$this, "parse_var"],
+                    $val);
+
                 if (!isset(self::$_methods_cache[$method])) {
                     self::$_methods_cache[$method] = method_exists($this, $method);
                 }
@@ -1445,6 +1513,7 @@ class Style
                     foreach (self::$_props_shorthand[$prop] as $sub_prop) {
                         $sub_val = $values[$sub_prop] ?? self::$_defaults[$sub_prop];
                         $this->set_prop($sub_prop, $sub_val, $important, $clear_dependencies);
+                        unset($this->_props_specified[$sub_prop]);
                     }
                 }
             }
@@ -1477,13 +1546,10 @@ class Style
                 $val = self::$_defaults[$prop];
             }
 
-            // Do not resolve CSS custom property values when setting them,
-            // because the required custom property might not have been set yet,
-            // or need to be derived from a parent. Let the compute() method
-            // resolve this when the value is needed. Just set the property now.
-            if (is_string($val) && strpos($val, 'var(') !== false) {
+            // Always set the specified value for properties that use CSS variables
+            // so that an invalid initial value does not prevent re-computation later.
+            if (\is_string($val) && strpos($val, 'var(') !== false) {
                 $this->_props[$prop] = $val;
-                return;
             }
 
             $computed = $this->compute_prop($prop, $val);
@@ -1496,6 +1562,27 @@ class Style
             $this->_props[$prop] = $val;
             $this->_props_computed[$prop] = $computed;
             $this->_props_used[$prop] = null;
+
+            //TODO: this should be a directed dependency map
+            if (substr($prop, 0, 2) === "--" && !\in_array($prop, $this->_prop_stack, true)) {
+                array_push($this->_prop_stack, $prop);
+                $specified_props = array_filter($this->_props, function($key) {
+                    return \array_key_exists($key, $this->_props_specified);
+                }, ARRAY_FILTER_USE_KEY); // copy existing props filtered by those set explicitly before parsing vars
+                foreach ($specified_props as $specified_prop => $specified_value) {
+                    if (substr($specified_prop, 0, 2) !== "--" || strpos($specified_value, "var($prop") !== false) {
+                        $this->set_prop($specified_prop, $specified_value, \in_array($specified_prop, $this->_important_props, true), true);
+                        if (isset(self::$_props_shorthand[$specified_prop])) {
+                            foreach (self::$_props_shorthand[$specified_prop] as $sub_prop) {
+                                if (\array_key_exists($sub_prop, $specified_props)) {
+                                    $this->set_prop($sub_prop, $specified_props[$sub_prop], \in_array($sub_prop, $this->_important_props, true), true);
+                                }
+                            }
+                        }
+                    }
+                }
+                array_pop($this->_prop_stack);
+            }
 
             if ($clear_dependencies) {
                 // Clear the computed values of any dependent properties, so
@@ -1719,8 +1806,18 @@ class Style
     protected function computed(string $prop)
     {
         if (!\array_key_exists($prop, $this->_props_computed)) {
+            if (!\array_key_exists($prop, $this->_props) && substr($prop, 0, 2) === "--") {
+                return null;
+            }
             $val = $this->_props[$prop] ?? self::$_defaults[$prop];
             $computed = $this->compute_prop($prop, $val);
+
+            if ($computed === null) {
+                if (substr($prop, 0, 2) === "--") {
+                    return null;
+                }
+                $computed = $this->compute_prop($prop, self::$_defaults[$prop]);
+            }
 
             $this->_props_computed[$prop] = $computed;
         }
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -1012,10 +1012,10 @@ class Style
         return $this->_stylesheet;
     }
 
-  public function is_custom_property(string $prop): bool
-  {
-      return \substr($prop, 0, 2) === "--";
-  }
+    public function is_custom_property(string $prop): bool
+    {
+        return \substr($prop, 0, 2) === "--";
+    }
 
     public function is_absolute(): bool
     {
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -1012,6 +1012,11 @@ class Style
         return $this->_stylesheet;
     }
 
+  public function is_custom_property(string $prop): bool
+  {
+      return \substr($prop, 0, 2) === "--";
+  }
+
     public function is_absolute(): bool
     {
         $position = $this->__get("position");
@@ -1443,12 +1448,12 @@ class Style
                     !isset($this->_props[$prop])
                     && (
                         isset(self::$_inherited[$prop])
-                        || substr($prop, 0, 2) === "--"
+                        || $this->is_custom_property($prop)
                     )
                 ) {
                     $parent_val = $parent->computed($prop);
 
-                    if (substr($prop, 0, 2) === "--") {
+                    if ($this->is_custom_property($prop)) {
                         $this->set_prop($prop, $parent_val);
                     } else {
                         $this->_props[$prop] = $parent_val;
@@ -1460,7 +1465,7 @@ class Style
         }
 
         foreach ($this->_props as $prop => $val) {
-            if ($val === "inherit" && substr($prop, 0, 2) !== "--") {
+            if ($val === "inherit" && !$this->is_custom_property($prop)) {
                 if ($parent && isset($parent->_props[$prop])) {
                     $parent_val = $parent->computed($prop);
 
@@ -1496,7 +1501,7 @@ class Style
                 $this->_important_props[$prop] = true;
             }
 
-            if (substr($prop, 0, 2) === "--") {
+            if ($this->is_custom_property($prop)) {
                 $this->set_prop($prop, $val, $important);
             } else {
                 $this->_props[$prop] = $val;
@@ -1521,7 +1526,7 @@ class Style
 
         // re-evalutate CSS variables
         foreach (array_keys($this->_props) as $prop) {
-            if (substr($prop, 0, 2) !== "--") {
+            if (!$this->is_custom_property($prop)) {
                 continue;
             }
             $this->set_prop($prop, $this->_props[$prop], isset($this->_important_props[$prop]));
@@ -1581,7 +1586,7 @@ class Style
     public function set_prop(string $prop, $val, bool $important = false, bool $clear_dependencies = true): void
     {
         // Skip some checks for CSS custom properties.
-        if (substr($prop, 0, 2) !== "--") {
+        if (!(\strpos($prop, '--') === 0)) {
 
             $prop = str_replace("-", "_", $prop);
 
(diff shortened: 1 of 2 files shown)


Neograph734:
Funny how the tests do pass on my local machine :/

Mellthas:
> Funny how the tests do pass on my local machine :/

CI failed with undefined index notices on PHP 7.3, here’s the output:

<details>
<summary>There were 6 errors</summary>

```
1) Dompdf\Tests\Css\StyleTest::testVar with data set "invalid_value" ('Helvetica', 'var(--undefined)', 'Helvetica')
Undefined index: --undefined

/home/runner/work/dompdf/dompdf/src/Css/Style.php:1235
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1244
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1694
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1723
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1661
/home/runner/work/dompdf/dompdf/tests/Css/StyleTest.php:1083
phpvfscomposer:///home/runner/work/dompdf/dompdf/vendor/phpunit/phpunit/phpunit:106

2) Dompdf\Tests\Css\StyleTest::testInheritedVar with data set "green" (0, '#00ff00FF')
Undefined index: --fallback-propert

/home/runner/work/dompdf/dompdf/src/Css/Style.php:1235
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1250
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1694
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1723
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1283
/home/runner/work/dompdf/dompdf/src/Css/Stylesheet.php:1069
/home/runner/work/dompdf/dompdf/src/Dompdf.php:731
/home/runner/work/dompdf/dompdf/tests/Css/StyleTest.php:1164
phpvfscomposer:///home/runner/work/dompdf/dompdf/vendor/phpunit/phpunit/phpunit:106

3) Dompdf\Tests\Css\StyleTest::testInheritedVar with data set "red" (1, '#ff0000FF')
Undefined index: --fallback-propert

/home/runner/work/dompdf/dompdf/src/Css/Style.php:1235
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1250
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1694
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1723
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1283
/home/runner/work/dompdf/dompdf/src/Css/Stylesheet.php:1069
/home/runner/work/dompdf/dompdf/src/Dompdf.php:731
/home/runner/work/dompdf/dompdf/tests/Css/StyleTest.php:1164
phpvfscomposer:///home/runner/work/dompdf/dompdf/vendor/phpunit/phpunit/phpunit:106

4) Dompdf\Tests\Css\StyleTest::testInheritedVar with data set "yellow" (2, '#ffff00FF')
Undefined index: --fallback-propert

/home/runner/work/dompdf/dompdf/src/Css/Style.php:1235
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1250
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1694
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1723
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1283
/home/runner/work/dompdf/dompdf/src/Css/Stylesheet.php:1069
/home/runner/work/dompdf/dompdf/src/Dompdf.php:731
/home/runner/work/dompdf/dompdf/tests/Css/StyleTest.php:1164
phpvfscomposer:///home/runner/work/dompdf/dompdf/vendor/phpunit/phpunit/phpunit:106

5) Dompdf\Tests\Css\StyleTest::testInheritedVar with data set "blue" (3, '#0000ffFF')
Undefined index: --fallback-propert

/home/runner/work/dompdf/dompdf/src/Css/Style.php:1235
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1250
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1694
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1723
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1283
/home/runner/work/dompdf/dompdf/src/Css/Stylesheet.php:1069
/home/runner/work/dompdf/dompdf/src/Dompdf.php:731
/home/runner/work/dompdf/dompdf/tests/Css/StyleTest.php:1164
phpvfscomposer:///home/runner/work/dompdf/dompdf/vendor/phpunit/phpunit/phpunit:106

6) Dompdf\Tests\Css\StyleTest::testInheritedVar with data set "purple" (4, '#ff00ffFF')
Undefined index: --fallback-propert

/home/runner/work/dompdf/dompdf/src/Css/Style.php:1235
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1250
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1694
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1723
/home/runner/work/dompdf/dompdf/src/Css/Style.php:1283
/home/runner/work/dompdf/dompdf/src/Css/Stylesheet.php:1069
/home/runner/work/dompdf/dompdf/src/Dompdf.php:731
/home/runner/work/dompdf/dompdf/tests/Css/StyleTest.php:1164
phpvfscomposer:///home/runner/work/dompdf/dompdf/vendor/phpunit/phpunit/phpunit:106
```
</details>

When running locally on PHP 8.2, I get several deprecation notices (need to pass `--display-deprecations` when using PHPUnit 10).

bsweeney:
Approved running the tests so you can see any failures as you make updates.

src/Css/Stylesheet.php
```diff
@@ -1609,7 +1609,7 @@ private function _parse_font_face($str)
      */
     private function _parse_properties($str)
     {
-        $properties = preg_split("/;(?=(?:[^\(]*\([^\)]*\))*(?![^\)]*\)))/", $str);
+        $properties = preg_split("/;(?=(?:[^\(]*\((.*)?[^\)]*\))*(?![^\)]*\)))/", $str);
```
bsweeney, line 1612: Is this change necessary? It looks like you're capturing the content of the potential CSS function but not doing anything with it?

src/Css/Style.php
```diff
@@ -1640,6 +1688,12 @@ protected function compute_prop(string $prop, $val)
             return $val;
         }
 
+        // Resolve CSS custom property value.
+        if (substr($val, 0, 4) === 'var(') {
+            $var = substr(trim($val), 4, -1);
+            return $this->parse_var($var);
```
bsweeney, line 1694: We should allow normal computation after this so we can accommodate something like:

```
<style>
    :root {
        --half: 50%;
    }
    div {
        width: var(--half);
    }
</style>
<div>dompdf</div>
```

src/Css/Style.php
```diff
@@ -1640,6 +1688,12 @@ protected function compute_prop(string $prop, $val)
...
+        }
```
bsweeney, line 1695: This assumes the entirety of the value is a variable reference. But that may not be true, for example:

```
<style>
    :root {
        --bg: orange;
    }
    div {
        border: 2px solid var(--bg);
    }
</style>
<div>dompdf</div>
```

bsweeney (changes requested):
Some initial thoughts for discussion. I haven't thoroughly evaluated this in relation to the spec nor in relation to how we generally parse CSS properties.

src/Css/Style.php
```diff
@@ -1220,6 +1220,39 @@ private function evaluate_func_calc(array $rpn, float $ref_size = 0, ?float $fon
...
+        if (is_string($value) && substr($value, 0, 4) === 'var(') {
+            $var = substr(trim($value), 4, -1);
+            $value = $this->parse_var($var);
+        }
+
+        // If the resolved fallback is a variable, resolve that too.
+        if (is_string($fallback) && substr($fallback, 0, 4) === 'var(') {
+            $var = substr(trim($fallback), 4, -1);
+            $fallback = $this->parse_var($var);
+        }
```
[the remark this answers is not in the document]
Mellthas: Relevant section from the spec: https://www.w3.org/TR/css-variables-1/#cycles

src/Css/Style.php
```diff
@@ -1220,6 +1220,39 @@ private function evaluate_func_calc(array $rpn, float $ref_size = 0, ?float $fon
...
+        // Split property name and an optional fallback value.
+        [$custom_prop, $fallback] = explode(',', $variable, 2) + [null, null];
+        $fallback = trim($fallback);
+
+        // Recursively try to resolve the property value through all parents.
+        $value = $this->_props[$custom_prop] ?? (
+            isset($this->parent_style)
+                ? $this->parent_style->parse_var($variable)
+                : null
+        );
```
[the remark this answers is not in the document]
Neograph734: I'll look into that. Sound like the better way to go indeed.

src/Css/Style.php
```diff
@@ -1361,6 +1394,12 @@ protected function clear_cache(string $prop): void
      */
     public function set_prop(string $prop, $val, bool $important = false, bool $clear_dependencies = true): void
     {
+        // Store CSS custom property as-is and leave.
+        if (substr($prop, 0, 2) === "--") {
+            $this->_props[$prop] = trim($val);
+            return;
+        }
```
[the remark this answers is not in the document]
Neograph734: In the lines below this block, all dashes are converted to underscores. This would need to get an exception too because `--custom-var` and `--custom_var` are both valid, but are two different properties. I am not sure if the getters accept dashes.

src/Css/Stylesheet.php
Neograph734: I am not a hero with regex, so I might have done it wrong. But I made chis change to get `var(var(...))` to work. Without this change all properties became one big property and everything broke. Arguably this is still not right because `var(var(var(...)))` might still fail, but I don't really know how realistic that is.

src/Css/Style.php
Neograph734: True, or with 2 vars:  `border: 2px vra(--border-style) var(--bg);`

src/Css/Stylesheet.php
bsweeney: Let me think on that, then. RegEx isn't great for parsing out nested patterns like that so it may be hard to fully accommodate.

src/Css/Style.php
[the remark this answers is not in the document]
bsweeney: The purpose of swapping out dash for underscore is so that the getter/setter can be used at all, since dashes aren't valid identifiers. In this way we can get the `font-face` value by accessing the font_face property (`$style->font_face`).

But you're right that both dash and underscore are permitted since the custom property name is defined as a [dashed identity](https://drafts.csswg.org/css-values-4/#typedef-dashed-ident). So, skipping replacement makes sense. And any "get" would just have to use the less convenient `$style->__get("--custom-prop");`.

I do still think we should continue processing rather than set and return, with some additional handling where needed.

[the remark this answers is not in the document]
Neograph734: I think this is resolved. 

        // Skip some checks for CSS custom properties.
        if (substr($prop, 0, 2) !== "--") {

            $prop = str_replace("-", "_", $prop);

            // Legacy property aliases
            if (isset(self::$_props_alias[$prop])) {
                $prop = self::$_props_alias[$prop];
            }

            if (!isset(self::$_defaults[$prop])) {
                global $_dompdf_warnings;
                $_dompdf_warnings[] = "'$prop' is not a recognized CSS property.";
                return;
            }
        }

src/Css/Style.php
Neograph734: Resolved by using preg_replace_callback().

src/Css/Style.php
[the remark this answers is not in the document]
Neograph734: Implemented inheritance.

src/Css/Style.php
Neograph734: The property is parsed before the value is computed, so I think this should work... Will add a test for it though.

src/Css/Style.php
```diff
@@ -1446,6 +1485,15 @@ public function set_prop(string $prop, $val, bool $important = false, bool $clea
...
             }
 
+            // Do not resolve CSS custom property values when setting them,
+            // because the required custom property might not have been set yet,
+            // or need to be derived from a parent. Let the compute() method
+            // resolve this when the value is needed. Just set the property now.
+            if (is_string($val) && substr($val, 0, 4) === 'var(') {
+                $this->_props[$prop] = $val;
+                return;
+            }
```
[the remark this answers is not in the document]
bsweeney: `compute_prop` doesn't necessarily run for nothing. It's just that it will re-calculate if the custom property changes. One thing I think helps when looking at the code is consistency in implementation and this is how other properties are handled.

On the other hand, resolving the value when it is retrieved would probably save us a few cpu cycles, primarily because inheritance isn't performed until later. I could be convinced that it's a good idea to skip initial computation for properties that utilize the CSS variables. @Mellthas may have thoughts on this.

src/Css/Style.php
bsweeney: I think we're good now.

src/Css/Style.php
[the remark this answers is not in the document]
Neograph734: I would say that a `var()` by definition is depending on another property. This might be on the same style, but most likely this will have to be inherited because its value is defined on `:root` or some other element. 

But I'll have a look at creating a proper dependency map in the meantime.

Neograph734:
Awsome, those 3 tests fail locally as well because the border shorthand property is never running `compute_prop()`. That's enough for today.

src/Css/Style.php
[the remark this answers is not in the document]
Mellthas: Our current logic does syntax checking and computation as a combined step. I think it makes sense to defer computation to later here. From the spec:

> If a property contains one or more [var()](https://www.w3.org/TR/css-variables-1/#funcdef-var) functions, and those functions are syntactically valid, the entire property’s grammar must be assumed to be valid at parse time. It is only syntax-checked at computed-value time, after var() functions have been [substituted](https://www.w3.org/TR/css-variables-1/#substitute-a-var).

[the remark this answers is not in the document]
Mellthas: The check here also needs to account for `var()` occurring anywhere within the value.

src/Css/Style.php
```diff
@@ -1446,6 +1493,16 @@ public function set_prop(string $prop, $val, bool $important = false, bool $clea
                 $val = self::$_defaults[$prop];
             }
 
+            // Do not resolve CSS custom property values when setting them,
+            // because the required custom property might not have been set yet,
+            // or need to be derived from a parent. Let the compute() method
+            // resolve this when the value is needed. Just set the property now.
+            if (is_string($val) && strpos($val, 'var(') !== false) {
+                $this->_props[$prop] = $val;
+                self::$_dependent_props[$prop] = true;
```
Mellthas, line 1502: Since the `var` dependencies are style-specific, using the static map here does not work correctly.

src/Css/Style.php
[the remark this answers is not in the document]
bsweeney: I agree, some more thought on dependency handling is probably required. Though honestly when I attempted to trigger an invalid resolution by modifying the value of a custom property in later declarations it always came back the expected value. Still, we should be handling that dependency in case it's needed.

src/Css/Style.php
bsweeney, line 1502: Flagging the value as being dependent doesn't really seem to do anything by itself and removing from the code doesn't seem to have any impact. Further comments made in the previous discussion.

src/Css/Style.php
```diff
@@ -1446,6 +1493,16 @@ public function set_prop(string $prop, $val, bool $important = false, bool $clea
                 $val = self::$_defaults[$prop];
             }
 
+            // Do not resolve CSS custom property values when setting them,
+            // because the required custom property might not have been set yet,
+            // or need to be derived from a parent. Let the compute() method
+            // resolve this when the value is needed. Just set the property now.
+            if (is_string($val) && strpos($val, 'var(') !== false) {
+                $this->_props[$prop] = $val;
```
bsweeney, line 1501: Setting the specified value here is problematic due to assumptions made elsewhere in the code. Normally when a property has an invalid computation we drop the specified value entirely. Which is fine since the value must be valid when set, and the default value is used otherwise.

CSS variables may not be valid at the time they are set. Or may contain invalid values once they are resolved. But we want to maintain the specified value in case the variable resolves correctly later on.

The following example will trigger an error because when Dompdf retrieves the font size it returns an unexpected value (null):

```
<style>
    div {
        font-size: var(--bad);
    }
</style>
<div>dompdf</div>
```

I think the resolution might be to modify the `computed` method to recompute the property if computing with the specified value returns null.

bsweeney (changes requested):

src/Css/Style.php
```diff
@@ -1254,6 +1283,20 @@ public function inherit(?Style $parent = null): void
                     $this->_props_used[$prop] = null;
                 }
             }
+
+            // Iterate over the parent properties and inherit all the custom
+            // properties that are not defined on the own element.
+            foreach ($parent->_props as $parent_prop => $parent_val) {
```
Mellthas, line 1289: It think it makes sense to combine this loop with the one above:

* Iterate over all parent props.
* The shorthand check can go, since shorthand values are not stored in `_props`.
* Skip, if the property is set on the current element, does not inherit by default, or is not a custom property.
* `Style::$_inherited` could be turned into a map (`[$prop => true]`) to make the inherit-by-default check more efficient.

Mellthas: I have done the preparatory work in #3322. With that, combining the loops should be trivial.

src/Css/Style.php
bsweeney: The cited change doesn't work on shorthand properties because the CSS variable isn't resolved prior to applying the value to the more specific CSS properties. In this particular example since `var(--bg)` isn't resolved it doesn't match a valid border color value and so the entire declaration is dropped.

First thing to do would be to add variable resolution for shorthand properties. If we do that in `set_prop` before calling the shorthand method then the variable reference parses.

The next thing to do is more of a challenge. If the variable value changes after the shorthand has been parsed we need to call `set_prop` on that shorthand again to propagate the value down to the specific CSS properties. However, if we did a straight up variable dependency, then something like the following:

```
<style>
    div {
        --color: orange;
        border: 2px solid var(--color);
        border-color: blue;
    }
</style>
```

will get the wrong result because re-processing the `border` property will overwrite what was set by `border-color`.

Perhaps the easiest solution to this scenario would be to completely re-process the standard properties that have been set whenever a custom property value is set. It's a bit inefficient but I did some quick testing and it works.

bsweeney (changes requested):

src/Css/Style.php
bsweeney: > The shorthand check can go, since shorthand values are not stored in _props.

I think we're going to have to start storing the shorthand properties so that we can re-calculate when the value of a custom property changes. I don't see an obvious way to make this work otherwise, because you don't know until after the variable has resolved which specific property it applies to.

src/Css/Stylesheet.php
bsweeney: This does work, so good job. I believe you can exclude the capturing group since it's not being used:

```suggestion
        $properties = preg_split("/;(?=(?:[^\(]*\(.*?[^\)]*\))*(?![^\)]*\)))/", $str);
```

src/Css/Style.php
bsweeney: Reference the following sample:

```
<style>
    :root {
        --color: orange;
    }
    div {
        font-size: 150px;
        border: 2px solid var(--color);
    }
</style>
<div>dompdf</div>
```

Since the custom property isn't defined until inheritance is applied the border property parses as invalid.

In order to work around the issue I had to add two steps:

1. pre-processing variables before calling the shorthand method
2. when a custom property is set re-compute all standard properties in the style

The latter, while non-optimal from a performance perspective, is necessary because a specific property is overridden when re-calculating any variable-dependent properties. Such as in the following:

```
<style>
    :root {
        --color: orange;
    }
    div {
        font-size: 150px;
        border: 2px solid var(--color);
        border-left-color: black;
    }
</style>
<div>dompdf</div>
```

I'm happy to push the adjustments I tried locally that appear to resolve current issues.

src/Css/Stylesheet.php
Neograph734: Ah yes, of course. Good idea.

Neograph734:
Thanks for all the reviews, but I am getting the feeling that the different conversations are getting mixed (and duplicated). So I am struggling to keep up... 

> I'm happy to push the adjustments I tried locally that appear to resolve current issues.
Feel free to commit what you have, I believe you have the required permissions. I don't want to break anything of your work in an attempt to close some of the above threads.

src/Css/Style.php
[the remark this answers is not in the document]
Neograph734: Based on Melthias comment above I assume we are (for now) fine with setting the value and returning. Marking resolved as we can follow on dependency handline in another thread,

src/Css/Style.php
Neograph734: Ah, I thought the static would live within each style. I did not realise it would apply to all of them. Reverted.

src/Css/Style.php
```diff
@@ -1220,6 +1221,34 @@ private function evaluate_func_calc(array $rpn, float $ref_size = 0, ?float $fon
...
+     *
+     * This function receives the whole content of the var() function, which
+     * can also include a fallback value.
+     */
+    private function parse_var($matches) {
+        $variable = is_array($matches) ? $matches[1] : $matches;
+
+        // Split property name and an optional fallback value.
+        [$custom_prop, $fallback] = explode(',', $variable, 2) + [null, null];
+        $fallback = trim($fallback);
```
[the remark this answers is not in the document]
Neograph734: Indeed. Fixed and resolved.

src/Css/Style.php
Neograph734: Reverted to setting the value only and then return. Resolving this thread because there are more extended threads regarding dependencies.

src/Css/Style.php
Neograph734: Combined the inheritance logic. Leaving open for Brian's last comment.

bsweeney:
> Thanks for all the reviews, but I am getting the feeling that the different conversations are getting mixed (and duplicated). So I am struggling to keep up...

Yes, I noticed that. My apologies. I dunno what happened I think I somehow started a new review when I was commenting on another thread. And then I made a similar comment twice. 😞 

> > I'm happy to push the adjustments I tried locally that appear to resolve current issues.
> 
> Feel free to commit what you have, I believe you have the required permissions. I don't want to break anything of your work in an attempt to close some of the above threads.

I'll make those changes and resolve a few conversations.

src/Css/Stylesheet.php
Mellthas: Hm, from reading the pattern, it looks like it is supposed to detect a single closing parenthesis (with balanced parentheses in-between), so that a semicolon can be used in a `url()` function, e.g. `url(123;45.png)`; are there other cases where a semicolon makes sense, obviously outside a string? Based on that, I think it would be better to simplify the pattern radically (pattern with added white space for readability):

```
"/; (?! [^\(]* \) )/x"
```

So only try to detect a lone closing parenthesis after the semicolon, and don’t bother with trying to detect balanced pairs. I have prepared a PR based on that pattern.

Mellthas: See #3323

src/Css/Style.php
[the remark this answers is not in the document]
Neograph734: Depending on how the dependency issue will be solved, we might be able to use the dependency map to detect circulair reference?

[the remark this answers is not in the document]
Neograph734: The more I think of it, the less it makes sense to work with dependencies. The whole fallback thing is there because a custom property might not have been defined at all. Using with dependencies will not work and properties that have a falid fallback might get dropped as invalid/incomplete. 

I suppose the simplest way to keep the recursion under control is adding something like this?

    if (count(debug_backtrace()) > 99) {
       // Add error message to global var.
       return;
    }

Or you want to add the max depth as an option in Dompdf/Options? PHP does not enfore a max recursion depth, though Xdebug stops at 100. Can we assume that people test their implementation? i.e. The performance impact of stopping the recursion relatively late should not impact a production system? 
If we stop at a lower version to save performance we might break a valid setup in a framework with lots of calls?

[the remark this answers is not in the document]
bsweeney: For the recursion check I'm not using a dependency map. I added a stack to track which variables have been processed. If we encounter the same custom property reference more than once the method bails. I'm still thinking through fallback handling before I check something in for review.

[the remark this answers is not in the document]
Neograph734: Oké, Thanks. I know we spoke about dependencies in other threads as well, but this was the most relevant open one.

I just suddenly realized how strict dependency checking could result in unexpected behavior for undefined properties, which I wanted to write down somewhere.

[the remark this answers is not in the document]
bsweeney: The spec does dictate a "directed dependency graph" which this is most certainly not. Going that route would require a bit more effort that I'm able to dedicate at the moment. Though that level of implementation would probably provide some relief from the issues I'm trying to work around right now.

bsweeney:
OK I've added changes that handle recursion plus improve variable resolution dependency. The logic is kinda kludgey, but I was having trouble finding a better way to handle it. I have a vague idea about how to improve the dependency handling in set_prop but I don't think I'm able to sufficiently articulate it right now. Plus the direction I'm thinking would require a change to the internal structure of the Style class related to specified properties.

I added some test cases to cover some scenarios that weren't working previously. The document I was using to test things follows.

<details>
<summary>sample document</summary>

```
<style>
    :root {
        --color: orange;
    }
    .border {
        font-size: var(--color);
        border: 2px solid var(--color);
        --color: green;
    }
    .inherit {
        border: inherit;
    }
</style>
<p class="border"><span class="inherit">inherit (green)</span></p>

<style>
    :root {
        --color: orange;
    }
    .merge {
        border: 2px solid var(--color);
        --color: green;
    }
    p.merge {
        --color: blue;
    }
</style>
<p class="merge">merge (blue)</p>

<style>
    .specific {
        border: 2px solid var(--color);
        border-color: red;
        --color: blue;
    }
</style>
<p class="specific">specific (red)</p>

<style>
    .specific2 {
        border: 2px solid var(--color);
        border-top-color: red;
        --color: blue;
    }
</style>
<p class="specific2">specific2 (blue, red top)</p>

<style>
    .specific3 {
        border-top-color: red;
        border: 2px solid var(--color);
        --color: blue;
    }
</style>
<p class="specific3">specific3 (blue)</p>

<style>
    .twice {
        border: 2px solid var(--color);
        --color: red;
        --color: purple;
    }
</style>
<p class="twice">twice (purple)</p>

<style>
    .multiple {
        border: 2px var(--style) var(--color);
        --style: dashed;
    }
</style>
<p class="multiple">multiple (orange dashed)</p>

<style>
    .nested {
        border: var(--border-specification);
        --border-specification: 2px solid var(--bg);
        --bg: #ff0000FF;
    }
</style>
<p class="nested">nested (red)</p>

<style>
    .undef {
        font-size: var(--undef);
    }
</style>
<p class="bad">bad (none)</p>

<style>
    .recurse {
        --one: var(--two);
        --two: var(--one);
    }
</style>

<style>
    .recurse-with-fallback {
        --one: var(--two);
        --two: var(--one, black);
    }
</style>

<style>
    .recurse-with-recursive-fallback {
        --one: var(--two, var(--two));
        --two: var(--one, var(--one));
    }
</style>

<style>
    .recurse-self {
        --one: var(--one, black);
    }
</style>

<style>
    .recurse-self-with-fallback {
        --one: var(--one, black);
    }
</style>

<style>
    .recurse-self-with-recursive-self-fallback {
        --one: var(--one, var(--one));
    }
</style>

<style>
    .recurse-self-with-recursive-self-fallback-with-fallback {
        --one: var(--one, var(--one, black));
    }
</style>
```

</details>

src/Css/Stylesheet.php
```diff
@@ -1612,6 +1612,7 @@ private function _parse_font_face($str)
      */
     private function _parse_properties($str)
     {
+        $properties = preg_split("/;(?=(?:[^\(]*\(.*?[^\)]*\))*(?![^\)]*\)))/", $str);
```
Mellthas, line 1615: Just a quick note: It seems this line was added during merge, due to the changes on master. It is obsolete now.

bsweeney: ah thank you I thought I had removed that

Neograph734:
I have just implemented this in my development environment. This obviously contains much more CSS than the test cases, but I must say that it appears to work quite good. My color variables all neatly resolve. Thanks for giving this the last push Brian! 

By the way, I cannot find that last requested change that github is complaining about...

bsweeney (approved):

bsweeney:
Glad it's working well for you as well.

I think that change request is just because I haven't updated my review so it's still reporting that I have requested changes. I added an approval review to make that go away. I have some other things I need to work on so I'll let this incubate a little longer in case anybody else has suggestions.

src/Css/Style.php
```diff
@@ -1440,6 +1546,12 @@ public function set_prop(string $prop, $val, bool $important = false, bool $clea
                 $val = self::$_defaults[$prop];
```
Mellthas, line 1546: `--custom: initial` and `--custom: unset` produce an undefined array-key warning here.

src/Css/Style.php
```diff
@@ -1451,6 +1563,27 @@ public function set_prop(string $prop, $val, bool $important = false, bool $clea
...
 
+            //TODO: this should be a directed dependency map
+            if (substr($prop, 0, 2) === "--" && !\in_array($prop, $this->_prop_stack, true)) {
+                array_push($this->_prop_stack, $prop);
+                $specified_props = array_filter($this->_props, function($key) {
+                    return \array_key_exists($key, $this->_props_specified);
+                }, ARRAY_FILTER_USE_KEY); // copy existing props filtered by those set explicitly before parsing vars
+                foreach ($specified_props as $specified_prop => $specified_value) {
+                    if (substr($specified_prop, 0, 2) !== "--" || strpos($specified_value, "var($prop") !== false) {
+                        $this->set_prop($specified_prop, $specified_value, \in_array($specified_prop, $this->_important_props, true), true);
```
Mellthas, line 1574: Need to use `isset` here instead of `in_array`, since the property names are stored as the array keys.

src/Css/Style.php
```diff
@@ -1451,6 +1563,27 @@ public function set_prop(string $prop, $val, bool $important = false, bool $clea
...
+                        if (isset(self::$_props_shorthand[$specified_prop])) {
+                            foreach (self::$_props_shorthand[$specified_prop] as $sub_prop) {
+                                if (\array_key_exists($sub_prop, $specified_props)) {
+                                    $this->set_prop($sub_prop, $specified_props[$sub_prop], \in_array($sub_prop, $this->_important_props, true), true);
```
Mellthas, line 1578: Same as above.

Mellthas (changes requested):
Don’t have the time for a full review currently, so just some minor review comments for now.

Neograph734:
I was wondering, we have introduced a lot of different variants of: 
```
substr($prop, 0, 2) === "--"
substr($prop, 0, 2) !== "--")
!substr($prop, 0, 2) == "--"
```

I am tempted to create a regex pattern for it, to bring it more in line with the rest of the code. At the same time I always learned that regex is slow. I am not sure if performance or code readbility is more important here. A small `is_custom_property()` function to wrap substr might also work?

Similarly, on line 1559 of Style.php we might want to use the self::CSS_VAR pattern instead of `strpos($val, 'var(') !== false)`?

bsweeney:
> I am tempted to create a regex pattern for it, to bring it more in line with the rest of the code. At the same time I always learned that regex is slow. I am not sure if performance or code readbility is more important here. A small `is_custom_property()` function to wrap substr might also work?

Performance is always top of mind, but I think we're safe adding a wrapper function to improve maintainability. I would expect that would be a compiler optimizable call.

> Similarly, on line 1559 of Style.php we might want to use the self::CSS_VAR pattern instead of strpos($val, 'var(') !== false)?

Yes, that makes sense. We should use consistent logic to ensure that the code is more maintainable.

src/Css/Style.php
```diff
@@ -1581,7 +1586,7 @@ protected function clear_cache(string $prop): void
     public function set_prop(string $prop, $val, bool $important = false, bool $clear_dependencies = true): void
     {
         // Skip some checks for CSS custom properties.
-        if (substr($prop, 0, 2) !== "--") {
+        if (!(\strpos($prop, '--') === 0)) {
```
bsweeney, line 1589: Should we call is_custom_property for this one as well?

Neograph734: Ah, not sure how I've overlooked that one. Will be there in a minute.

bsweeney:
I think we're brought this far enough, though I'm sure we could continue to optimize. I'll merge next week if no further updates or feedback are forthcoming.

Neograph734:
Great news, much appreciated!
I have nothing to add anymore.

PR #3316, opened 2023-11-03
Optimize `font_supports_text` to only check a single character

Mellthas:
We discussed this simplification during review of #3155, the method is only used in that way.

The performance impact is measurable with the test document from #3302. Here are average render times for that test case running on my machine (100 iterations with several repeats to account for outliers):

branch/tag | time
--- | ---
2.0.3 | 110.40 ms
master | 137.07 ms
font-fallback-perf-2 | 126.50 ms

Running the tests with other documents, the impact seems to be generally much less, up to being barely noticeable. I still think it is worth it.

I would just change the existing implementation and rename the method, but we could also keep `font_supports_text` in addition to `font_supports_char`. Thus marking as draft for now.

### DIFF
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -782,37 +782,21 @@ class CPDF implements Canvas
         }
     }
 
-    public function font_supports_text(string $font, string $text): bool
+    public function font_supports_char(string $font, string $char): bool
     {
-        if ($text === "") {
+        if ($char === "") {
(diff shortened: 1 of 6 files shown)


bsweeney (approved):
I think I'm fine with this change since it aligns with our actual usage.

It's too bad PHP doesn't support a char data type so we could, without extra logic, have this function and a function to check an entire string.

PR #3319, opened 2023-11-05
Fix undefined array key access with `@font-face` declaration

Mellthas:
When no format is specified or the format is not quoted.


bsweeney (approved):
This is a good tweak to the logic, thank you.

PR #3320, opened 2023-11-05
Fix loading stylesheets

fluffycondor:
- Fix bug with loading stylesheets with `rel="STYLESHEET"`, `rel="Stylesheet"` and etc. (Introduced in dompdf/dompdf@6fb7227)
- Don't load alternate stylesheets because they should be disabled by default (https://www.w3.org/Style/Examples/007/alternatives.en.html, https://developer.mozilla.org/en-US/docs/Web/CSS/Alternative_style_sheets#details)

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -589,9 +589,9 @@ class Dompdf
                 // load <link rel="STYLESHEET" ... /> tags
                 case "link":
                     if (
-                        (stripos(mb_strtolower($tag->getAttribute("rel")), "stylesheet") !== false // may be "appendix stylesheet"
+                        (stripos($tag->getAttribute("rel"), "stylesheet") !== false // may be "appendix stylesheet"
                         || mb_strtolower($tag->getAttribute("type")) === "text/css")
-                        && stripos(mb_strtolower($tag->getAttribute("rel")), "alternate") === false // don't load "alternate stylesheet"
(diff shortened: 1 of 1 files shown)


bsweeney (approved):

src/Dompdf.php
```diff
@@ -588,8 +588,10 @@ private function processHtml()
             switch (strtolower($tag->nodeName)) {
                 // load <link rel="STYLESHEET" ... /> tags
                 case "link":
-                    if (mb_strtolower(stripos($tag->getAttribute("rel"), "stylesheet") !== false) || // may be "appendix stylesheet"
-                        mb_strtolower($tag->getAttribute("type")) === "text/css"
+                    if (
+                        (stripos(mb_strtolower($tag->getAttribute("rel")), "stylesheet") !== false // may be "appendix stylesheet"
```
Mellthas, line 592: Since the value is lower-cased, you can use `strpos` instead of `stripos`; or remove the call to `mb_strtolower`. But we can also change that later.

Mellthas: Since the strings are ASCII-only, `mb_strtolower` could also be replaced by `strtolower`.

Mellthas (approved):

src/Dompdf.php
fluffycondor: You're right! Fixed

PR #3321, opened 2023-11-06
Add support for math CSS functions

samuferenc:
Add "function" support to the parser and implement some (abs, clamp, max, min).
"calc" already worked, "var" is need some guide because there is some PR in this topic

### DIFF
--- a/src/Css/Style.php
+++ b/src/Css/Style.php
@@ -188,6 +188,39 @@ class Style
         "'(?:[^']|\\\\['])*(?<!\\\\)'";   // String ''
 
     /**
+     * https://www.w3.org/TR/css-values-4/#calc-syntax
+     */
+    protected const CSS_MATH_FUNCTIONS = [
+        // Basic Arithmetic
+        "calc",
+        // Comparison Functions
+        "min",
+        "max",
+        "clamp",
+        // Stepped Value Functions
+        "round",                          // Not fully supported yet!!!
+        "mod",
+        "rem",
+        // Trigonometric Functions
+        "sin",
+        "cos",
+        "tan",
+        "asin",
+        "acos",
+        "atan",
+        "atan2",
+        // Exponential Functions
+        "pow",
+        "sqrt",
+        "hypot",
+        "log",
+        "exp",
+        // Sign-Related Functions
+        "abs",
+        "sign"
+    ];
+
+    /**
      * https://www.w3.org/TR/css-values-3/#custom-idents
      */
     protected const CUSTOM_IDENT_FORBIDDEN = ["inherit", "initial", "unset", "default"];
@@ -1025,11 +1058,6 @@ class Style
 
         $font_size = $font_size ?? $this->__get("font_size");
 
-        $calcPattern = "/^calc\((.*)?\)$/i";
-        if (preg_match($calcPattern, $l, $calcMatches)) {
-            return $this->evaluate_func_calc($this->parse_calc($calcMatches[1]), $ref_size, $font_size);
-        }
-
         $key = "$l/$ref_size/$font_size";
 
         if (\array_key_exists($key, $cache)) {
@@ -1040,6 +1068,12 @@ class Style
         $pattern = "/^($number)(.*)?$/";
 
         if (!preg_match($pattern, $l, $matches)) {
+            $ident = self::CSS_IDENTIFIER;
+            $pattern = "/^($ident)\(.*\)$/i";
+            if (preg_match($pattern, $l)) {
+                $value = $this->evaluate_func($this->parse_func($l), $ref_size, $font_size);
+                return $cache[$key] = $value;
+            }
             return null;
         }
 
@@ -1119,20 +1153,20 @@ class Style
      * @param string $expr infix expression
      * @return array
      */
-    private function parse_calc(string $expr): array
+    private function parse_func(string $expr): array
     {
         if (substr_count($expr, '(') !== substr_count($expr, ')')) {
             return [];
         }
 
-        $expr = str_replace(['(', ')', '*', '/'], [' ( ', ' ) ', ' * ', ' / '], $expr);
+        $expr = str_replace(['(', ')', '*', '/', ','], [' ( ', ' ) ', ' * ', ' / ', ' , '], $expr);
         $expr = trim(preg_replace('/\s+/', ' ', $expr));
 
         if ($expr === '') {
             return [];
         }
 
-        $precedence = ['*' => 3, '/' => 3, '+' => 2, '-' => 2];
+        $precedence = ['*' => 3, '/' => 3, '+' => 2, '-' => 2, ',' => 1];
 
         $opStack = [];
         $queue = [];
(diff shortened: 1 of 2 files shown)


bsweeney:
If we pull this one out of draft I'll re-target for the next release.

Mellthas:
Since this is about the math functions, `var` doesn’t need to handled here. It needs completely different treatment.

bsweeney:
> Since this is about the math functions, `var` doesn’t need to be handled here. It needs completely different treatment.

I think this is right. Limiting the parse method to just the [math functions](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Functions#math_functions) better targets the logic to a specific operation.

src/Css/Style.php
```diff
@@ -1203,7 +1278,17 @@ private function evaluate_func_calc(array $rpn, float $ref_size = 0, ?float $fon
...
+                    case ',':
+                        if (is_array($leftValue)) {
+                            $leftValue[] = $rightValue;
+                            $stack[] = $leftValue;
+                        } else {
+                            $stack[] = [$leftValue, $rightValue];
+                        }
+                        break;
                 }
+            } elseif (strncmp($part, '--', 2) === 0) {
```
Mellthas, line 1290: Okay, now I think I understand: You want to implement general CSS function parsing here. That would definitely be useful. For now, though, I would concentrate on just the math functions and ignore `var`.

We can then in the next step see how the logic here can be utilized in other places throughout the `Style` class. That would be a more ambitious task, and probably requires shuffling around a lot of code. If you want to look into that, you can open a separate PR later.

src/Css/Style.php
```diff
@@ -186,6 +186,10 @@ class Style
     protected const CSS_STRING = "" .
         '"(?:[^"]|\\\\["])*(?<!\\\\)"|' . // String ""
         "'(?:[^']|\\\\['])*(?<!\\\\)'";   // String ''
+    protected const CSS_MATH_FUNCTION_PATTERN = "/^(abs|acos|asin|atan|atan2|calc|clamp|cos|max|min|mod|rem|sin|tan)\((.*)?\)$/i";
+
+    /** List of supported css math functions  */
+    protected const CSS_MATH_FUNCTIONS = ["abs", "acos", "asin", "atan", "atan2", "calc", "clamp", "cos", "max", "min", "mod", "rem", "sin", "tan", "var"];
```
Mellthas, line 192: It would be clearer to format this into multiple lines, and use the order from the spec (https://www.w3.org/TR/css-values-4/#calc-syntax, plus add the link to the doc block); that would make it easier to see which functions are implemented and which are still missing.

src/Css/Style.php
```diff
@@ -1025,9 +1029,8 @@ protected function single_length_in_pt(string $l, float $ref_size = 0, ?float $f
 
         $font_size = $font_size ?? $this->__get("font_size");
 
-        $calcPattern = "/^calc\((.*)?\)$/i";
-        if (preg_match($calcPattern, $l, $calcMatches)) {
-            return $this->evaluate_func_calc($this->parse_calc($calcMatches[1]), $ref_size, $font_size);
+        if (preg_match(self::CSS_MATH_FUNCTION_PATTERN, $l)) {
```
Mellthas, line 1032: I would generalize the pattern to any function, since the code below is only concerned with single values and `evaluate_func` already returns `null` for unsupported functions.

tests/Css/StyleTest.php
```diff
@@ -43,15 +43,36 @@ public static function lengthInPtProvider(): array
             ["calc(50pt*2)", null, 100.0],                 // * do not require whitespace
             ["calc(50%/2)", 120, 30.0],                    // / do not require whitespace
             ["calc(10pt + -50%)", 12, 4.0],                // negative value
-            ["CalC(10)", null, 10.0],                      // case-insensitive
```
Mellthas, line 46: Why remove that?

samuferenc: Mistake, I corrected case-insensitive handling

samuferenc:
I changed the entry point, because I think this is cacheable

src/Css/Style.php
Mellthas: Sorry to bother you again, can you just remove any `var`-related code for now? :-) So remove the check here, remove the test cases, and do not merge `['var']` above.

If it turns out that here is the right place to handle it, we can always add that back later as part of or as follow-up to the other PRs.

tests/Css/StyleTest.php
```diff
@@ -51,7 +52,43 @@ public static function lengthInPtProvider(): array
...
+
+            // Stepped Value Functions
+            ["round(12.31, 1)", null, 12.3],               // complex
+            ["round(down, 12.31, 1)", null, 0.0],          // Yet not supported!!!
+            ["mod(7, 2)", null, 1.0],                      // mod function
+            ["rem(21, 6)", null, 3.0],                     // rem function
+
+            // Trigonometric Functions
+            // Can't test because different calc precision
+            // "sin","cos","tan","asin","acos","atan","atan2",
```
Mellthas, line 72: How about testing edge cases that are certain to have an integer result, e.g. `sin(0) == 0`, `cos(0) == 1` etc? For other cases, it should work to allow the result to match the expected value with an error margin: You could allow `testLengthInPt` to round the result with an optional fourth parameter, and use that for the trigonometric functions.

Mellthas (changes requested):

src/Css/Style.php
```diff
@@ -1173,18 +1212,103 @@ private function parse_calc(string $expr): array
...
+                        break;
+                    case 'rem':
+                        if ($argc !== 2) {
+                            return null;
+                        }
+                        $stack[] = fmod($argv[0], $argv[1]);
+                        break;
+                    case 'round':
+                        if ($argc === 2) {
+                            $stack[] = round($argv[0], $argv[1], PHP_ROUND_HALF_UP);
```
Mellthas, line 1284: From first glance, the second number parameter of CSS `round` does not directly map to the precision parameter of PHP `round`. The relation might be something like `precision = -log10(B)`, although PHP `round` only takes integer values for precision. See https://www.w3.org/TR/css-values-4/#round-func.

As with the modulo functions, I would rather leave the function unimplemented instead of incorrectly implemented.

Mellthas (changes requested):

src/Css/Style.php
samuferenc: I looked into it more closely, it's not math round. Not precision but "step"
I think the correct way is

```
$stack[] = round($argv[0] / $argv[1]) * $argv[1];

["round(0, 20)", null, 0.0],
["round(10, 20)", null, 20.0],
["round(20, 20)", null, 20.0],
["round(25, 20)", null, 20.0],
["round(30, 20)", null, 40.0],
["round(40, 20)", null, 40.0],

["round(0.54, 0.1)", null, 0.5],
["round(0.56, 0.1)", null, 0.6],

```

Mellthas: Yes, that’s much simpler. I think you are mostly there now, just need to figure out correct treatment of `A < 0`, and also verify it behaves correctly for `B < 0`. The spec states that the *nearest* rounding strategy rounds in the direction of +∞ if exactly between the bounds. That means `round(-30, 20)` should resolve to `-20`, not `-40` like the current implementation. See https://www.w3.org/TR/css-values-4/#css-round-to-the-nearest-integer.

The other rounding strategies should be similarly straightforward then, since you already have implemented them as part of `mod` and `rem`.

src/Css/Style.php
```diff
@@ -1173,18 +1212,108 @@ private function parse_calc(string $expr): array
...
+                            // calc(A - sign(B)*round(up, A*sign(B), B))
+                            $stack[] = $argv[0] - (ceil($argv[0] * -1 / $argv[1]) * $argv[1] * -1) ;
+                        }
+                        break;
+                    case 'rem':
+                        if ($argc !== 2 || $argv[1] === 0.0) {
+                            return null;
+                        }
+                        // calc(A - round(to-zero, A, B))
+                        $stack[] = $argv[0] - (intval($argv[0] / $argv[1]) * $argv[1]);
```
Mellthas, line 1287: The outer parentheses are superfluous.

src/Css/Style.php
```diff
@@ -1173,18 +1212,108 @@ private function parse_calc(string $expr): array
...
+                        break;
+                    case 'mod':
+                        if ($argc !== 2 || $argv[1] === 0.0) {
+                            return null;
+                        }
+                        if ($argv[1] > 0) {
+                            // calc(A - sign(B)*round(down, A*sign(B), B))
+                            $stack[] = $argv[0] - (floor($argv[0] / $argv[1]) * $argv[1]);
+                        } else {
+                            // calc(A - sign(B)*round(up, A*sign(B), B))
```
Mellthas, line 1278: This comment is misleading now, the first one is valid for both cases.

Mellthas (commented):
The other functions look fine so far.

Mellthas (approved):

PR #3323, opened 2023-11-08
Simplify regex for parsing property sections

Mellthas:
The original pattern fails on nested parentheses. Instead of trying to be too clever, simplify the pattern so the rules are clearer. Allow semicolons and parentheses to be escaped, which allows dealing with the parser shortcomings more easily.

See https://github.com/dompdf/dompdf/pull/3314#discussion_r1387312858.

### DIFF
--- a/src/Css/Stylesheet.php
+++ b/src/Css/Stylesheet.php
@@ -1614,7 +1614,9 @@ EOL;
     {
         $DEBUGCSS = $this->_dompdf->getOptions()->getDebugCss();
 
-        if ($DEBUGCSS) print '[_parse_properties';
+        if ($DEBUGCSS) {
+            print '[_parse_properties';
+        }
 
         // Split on non-escaped semicolons which are not followed by a single
         // closing parenthesis, to support semicolons as part of `url()`.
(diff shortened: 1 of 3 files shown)


Mellthas:
Example for escaped characters, which is handled correctly now:

```
content: "\;\)";
```

Mellthas:
Slipped in a small change to not lowercase custom-property names, which are case-sensitive. Can split that change out if necessary.

src/Css/Stylesheet.php
```diff
@@ -1612,39 +1612,26 @@ private function _parse_font_face($str)
      */
     private function _parse_properties($str)
     {
-        $properties = preg_split("/;(?=(?:[^\(]*\([^\)]*\))*(?![^\)]*\)))/", $str);
         $DEBUGCSS = $this->_dompdf->getOptions()->getDebugCss();
 
-        if ($DEBUGCSS) {
-            print '[_parse_properties';
-        }
+        if ($DEBUGCSS) print '[_parse_properties';
```
bsweeney, line 1617: Should we go ahead and add brackets even if the condition is a single line (just when changing relevant code)?

src/Css/Stylesheet.php
```diff
@@ -1612,39 +1612,26 @@ private function _parse_font_face($str)
...
 
-        // Create the style
+        // Split on non-escaped semicolons which are not followed by a single
+        // closing parenthesis, to support semicolons as part of `url()`.
+        // As a consequence, semicolons and closing parentheses should be
+        // escaped if used in a string
+        $properties = preg_split("/(?<!\\\\); (?! [^\(]* (?<!\\\\)\) )/x", $str);
```
bsweeney, line 1623: I haven't actually tested but just a read through and I think one thing that does not parse correctly (but that wasn't parsed correctly previously either) is a semi-colon in a string, e.g. `content: "dompdf;";`.

bsweeney (approved):
This seems to do a bit better than the previous logic.

src/Css/Stylesheet.php
Mellthas: Done. Just wanted to make the statement consistent with the other debug lines within the method, but I agree.

src/Css/Stylesheet.php
Mellthas: Yes, didn’t intend to change anything about that. But together with the updates to `Style::parse_string()` here, you can use it in escaped form now, at  least. So

```
content: "dompdf\;";
```

does work.

Mellthas:
Okay, let’s roll with it for now. I have some further related changes coming.

PR #3331, opened 2023-11-14
Fix checking for character support with core fonts

Mellthas:
Avoids unnecessarily splitting text frames when using one of the core fonts and characters which are included in Windows ANSI encoding, but not in ISO-8859-1, such as typographical quotation marks, or the en dash.

See https://en.wikipedia.org/wiki/Windows-1252

Changing the substitution character via `mb_substitute_character` is kind of awkward, but I haven’t found a better way (comparing to the current one is even more complicated).

### DIFF
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -788,15 +788,58 @@ class CPDF implements Canvas
             return true;
         }
 
-        $is_font_subsetting = $this->_dompdf->getOptions()->getIsFontSubsettingEnabled();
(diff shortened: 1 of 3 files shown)


bsweeney (approved):

Mellthas:
Okay, here is an alternative implementation that is a bit more performant: Instead of converting encoding, use the Unicode code points as basis and just translate the ones which are different.

bsweeney:
That's a nice tweak to more directly address where things differ.

PR #3345, opened 2023-12-05
Support `calc` in `border-spacing`, `transform-origin`, and `transform`

Mellthas:


src/Css/Style.php
```diff
@@ -3965,159 +4002,185 @@ protected function _compute_size(string $val)
...
-                return 0;
-            } elseif (\in_array($value, ["bottom", "right"], true)) {
-                return "100%";
-            } else {
-                return $value;
-            }
-        }, $values);
+        $v1 = $parts[0];
+        $v2 = $parts[1] ?? "center";
+        [$x, $y] = $this->computeBackgroundTransformPosition($v1, $v2);
```
bsweeney, line 4176: I don't feel too strong about this, but should we extract the logic of this method? Something like `_compute_transform_position`? `computeBackgroundTransformPosition` would then just be a passthrough to that method, but then the naming would be better aligned.

bsweeney (approved):

src/Css/Style.php
Mellthas: I am not sure I follow? Trying to shorten the name of the method made it more confusing than necessary probably; renamed it to `computeBackgroundPositionTransformOrigin` to make it clearer that it is simply common logic for both properties.

bsweeney: Looks like I misread things a bit. The naming confused me since it's a mash up of both property names. I was thinking it was meant for background position computation and was reused. I guess I don't really have a good naming suggestion then.

PR #3377, opened 2024-01-13
Implement new option "allowedRemoteHosts" to restrict which remote hosts can be requested

bohwaz:
This is a simple patch to add some kind of security when allowing remote requests.

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -1190,7 +1190,7 @@ class Options
             return [false, "Remote file requested, but remote file download is disabled."];
         }
 
-        if (is_array($this->allowedRemoteHosts)) {
+        if (is_array($this->allowedRemoteHosts) && count($this->allowedRemoteHosts) > 0) {
             $host = parse_url($uri, PHP_URL_HOST);
 
             if (!in_array($host, $this->allowedRemoteHosts, true)) {
--- a/src/Options.php
+++ b/src/Options.php
@@ -1053,6 +1053,15 @@ class Options
      */
     public function setAllowedRemoteHosts($allowedRemoteHosts)
     {
+        if (is_array($allowedRemoteHosts)) {
+            // Set hosts to lowercase
+            foreach ($allowedRemoteHosts as &$host) {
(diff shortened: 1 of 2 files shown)


bsweeney:
A user could already perform this validation using the existing `allowedProtocols` option. For example:

```
$options = new Dompdf\Options($dompdf_options);
$options->addAllowedProtocol("https://", function (string $uri) use (&$options) {
    $ret = $options->validateRemoteUri($uri);
    if ($ret[0] === false) {
        return $ret;
    }

    $parsed_url = parse_url($uri);
    if (strtolower($parsed_url["host"]) != "example.com") {
        return [false, "The remote file domain is not allowed."];
    }

    return [true, null];
});
$dompdf = new Dompdf\Dompdf($options);
```

bohwaz:
Yes but the callback approach is not documented, and this is much easier to use, and also promotes good practices :)

bsweeney (approved):

src/Options.php
```diff
@@ -1154,6 +1190,14 @@ public function validateRemoteUri(string $uri)
             return [false, "Remote file requested, but remote file download is disabled."];
         }
 
+        if (is_array($this->allowedRemoteHosts)) {
+            $host = parse_url($uri, PHP_URL_HOST);
+
+            if (!in_array($host, $this->allowedRemoteHosts, true)) {
```
bsweeney, line 1196: I wonder if we should `mb_strtolower` these values before comparison since case doesn't matter for host resolution.

src/Options.php
```diff
@@ -1154,6 +1190,14 @@ public function validateRemoteUri(string $uri)
             return [false, "Remote file requested, but remote file download is disabled."];
         }
 
+        if (is_array($this->allowedRemoteHosts)) {
```
bsweeney, line 1193:
```suggestion
        if (is_array($this->allowedRemoteHosts) && count($this->allowedRemoteHosts) > 0) {
```

?

PR #3405, opened 2024-02-28
Update FrameListIterator.php return types to be compatible with Iterator

wyatt3:
I have encountered the following warning when using the FrameListIterator class on PHP 8.1 using Dompdf version 2.0.4:

```Return type of Dompdf\Frame\FrameListIterator::key() should either be compatible with Iterator::key(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice```

The return type of `FrameListIterator::key()` is currently `int`. According to [php.net](https://www.php.net/manual/en/class.iterator.php), it should be `mixed` in order to be compatible.

The same is the case with `FrameListIterator::current()`, having a return type of `?Frame` when it should be `mixed` in order to be compatible.

### DIFF
--- a/src/Frame/FrameListIterator.php
+++ b/src/Frame/FrameListIterator.php
@@ -1,11 +1,9 @@
 <?php
-
 /**
  * @package dompdf
  * @link    https://github.com/dompdf/dompdf
  * @license http://www.gnu.org/copyleft/lesser.html GNU Lesser General Public License
  */
-
 namespace Dompdf\Frame;
 
 use Iterator;
--- a/src/Frame/FrameListIterator.php
+++ b/src/Frame/FrameListIterator.php
@@ -1,9 +1,11 @@
 <?php
+
 /**
  * @package dompdf
  * @link    https://github.com/dompdf/dompdf
  * @license http://www.gnu.org/copyleft/lesser.html GNU Lesser General Public License
  */
+
 namespace Dompdf\Frame;
 
 use Iterator;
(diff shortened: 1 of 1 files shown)


bsweeney:
What, specifically, is triggering this notice? Other systems running PHP 8.1 do not receive the message. Since Dompdf hasn't yet cut off support for PHP < 8.0 this would be a breaking change, so it'll have to wait for a major release.

wyatt3:
According to [this reference page](https://www.php.net/manual/en/language.oop5.inheritance.php) and [this RFC](https://wiki.php.net/rfc/internal_method_return_types) any class that implements or extends an internal class should have matching return types or it will cause a deprecation notice since PHP 8.1.

As far as preventing a breaking change for < 8.0, for now, we could add the `#[\ReturnTypeWillChange]` attribute to temporarily suppress this deprecation notice

bsweeney:
> According to [this reference page](https://www.php.net/manual/en/language.oop5.inheritance.php) and [this RFC](https://wiki.php.net/rfc/internal_method_return_types) any class that implements or extends an internal class should have matching return types or it will cause a deprecation notice since PHP 8.1.

I am aware, generally, of interface implementation requirements. And more specifically, after researching the changes in PHP, the specific issue here. I was just curious how you were able to raise the notice at all since I have been unable to reproduce.

> As far as preventing a breaking change for < 8.0, for now, we could add the #[\ReturnTypeWillChange] attribute to temporarily suppress this deprecation notice without changing the current return types

I do agree that this is the best mechanism for intermediate resolution of the issue.

Ultimately, we should match the signature of the interface so I'd like to keep this open as a reminder.

Issue #3419, opened 2024-03-17
Support for Hebrew / RTL


vraja-pro:
Following the discussion in [stack overflow,](https://stackoverflow.com/questions/21201257/arabic-fonts-display-in-reverse-order-in-dompdf) can you please add this code?

To fix hebrew letters direction add code to `src/Renderer/Text.php` before `$this->_canvas->text` (line 73).
I added a PR https://github.com/dompdf/dompdf/pull/3420

```php
 if ( strtolower( $style -> direction ) == 'rtl' && preg_match( "/\p{Hebrew}/u", $text ) ){

    preg_match_all('/./us', $text, $ar);

    // reverse the whole line
    $text = join('',array_reverse($ar[0]));

    // flip english back to ltr
    $words = explode( ' ', $text );
    foreach( $words as $i => $word ):
        if( !preg_match( "/\p{Hebrew}/u", $word ) ):
            $words[$i] = implode( '', array_reverse( str_split( $word ) ) );
        endif;
        if( $word === "–"){
            $words[$i] = "-";
        }
    endforeach;

    $text = implode( ' ', $words );
    }
```

bsweeney:
The PR is sufficient. Also, watch #2619 for updates on script support for various languages.

PR #3420, opened 2024-03-17
add Hebrew support

vraja-pro:
Following the discussion in [stack overflow,](https://stackoverflow.com/questions/21201257/arabic-fonts-display-in-reverse-order-in-dompdf) can you please add this code?

Also opened an issue https://github.com/dompdf/dompdf/issues/3419

Test instruction:
render to pdf the following code:
```html
<div style="font-family:firefly, DejaVu Sans, sans-serif;direction:rtl;text-align:right">בדיקה– 123</div>
```

With this PR, Check you get in the pdf the following result:

> בדיקה- 123

Without this PR the letters would appear in the reversed order.

### DIFF
--- a/src/Renderer/Text.php
+++ b/src/Renderer/Text.php
@@ -71,7 +71,27 @@ class Text extends AbstractRenderer
           array($this->_canvas->get_page_number()),
           $text
         );*/
+        if ( strtolower( $style -> direction ) == 'rtl' && preg_match( "/\p{Hebrew}/u", $text ) ){
 
+            preg_match_all('/./us', $text, $ar);
+        
+            // reverse the whole line
+            $text = join('',array_reverse($ar[0]));
+        
+            // flip english back to ltr
+            $words = explode( ' ', $text );
+            foreach( $words as $i => $word ):
+                if( !preg_match( "/\p{Hebrew}/u", $word ) ):
+                    $words[$i] = implode( '', array_reverse( str_split( $word ) ) );
+                endif;
+                if( $word === "–"){
+                    $words[$i] = "-";
+                }
+            endforeach;
+        
+            $text = implode( ' ', $words );
+        }
+        
         $this->_canvas->text($x, $y, $text,
             $font, $size,
             $style->color, $word_spacing, $letter_spacing);


bsweeney:
I'll review but a better solution would probably be to finalize the work started in #2107. Parsing your sample through the changes proposed in the PR also produce the desired result.

vraja-pro:
Hey @bsweeney! Sure, what is there to finalize? Can I help in any way?

bsweeney:
It's been a while since I last looked at 2107, so I don't have any definitive guidance at the moment. I need to review the functionality from the ground up to be sure that specific implementation makes the most sense.

That being said there are two things that it needs:
1. Rebasing on master so that it can be more directly merged after the change is complete.
2. Multi-frame, multi-line support. I'm pretty sure lines with multiple frames (e.g., text broken up by inline elements like span tags) or text that spans more than one line are not yet rendered correctly.

vraja-pro:
I checked out the branch in https://github.com/dompdf/dompdf/pull/2107 and rebased, the solution doesn't seem to work for me.
To be honest, my PR seems more straight forward and it actually does the work. Could be improved by adding a check for Arabic though.

bsweeney:
> I checked out the branch in https://github.com/dompdf/dompdf/pull/2107 and rebased, the solution doesn't seem to work for me.

That change is working OK for me with your sample. Still, it has a lot of deficiencies and, as I said, more work would be required before I'd consider releasing it.

Since that branch is pretty far behind master, I recreated the changes in the [rtl branch](https://github.com/dompdf/dompdf/tree/rtl).

> To be honest, my PR seems more straight forward and it actually does the work. Could be improved by adding a check for Arabic though.

Straight forward, sure, but it's not exactly a robust implementation even discounting that it only supports Hebrew.

A big deficiency is that this change performs the work within the context of a single frame, but a string of text can be made of multiple frames, so a more complex example won't render as expected. Try the following:

```
<p style="font-family:DejaVu Sans, sans-serif;direction:rtl;">Lorem Ipsum הוא פשוט טקסט <strong>גולמי של תעשיית</strong> ההדפסה וההקלדה.</p>
```

Additionally, this technique handles one aspect of complex text layout (directionality), but some languages require additional features (e.g., Arabic requires shaping).

PR #3433, opened 2024-04-15
FIX artifact path validation

semul:
### There Is A Small Problem

Instead Of: ~~public function validateArtifactPath(string $path, string $option)~~

We Should Use: `public function validateArtifactPath(?string $path, string $option)`

The `PATH` might be `NULL` in some cases.

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -1192,7 +1192,7 @@ class Options
     }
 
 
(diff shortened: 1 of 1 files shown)


bsweeney:
Apologies, that's a scenario I did not account for. There should be no need to set a path to null, but you are correct that it could happen.

PR #3435, opened 2024-04-15
Updates from 2.0.x releases not yet applied to master

bsweeney:
- Addresses a PHP compatibility issue in the GD back end
- Adds Options class support for validating artifact paths. The default validation does not accept paths that utilize the PHAR protocol.
- Bumps the minimum version of SvgLib to 0.5.2.

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -1194,6 +1194,9 @@ class Options
 
     public function validateArtifactPath(?string $path, string $option)
     {
+        if ($path === null) {
+            return true;
+        }
         $parsed_uri = parse_url($path);
         if ($parsed_uri === false || (array_key_exists("scheme", $parsed_uri) && strtolower($parsed_uri["scheme"]) === "phar")) {
             return false;
--- a/src/Options.php
+++ b/src/Options.php
@@ -84,6 +84,13 @@ class Options
     ];
 
     /**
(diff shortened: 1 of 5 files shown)


williamdes (approved):

src/Options.php
```diff
@@ -1150,6 +1191,16 @@ public function getHttpContext()
         return $this->httpContext;
     }
 
+
+    public function validateArtifactPath(?string $path, string $option)
+    {
+        $parsed_uri = parse_url($path);
```
williamdes, line 1197: This creates a PHP warning: `parse_url(): Passing null to parameter #1 ($url) of type string is deprecated in vendor/dompdf/dompdf/src/Options.php on line 1152`

bsweeney: Thank you for that. I really flubbed that implementation 😞.

williamdes: If you find the time it would be nice to have it fixed on 2.0.x :)

bsweeney: Absolutely! I will do both.

williamdes: phpunit has a setting to convertdeprecationsToExceptions

Maybe it would work

bsweeney: Indeed, it does! Thank you.

PR #3444, opened 2024-04-22
allow searching files in chroot options directories

eltharin:
with chroot option set to ['/dir/dir1', /dir/dir2'] files will be search there too, 

ex : <link src='/css/file.css'> wiil be search in basePath AND in /dir/dir1/css/file.css or /dir/dir2/css/file.css

### DIFF
--- a/src/Helpers.php
+++ b/src/Helpers.php
@@ -59,7 +59,7 @@ class Helpers
      * Vice versa, on using the local file system path of a file, make sure that the slash
      * is appended (o.k. also for Windows)
      */
-    public static function build_url($protocol, $host, $base_path, $url)
+    public static function build_url($protocol, $host, $base_path, $url, $chrootDirs = [])
     {
         $protocol = mb_strtolower($protocol);
         if (empty($protocol)) {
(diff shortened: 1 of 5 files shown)


src/Helpers.php
```diff
@@ -113,13 +113,26 @@ public static function build_url($protocol, $host, $base_path, $url)
...
+            if ($filepath !== false) {
+                $ret = "$protocol$filepath$res";
+
+                return $ret;
             }
 
-            $ret = "$protocol$filepath$res";
+            if ($url[0] == '/' && !empty($chrootDirs)) {
+                foreach ($chrootDirs as $dir) {
+                    $ret = realpath($dir) . '/' . $url;
```
bsweeney, line 124: Since we only reach this section when $url starts with '/' this should probably just be:

```suggestion
                    $ret = realpath($dir) . $url;
```

bsweeney (commented):
Seems like a reasonable modification for a common request (setting the file root to a document directory).

In addition to this we'll need to change all methods that call `build_url` to include `$chroot`.

eltharin:
yes it can be better to call build_url with with argument now :)

PR #3462, opened 2024-05-26
Update README.md: Adding "Framework Integration"

ThomasLandauer:
The idea would be to collect all bundles for all frameworks...

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,8 @@ For details see the [autoloader in the utils project](https://github.com/dompdf/
 ## Framework Integration
 
 * For Symfony: [nucleos/dompdf-bundle](https://github.com/nucleos/NucleosDompdfBundle)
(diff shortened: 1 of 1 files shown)


bsweeney:
[There is a wiki page for that](https://github.com/dompdf/dompdf/wiki/Third-party-Integrations).

ThomasLandauer:
OK, but since it's that short, I'd rather add those two on README too. Why? To me, it's an installation question: The earlier I know what I have to actually install, the better.

bsweeney:
I have no problem with that. Would you mind adding existing noted integrations from the wiki, plus the laravel package?

ThomasLandauer:
Sure. I also deleted the wiki page.

bsweeney:
If that list were to expand too much, we may want to restore the wiki page and list the most popular on the readme with a link. But the list is small so that's fine.

PR #3471, opened 2024-06-01
Refactor Options class set and get methods, and add comprehensive tes…

MuneebStack:
Refactor Options class set and get methods for improved readability and maintainability

- Updated the set and get methods in the Options class to enhance readability and reduce time complexity by eliminating long lists of if-else statements inside the loop. The methods now dynamically match attribute names using camelCase or underscore naming conventions.
- Added test cases to cover both camelCase and underscore naming conventions for setter and getter methods.
- Expanded existing test cases to include previously missing keys.

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -373,69 +373,25 @@ class Options
         if (!is_array($attributes)) {
             $attributes = [$attributes => $value];
         }
+
         foreach ($attributes as $key => $value) {
-            if ($key === 'tempDir' || $key === 'temp_dir') {
-                $this->setTempDir($value);
-            } elseif ($key === 'fontDir' || $key === 'font_dir') {
-                $this->setFontDir($value);
-            } elseif ($key === 'fontCache' || $key === 'font_cache') {
-                $this->setFontCache($value);
-            } elseif ($key === 'chroot') {
-                $this->setChroot($value);
-            } elseif ($key === 'allowedProtocols' || $key === 'allowed_protocols') {
-                $this->setAllowedProtocols($value);
-            } elseif ($key === 'artifactPathValidation') {
-                $this->setArtifactPathValidation($value);
-            } elseif ($key === 'logOutputFile' || $key === 'log_output_file') {
-                $this->setLogOutputFile($value);
-            } elseif ($key === 'defaultMediaType' || $key === 'default_media_type') {
-                $this->setDefaultMediaType($value);
-            } elseif ($key === 'defaultPaperSize' || $key === 'default_paper_size') {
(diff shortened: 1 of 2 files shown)


MuneebStack:
@bsweeney Please merge this PR.
Thank you !

bsweeney (changes requested):
In its current status the new unit tests do not pass due to additional validation in the Options class related to directory paths. There are a number of ways to resolve the issue, but the easiest fix would be to set it to a known path (e.g. `sys_get_temp_dir`).

MuneebStack:
> Are you able to resolve the conflict on this PR?

Ok @bsweeney I will update it today. Sorry I missed the message.

src/Options.php
```diff
@@ -388,71 +388,27 @@ public function set($attributes, $value = null)
...
-                $this->setDebugLayoutInline($value);
-            } elseif ($key === 'debugLayoutPaddingBox' || $key === 'debug_layout_padding_box') {
-                $this->setDebugLayoutPaddingBox($value);
-            } elseif ($key === 'pdfBackend' || $key === 'pdf_backend') {
-                $this->setPdfBackend($value);
-            } elseif ($key === 'pdflibLicense' || $key === 'pdflib_license') {
-                $this->setPdflibLicense($value);
-            } elseif ($key === 'httpContext' || $key === 'http_context') {
-                $this->setHttpContext($value);
             }
```
bsweeney, line 409: I think the conditionals can do without the first part (the "is*' check) since those would have matched on the very first `if` statement, right?

Thoughts on changing this block to the following slightly less optimized but slightly more readable version?

```
if ($methodForMatch === 'enablePhp') {
    $methodForCall = "setIsPhpEnabled";
} elseif ($methodForMatch === 'enableRemote') {
    $methodForCall = "setIsRemoteEnabled";
} elseif ($methodForMatch === 'enableJavascript') {
    $methodForCall = "setIsJavascriptEnabled";
} elseif ($methodForMatch === 'enableHtml5Parser') {
    $methodForCall = "setIsHtml5ParserEnabled";
} elseif ($methodForMatch === 'enableFontSubsetting') {
    $methodForCall = "setIsFontSubsettingEnabled";
} elseif ($methodForMatch === 'enablePdfA') {
    $methodForCall = "setIsPdfAEnabled";
}
if (method_exists($this, $methodForCall)) {
    $this->{$methodForCall}($value);
}
```

bsweeney: Same comments apply for the getter.

bsweeney (changes requested):

PR #3483, opened 2024-06-21
Update README.md

ThibaultPelloquin:
Adding info about using directly dompdf in Symfony without bundle

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,46 @@ Files accessed through the local file system have the following requirement:
  * Does not support CSS Grid.
  * A single Dompdf instance should not be used to render more than one HTML document
    because persisted parsing and rendering artifacts can impact future renders.
+
+## Tips
+
+### Use it in Symfony, without bundle
+
+A simple way to implements the feature in Symfony, without installing the NucleosDompdfBundle is to use the following code in your Controllers.
+
+```php
+<?php
+
+namespace App\Controller;
+
+use Dompdf\Dompdf;
+use Dompdf\Options;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\StreamedResponse;
+// ...
+
+class MyPdfController extends AbstractController
+{
+    public function getPdf(): Response
+    {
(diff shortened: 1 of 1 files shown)


Ayush9054:
<?php
require 'vendor/autoload.php';

use Dompdf\Dompdf;
use Dompdf\Options;

$options = new Options();
$options->set('isHtml5ParserEnabled', true);

$dompdf = new Dompdf($options);


$html = '<p>There is a less-than character after this word < Is it rendered?</p>';
$dompdf->loadHtml($html);

$dompdf->setPaper('A4', 'portrait');

$dompdf->render();

$dompdf->stream();
?>

Ayush9054:
THAT WAS A DUMMY COMMENT

README.md
```diff
@@ -231,6 +231,46 @@ Files accessed through the local file system have the following requirement:
...
+    public function getPdf(): Response
+    {
+        $html = $this->renderView(/* ... */);
+
+        $dompdf->loadHtml($html);
+        $dompdf->render();
+
+        $pdfFilename = "myfile.pdf";
+
+        // Preventing the ControllerDoesNotReturnResponseException
```
ronei-kunkel, line 263: To avoid this exception, instead use the `Dompdf::stream`, you can build your own response with `Dompdf::output`;

Example:
https://gist.github.com/ronei-kunkel/fdf6d1006f80a210ffce809608afc314

README.md
```diff
@@ -231,6 +231,46 @@ Files accessed through the local file system have the following requirement:
...
+        $response = new StreamedResponse();
+        $response->setCallback(function () use ($dompdf, $pdfFilename): void {
+            $dompdf->stream($pdfFilename, ['Attachment' => true]);
+        });
```
ronei-kunkel, line 267: To avoid this exception, instead use the `Dompdf::stream`, you can build your own response with `Dompdf::output`;

Example:
https://gist.github.com/ronei-kunkel/fdf6d1006f80a210ffce809608afc314

bsweeney:
Instead of a change to the README I recommend adding a page to the wiki.

Issue #2021, opened 2019-11-20
Helpers::encodeURI() breaks parameter strings


SlowFox71:
```
strtr(rawurlencode(rawurldecode($uri)), array_merge($reserved,$unescaped,$score));
```

This piece of code takes care that (amongst others), a "+" sign (used instead of a blank) remains a "+" sign. But problems arise if an URI contains a **real** plus sign, correctly encoded as "%2B" (or, even worse, an amphersand).

This small example reproduces the steps of the conversion:
```
$uri = '?a[]=1&b%5B%5D=1&c=d+e&f=g h&i=j%2Bk%26l';echo $uri . "\n";
echo rawurldecode($uri) . "\n";
echo rawurlencode(rawurldecode($uri)) . "\n";
echo strtr(rawurlencode(rawurldecode($uri)), array_merge($reserved, $unescaped, $score)) . "\n";
```

Results are:
> ?a[]=1&b%5B%5D=1&c=d+e&f=g h&i=j%2Bk%26l
> ?a[]=1&b[]=1&c=d+e&f=g h&i=j+k&l
> %3Fa%5B%5D%3D1%26b%5B%5D%3D1%26c%3Dd%2Be%26f%3Dg%20h%26i%3Dj%2Bk%26l
> ?a%5B%5D=1&b%5B%5D=1&c=d+e&f=g%20h&i=j+k&l

Everything is fine for characters which have to be escaped like "[]" and fine for characters which may not be escaped like "?" and "&". But if special characters are in the original string, they are broken in the result, as in "i=j+k&l".

I am not far enough into this to have a solution at my fingers. The first rawurldecode() ruins the string, so a completely different approach must be used here.

SlowFox71:
Ok, some findings:

Helpers::encodeURI($uri) does, what the comments say: "Assumes that the URI is a complete URI, so does not encode reserved characters that have special meaning in the URI".

But if there is an <img> tag in the HTML source, this is handled in the constructor of FrameDecorator\Image:

`$url = $frame->get_node()->getAttrlist("src");`
`list($this->_image_url, /*$type*/, $this->_image_msg) = Cache::resolve_url($url, ...);`

[...]

PR #3500, opened 2024-07-24
Use cURL over file_get_content when available

Starfox64:
Dompdf will always use `file_get_content` as long as `allow_url_fopen` is enabled (which it is by default).
This unfortunately means that cURL will essentially never be used and we are stuck with file_get_contents and it's limitations:
- `follow_location` does not work under certain circumstances
- #2021 still corrupts some characters (@ symbol)

I think usage of cURL should have priority over `file_get_contents` as it's more robust.

I considered making this configurable but since this method is static and doesn't have access to configuration, this becomes difficult.

### DIFF
--- a/src/Helpers.php
+++ b/src/Helpers.php
@@ -999,12 +999,12 @@ class Helpers
         $headers = null;
         [$protocol] = Helpers::explode_url($uri);
         $is_local_path = in_array(strtolower($protocol), ["", "file://", "phar://"], true);
-        $can_use_curl = in_array(strtolower($protocol), ["http://", "https://"], true);
+        $can_use_curl = in_array(strtolower($protocol), ["http://", "https://"], true) && function_exists('curl_exec');
 
         set_error_handler([self::class, 'record_warnings']);
 
         try {
-            if ($is_local_path || ini_get('allow_url_fopen') || !$can_use_curl) {
+            if ($is_local_path || ini_get('allow_url_fopen') && !$can_use_curl) {
                 if ($is_local_path === false) {
                     $uri = Helpers::encodeURI($uri);
                 }
@@ -1020,7 +1020,7 @@ class Helpers
                     $headers = $http_response_header;
                 }
 
-            } elseif ($can_use_curl && function_exists('curl_exec')) {
+            } elseif ($can_use_curl) {
                 $curl = curl_init($uri);
 
                 curl_setopt($curl, CURLOPT_RETURNTRANSFER, true);


bsweeney:
> `follow_location` does not work under certain circumstances

Under what circumstances? Follow location is [disabled by default](https://github.com/dompdf/dompdf/blob/v3.0.0/src/Options.php#L350-L355). When enabled, under what circumstances does it not work with `file_get_contents`?

> Helpers::encodeURI still corrupts some characters (@ symbol)

Do you have a sample URL?

Starfox64:
In my use case images point to some API that redirects to the actual image in Google Cloud Storage.

ie. https://storage.googleapis.com/foc-files/dev/private/5e2b15573533314e494a74f2/8b/dd/a3/66a0d03033cbadbc6c003b82_rendez_vos_pdf_interactifs_directement_sur_franchise_on_cloud_v2.png?fileRevision=1721815088&GoogleAccessId=extranet%40franchise-on-cloud.iam.gserviceaccount.com&Expires=1721944800&Signature=DMH5GiL07NmB9IjawapE5%2B%2FmKg29aIyqukjXXkD6aeqUiResrBiNV%2Bk%2B%2FfsWYw3y0AvdDIWy%2BBlCauH85g%2Fs0TNZiU88%2BIitXswaGS%2BraoZB7pvi96pbIRY5M4%2FhLn9nNF5IwkUfGctxeA1K%2FFm5e3gQRFhU8xo0aa9Pt8VKSicHNSER2Xq9YM0CGAw2BvRvFHOQ6KlO1gPJwSltlaC9%2FwJoFuYI9XBKqCqPp9ieRNDt3pQNgYMt8ogE5kt6R2tFQplTB8OGRCXOY44cNyMgT%2BF44shN5m%2BnZdtvgEFvN1Egld9jp%2FzhDvoMLj2b9lmTY1zEtAPqNis0vMW7inYY%2FQ%3D%3D&response-content-type=image/png&response-content-disposition=inline;%20filename=%22Rendez-vos-PDF-interactifs-directement-sur-Franchise-On-Cloud_v2.png%22

It's a signed temporary URL so any change will cause the signature to be voided, in this case `Helpers::encodeURI` will replace `GoogleAccessId`'s `%40` by `@`.

`file_get_contents` also seems to mangle the URL in some way during redirects because I end up with a 400 error from Google Cloud Storage, meaning that it understood my API's redirect but broke the URL.

bsweeney:
Thanks for the info. The change seems fine on first look. I'll let you know if I have any feedback once I have a chance to mull it over.

bsweeney:
I don't think the issue is with Helpers::encodeURI. When I run that sample URL through the method it returns without any changes (in the latest release). Using the URL in an image src attribute shows correct handling internally as far as I can tell. `file_get_content` handling could be problematic, but I don't have a valid URL I can use for testing.

bsweeney:
I did a test with the path+querystring from that sample URL using an online [request inspector](https://requestinspector.com/) and it showed exactly what was expected. Possible that we've addressed whatever issues you were previously experiencing. If you find otherwise let me know.

PR #3502, opened 2024-07-31
Add support for phpunit 11

williamdes:
~~Some tests seemed to have duplicate annotations, of course the last one wins. I removed the wrong one.~~

No breaking changes to be expected

### DIFF
--- a/tests/Canvas/CPDFTest.php
+++ b/tests/Canvas/CPDFTest.php
@@ -129,6 +129,7 @@ class CPDFTest extends TestCase
     /**
      * @dataProvider fontSupportsCharProvider
      */
+    #[\PHPUnit\Framework\Attributes\DataProvider('fontSupportsCharProvider')]
     public function testFontSupportsChar(string $font, string $char, bool $expected): void
     {
         $dompdf = new Dompdf();
--- a/tests/Css/AttributeTranslatorTest.php
+++ b/tests/Css/AttributeTranslatorTest.php
@@ -80,6 +80,7 @@ HTML
      *
      * @dataProvider attributeToStyleTranslationProvider
      */
+    #[\PHPUnit\Framework\Attributes\DataProvider('attributeToStyleTranslationProvider')]
     public function testAttributeToStyleTranslation(
         string $body,
         array $expectedStyles
--- a/tests/Css/StyleTest.php
+++ b/tests/Css/StyleTest.php
@@ -187,6 +187,7 @@ class StyleTest extends TestCase
     /**
      * @dataProvider lengthInPtProvider
      */
+    #[\PHPUnit\Framework\Attributes\DataProvider('lengthInPtProvider')]
     public function testLengthInPt(string $length, ?float $ref_size, $expected, ?int $precision = null): void
     {
         $dompdf = new Dompdf();
@@ -247,6 +248,8 @@ class StyleTest extends TestCase
      * @dataProvider cssImageNoBaseHrefProvider
      * @group regression
      */
+    #[\PHPUnit\Framework\Attributes\DataProvider('cssImageBasicProvider')]
+    #[\PHPUnit\Framework\Attributes\DataProvider('cssImageNoBaseHrefProvider')]
     public function testCssImageNoBaseHref(string $value, $expected): void
     {
         $dompdf = new Dompdf();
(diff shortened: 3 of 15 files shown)


Mellthas:
> Some tests seemed to have duplicate annotations, of course the last one wins. I removed the wrong one.

Multiple data providers work perfectly fine, they are applied in order. Some of the tests share common data sets this way.

tests/Canvas/CPDFTest.php
```diff
@@ -129,6 +129,7 @@ public static function fontSupportsCharProvider(): array
     /**
      * @dataProvider fontSupportsCharProvider
      */
+     #[\PHPUnit\Framework\Attributes\DataProvider('fontSupportsCharProvider')]
```
Mellthas, line 132: I think importing the attribute classes is nicer in general.

Mellthas: Oh, just realized that this wouldn’t work, as they are unavailable before PHP 8. Never mind.

williamdes:
> > Some tests seemed to have duplicate annotations, of course the last one wins. I removed the wrong one.
> 
> Multiple data providers work perfectly fine, they are applied in order. Some of the tests share common data sets this way.

Not sure why they do not fail with my changes then
Can you check some examples for me?

Mellthas:
This will cause no tests to fail, it simply means some test cases are not run anymore when using PHPUnit 10 or later.

Output on master (PHPUnit 10.5.29):

```
Tests: 1112, Assertions: 2733, Deprecations: 1.
```

Output with this PR (both PHPUnit 10.5.29 and PHPUnit 11.2.9):

```
Tests: 1026, Assertions: 2491, Deprecations: 1.
```

As an example, running `StyleTest::testCssImageNoBaseHref()` with testdox output on master (both `cssImageBasicProvider` and `cssImageNoBaseHrefProvider` applied):

```
Style (Dompdf\Tests\Css\Style)
 ✔ Css image no base href with no·value
 ✔ Css image no base href with keyword·none
 ✔ Css image no base href with bare·url
 ✔ Css image no base href with http
 ✔ Css image no base href with case
 ✔ Css image no base href with quoted·parens
 ✔ Css image no base href with escaped·parens
 ✔ Css image no base href with quotes
 ✔ Css image no base href with escaped·quotes
 ✔ Css image no base href with local·absolute
 ✔ Css image no base href with local·relative
```

With this PR (only `cssImageNoBaseHrefProvider` applied):

```
Style (Dompdf\Tests\Css\Style)
 ✔ Css image no base href with local·absolute
 ✔ Css image no base href with local·relative
```

tests/Css/AttributeTranslatorTest.php
```diff
@@ -80,15 +80,16 @@ public static function attributeToStyleTranslationProvider(): array
      *
      * @dataProvider attributeToStyleTranslationProvider
      */
+     #[\PHPUnit\Framework\Attributes\DataProvider('attributeToStyleTranslationProvider')]
     public function testAttributeToStyleTranslation(
         string $body,
         array $expectedStyles
     ): void {
-        $styles = array_fill_keys(array_keys($expectedStyles), []);
+         $styles = array_fill_keys(array_keys($expectedStyles), []);
```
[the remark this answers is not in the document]
Mellthas: This seems to be caused by the extra single-space indent in the attribute lines (5 spaces instead of 4). After removing these extra spaces, phpcs is happy.

williamdes:
By the way, I maintain this GitHub action: https://github.com/sudo-bot/action-scrutinizer

You may want to use it to enable back Scrutinizer coverage upload

tests/Css/StyleTest.php
```diff
@@ -247,6 +248,8 @@ public static function cssImageWithStylesheetBaseHrefProvider(): array
      * @dataProvider cssImageNoBaseHrefProvider
      * @group regression
```
Mellthas, line 249: Can you also add the `PHPUnit\Framework\Attributes\Group` attribute for the `@group` annotations?

Mellthas (approved):
Other than that, looks good, thanks!

tests/Css/StyleTest.php
williamdes: done 🎉

Mellthas (approved):

williamdes:
@bsweeney could we have this one merged ?

PR #3507, opened 2024-08-27
fix getimagesize

kurniagungk:


bsweeney:
What are you trying to fix with this change?

kurniagungk:
> What are you trying to fix with this change?


what I fixed was getimagesize(): Corrupt JPEG data: foreign bytes before marker

bsweeney:
Can you share an image that reproduces the issue?

kurniagungk:
image
php 8.3.0 windows 11

bsweeney:
And do you have a sample image that triggers the error?

lib/Cpdf.php
```diff
@@ -6249,7 +6249,7 @@ function addJpegFromFile($img, $x, $y, $w = 0, $h = 0)
             $imageHeight = $this->imagelist[$img]['h'];
             $channels = $this->imagelist[$img]['c'];
         } else {
-            $tmp = getimagesize($img);
+            $tmp = @getimagesize($img);
```
bsweeney, line 6252: If the image is not readable here `$tmp` is false and the following two lines will generate warnings since you can't access an offset of a boolean.

kurniagungk: This is a sample image that created the error
https://drive.google.com/file/d/1PSQKGyUVyj99I7zS-E9lLmdRtrrbQ1Ic/view?usp=drive_link

bsweeney: You'll need to give me permission to access the file.

kurniagungk: Yes, I have granted access.

bsweeney: Thanks, I can reproduce with that image. That image is problematic but can still be processed by GD. So, your change works for this particular case. We should still tweak the logic in case the image can't be processed at all. I think all we need to do is immediately return if the result is false:

```suggestion
            $tmp = @getimagesize($img);
            if ($tmp === false) {
                return;
            }
```

PR #3508, opened 2024-08-27
Address php8.4 deprecation of with type hinting on NULL

eileenmcnaughton:
This avoids the following php 8.4 deprecation 

Dompdf\Options::__construct(): Implicitly marking parameter $attributes as nullable is deprecated, the explicit nullable type must be used instead

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -330,7 +330,7 @@ class Options
     /**
      * @param array $attributes
      */
-    public function __construct(array $attributes = null)
+    public function __construct(?array $attributes = null)
     {
         $rootDir = realpath(__DIR__ . "/../");
         $this->setChroot(array($rootDir));


bsweeney:
Do you have plans to address the other instances where this notice arises? 3521 has a list of problematic functions.

eileenmcnaughton:
@bsweeney we picked this one up in our unit tests - it fails straight up on 8.4 - I'm assuming we will hit a bunch more in our tests once our tests get past this & if not fixed by then would put up more PRs but I'm not actively looking into 8.4 this month

It's probably worth merging this as is because it will likely be the first 8.4 one anyone hits & you probably want to flush out the more obscure ones

bsweeney:
I've already patched up what I found based on running my own local tests as well as searching using the regex provided in 3521. I just wanted to know if you intended to fix the rest.

eileenmcnaughton:
@bsweeney ok great thanks. I won't be looking at this further in the short-term

eileenmcnaughton:
@bsweeney just checking in on this - did you have some commits you were going to push up?

bsweeney:
Apologies things have been ... hectic. I will try to get to it over the next week.

eileenmcnaughton:
thanks @bsweeney appreciate it - it sounds like the work is mostly done, just not pushed up,  from prior comments

bsweeney:
#3521 beat me to the change so I'm closing this PR in favor of that one. Thanks for surfacing the compatibility issue.

Issue #3523, opened 2024-09-13
PHP Fatal error: Call to undefined function ini_set


maltfield:
This bug report fixes a PHP Fatal error caused by `Dompdf.php` when `ini_set` is disabled

# Problem

There is a bug in `vendor/dompdf/dompdf/src/Dompdf.php at the following lines:

```
        @ini_set('pcre.jit', '0');
```

and

```
            @ini_set('pcre.jit', $this->pcreJit);
```

These lines causes a PHP Fatal error on hardened systems with the `ini_set` function disabled.

```
PHP Fatal error:  Uncaught Error: Call to undefined function Dompdf\\ini_set() in /path/to/wp-content/plugins/woocommerce-pdf-invoices-packing-slips/vendor/dompdf/dompdf/src/Dompdf.php:293...
```

# Why this matters

For security reasons, orgs frequently configure php.ini to be hardened by adding many dangerous functions to the `disable_functions` variable in the php.ini file. For example, it's common to disable the 'exec' function

```
disable_functions = exec
```

Of course, if a php script could modify the php configuration, then it would defeat any hardening done by setting `disable_functions`. As such, it's common to add `ini_set` to the `disable_functions`

```
disable_functions = exec, ini_set
```

# Solution

To fix the PHP Fatal error, we should always check to see if the `ini_set` function exists before attempting to call it

```
if( function_exists( 'ini_set') ){
        @ini_set('pcre.jit', '0');
}
```

# Context

This error was discovered on a server running the latest version of all software on the latest version of Debian, at the time of writing.

```
root@host:~# cat /etc/issue
Debian GNU/Linux 12 \n \l

root@host:~# 

root@host:~# dpkg -l | grep -iE 'php-common'
ii  php-common                      2:93                                    all          Common files for PHP packages
root@host:~# 
```

maltfield:
See also downstream bug report:

 * https://github.com/dompdf/dompdf/issues/3523

maltfield:
Submitted PR to fix this:

 * https://github.com/dompdf/dompdf/pull/3524

maltfield:
This appears to be a regression bug?

 * https://github.com/dompdf/dompdf/pull/2049

bsweeney:
> This appears to be a regression bug?

Not a regression, the original fix prefixed those calls with an `@` to suppress errors and that change is still present.

Are you sure the error is from Dompdf? Your original message references PEAR.php:52.

maltfield:
[...]

PR #3524, opened 2024-09-13
check if ini_set is available to prevent Fatal Errors

maltfield:
* https://github.com/dompdf/dompdf/issues/3523

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -284,16 +284,18 @@ class Dompdf
      */
     private function setPhpConfig()
     {
-        if (sprintf('%.1f', 1.0) !== '1.0') {
-            $this->systemLocale = setlocale(LC_NUMERIC, "0");
-            setlocale(LC_NUMERIC, "C");
+        if (function_exists('ini_get') && function_exists('ini_set')) {
+            if (sprintf('%.1f', 1.0) !== '1.0') {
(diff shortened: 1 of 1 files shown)


src/Dompdf.php
```diff
@@ -284,36 +284,40 @@ public function __construct($options = null)
...
-        $this->pcreJit = @ini_get('pcre.jit');
-        @ini_set('pcre.jit', '0');
+            $this->pcreJit = @ini_get('pcre.jit');
+            @ini_set('pcre.jit', '0');
 
-        $this->mbstringEncoding = mb_internal_encoding();
-        mb_internal_encoding('UTF-8');
+            $this->mbstringEncoding = mb_internal_encoding();
+            mb_internal_encoding('UTF-8');
+        }
```
bsweeney, line 298: Why did you wrap this whole block in the condition instead of just the lines using `ini_get`/`ini_set`?

maltfield: because I assumed that was the best way to do this with the fewest number of lines added. I can add a few more conditionals so that only the 1-line `ini_set` and `ini_get` calls are wrapped, if others think that's better?

bsweeney: The preference would be to change only the relevant lines. You're also disabling changes to locale and internal encoding, both of which can completely break rendering if configured in a way that's incompatible with Dompdf.

PR #3563, opened 2024-12-10
xml_set_*_handler() deprecation fix for PHP 8.4

laoneo:
With the release of PHP 8.4, the xml_set_*_handler() function arguments must be callable or null. Otherwise is a deprecated warning thrown:

_Passing non-callable strings is deprecated since 8.4_

The message is a bit misleading as dompdf uses `false` as end handler and not a string.

### DIFF
--- a/src/Image/Cache.php
+++ b/src/Image/Cache.php
@@ -168,7 +168,7 @@ class Cache
                             }
                         }
                     },
-                    false
+                    function () {}
                 );
         
                 if (($fp = fopen($resolved_url, "r")) !== false) {
--- a/src/Image/Cache.php
+++ b/src/Image/Cache.php
@@ -168,7 +168,7 @@ class Cache
                             }
                         }
                     },
-                    function () {}
+                    null
                 );
         
                 if (($fp = fopen($resolved_url, "r")) !== false) {


src/Image/Cache.php
```diff
@@ -168,7 +168,7 @@ function ($parser, $name, $attributes) use ($options, $parsed_url, $full_url) {
                             }
                         }
                     },
-                    false
+                    function () {}
```
bsweeney, line 171: Not sure why we used false previously. Wouldn't it be better to just specify `null` here?

laoneo: Sure, don't know why I made it so complicated. Too many things to do at the same time.

williamdes:
Is this covered by tests, I did not see it fail 🤔

laoneo:
Don't know,but it's a deprecated warning only.

bsweeney:
I don't think there are any tests related to SVG parsing. I'll add some tests to this PR to cover some use cases (unless somebody beats me to it).

bsweeney:
I went ahead and added a test case via #3564 so I will move forward with merging this. Deprecations don't currently generate errors so unless somebody was really paying attention this could have been missed.

williamdes:
> I went ahead and added a test case via #3564 so I will move forward with merging this. Deprecations don't currently generate errors so unless somebody was really paying attention this could have been missed.

Thank you!
Indeed Debian is re building packages with PHP 8.4 and php-dompdf did not fail

laoneo:
Thank you guys, in the meantime I shipped my fork with the fix.

PR #3578, opened 2025-01-17
Allow custom adapter classes

### LINKED ISSUE #3577: Allow to write a custom backend with a PSR-4 class
In the code here
https://github.com/dompdf/dompdf/blob/723529bda1cbc3cf7a819d5fd4303554979172fb/src/CanvasFactory.php#L34-L40

There is no possibility to have a PSR-4 class loaded

If you change the code to
```php
        $backendOption = $dompdf->getOptions()->getPdfBackend();
        $backend = strtolower($backendOption);
        if (isset($class) && class_exists($class, false)) {
            $class .= "_Adapter";
        } elseif (class_exists($backendOption, true)) {
                $class = $backendOption;
        }
```

Then is is possible to load a backend that would extend Cpdf.
Defined in `"pdf_backend" => \App\Invoices\CustomCpdf::class,` in the options array.

Let me know if this is how it should work @bsweeney

bsweeney:
fixes #3577

### DIFF
--- a/src/CanvasFactory.php
+++ b/src/CanvasFactory.php
@@ -45,7 +45,9 @@ class CanvasFactory
             }
 
             else {
-                if ($backend === "gd" && extension_loaded('gd')) {
+                if (class_exists($backend, false)) {
+                    $class = $backend;
+                } elseif ($backend === "gd" && extension_loaded('gd')) {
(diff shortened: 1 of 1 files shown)


williamdes (approved):
This looks nice

PR #3579, opened 2025-01-20
Updates to PDF/A support

bsweeney:
- Improves structure of embedded file metadata
- Adds initial support for defining the relationship between an object and an embedded file


lib/Cpdf.php
```diff
@@ -5727,19 +5791,44 @@ function restoreSerializedObject($obj)
      * @param string $filepath path to the file to store inside the PDF
      * @param string $embeddedFilename the filename displayed in the list of embedded files
      * @param string $description a description in the list of embedded files
+     * @param string $mimeType the mime type of the file
+     * @param array $relatedTo array where the key is the object ID and the value is a string naming the relationship type
      */
-    public function addEmbeddedFile(string $filepath, string $embeddedFilename, string $description): void
+    public function addEmbeddedFile(string $filepath, string $embeddedFilename, string $description, string $mimeType = "application/octet-stream", $relatedTo = []): void
```
williamdes, line 5797: here is missing like in mpdf the array param to add other attributes like afRelationship and mime

williamdes (approved):

TimexPeachtree (approved):

lib/Cpdf.php
bsweeney: `$mimeType` will access the mime type of the file.

`$relatedTo` is intended to define the relationship to another object where the key is the object ID and the value is the relationship (e.g., "Data"). Thinking on this more, referencing a file from more than one object is ... unlikely? I don't recall seeing that it was not possible, but it doesn't seem like a common scenario. I should maybe change the signature to:

```
public function addEmbeddedFile(string $filepath, string $embeddedFilename, string $description, string $mimeType = "application/octet-stream", int? $relatedObject, string? $relationship = "Unspecified"): void
```

_Or_, I could just take the relationship arguments out and leave file association as an additional call for the user via the `associateFile` method (just below this one). However, that's a bit more work because you have to find the embedded file dictionary object number in the object collection first.

FYI, I set the relationship logic up this way because the PDF spec allows associating files with more than just the catalog. You can add associated files for other embedded object (fonts, images) as well. Looking at the mpdf logic that library seems to only support associating files with the catalog object. Though Cpdf currently only supports associating files with the catalog I wanted to make sure it could associate with other objects in the future with minimal change.

williamdes: Sorry @bsweeney 
My eyes where not aligned, the current signature is perfect
I am not not sure there is anything that needs a change

bsweeney: I'll be sure to document the last parameter better.

williamdes, line 5797:
```suggestion
    public function addEmbeddedFile(string $filepath, string $embeddedFilename, string $description, string $mimeType = "application/octet-stream", array $relatedTo = []): void
```

lib/Cpdf.php
```diff
@@ -5727,19 +5791,44 @@ function restoreSerializedObject($obj)
...
             [
                 'filepath' => $filepath,
                 'filename' => $embeddedFilename,
-                'description' => $description
+                'description' => $description,
+                'mimeType' => $mimeType
             ]
         );
+
+        if (isset($relatedTo) && count($relatedTo) > 0) {
```
williamdes, line 5811:
```suggestion
        if ($relatedTo !== []) {
```

lib/Cpdf.php
```diff
@@ -5727,19 +5791,44 @@ function restoreSerializedObject($obj)
...
+            }
+        }
+    }
+
+    /**
+     * Associated an embedded file with an existing PDF object.
+     *
+     * @param int    $fileId       The PDF object ID of the file dictionary
+     * @param int    $objectId     The PDF object ID of the object the file is related to
+     * @param string $relationship The relationship type, valid values are defined in section 7.11.3 of ISO-32000-2
```
williamdes, line 5823:
```suggestion
     * @param string $relationship The relationship type, valid values are defined in section 7.11.3 of ISO-32000-2
     * @phpstan-param 'Source'|'Data'|'Alternative'|'EncryptedPayload'|'FormData'|'Schema'|'Unspecified'
```

Copied from https://developer.adobe.com/document-services/docs/assets/5b15559b96303194340b99820d3a70fa/PDF_ISO_32000-2.pdf

lib/Cpdf.php
```diff
@@ -5727,19 +5791,44 @@ function restoreSerializedObject($obj)
      * @param string $filepath path to the file to store inside the PDF
      * @param string $embeddedFilename the filename displayed in the list of embedded files
      * @param string $description a description in the list of embedded files
+     * @param string $mimeType the mime type of the file
+     * @param array $relatedTo array where the key is the object ID and the value is a string naming the relationship type
```
williamdes, line 5795:
```suggestion
     * @param array $relatedTo array where the key is the object ID and the value is a string naming the relationship type
     * @phpstan-param array<int,'Source'|'Data'|'Alternative'|'EncryptedPayload'|'FormData'|'Schema'|'Unspecified'> $relatedTo
```

lib/Cpdf.php
williamdes: I added phpstan annotations, it will help a lot users that implement and use tools like Psalm or PHPStan.
That said the doc comment in unclear if the user should create it's own object IDs

williamdes:
@bsweeney can this be merged ?
maybe with the phpstan annotations I suggested, else discard them I can do a PR for it

lib/Cpdf.php
bsweeney: I'll tweak the $relatedTo description to read:
> relationship of the embedded file to other PDF objects. The key is the object ID generated by Cpdf when adding the related object and the value is a (constrained) string naming the relationship type.

This relationship value should probably just be an enum ... but for PHP 7 support 😬. We could use constants instead. Or just wait until we bump the minimum PHP version to 8 and fix it up then.

bsweeney: I'll also add information to the wiki about how to use this method in relation to factur-x.

Adesin-fr:
Hi @bsweeney 

Sorry to urge you, but we need to be FacturX ready soon here in France, and we need this PR to be merged so we can continue using dompdf for our invoices !

Do you have by any chance some time to merge this PR ?

Thanks a lot

bsweeney:
Apologies for the delay I'll get back into it so we can release the update.

Issue #3443, opened 2024-04-19
Be Factur-x ready


williamdes:
[What is Factur-x ?](https://github.com/akretion/factur-x?tab=readme-ov-file#factur-x-and-order-x-python-library)

related tools:
- https://github.com/akretion/factur-x
- https://mpdf.github.io/what-else-can-i-do/pdf-a3-xmp-rdf.html
- Seems to have been implemented in https://github.com/mpdf/mpdf/pull/558/
- rdf data in https://github.com/mpdf/mpdf/commit/4978b70eea405aed75633bd05ecbe3eab5b0b947

Related issue on TCPDF: https://github.com/tecnickcom/TCPDF/issues/578 (no progress)

Initial implementation in #3269 for PDF/A

Required tasks:
- [x] https://github.com/dompdf/dompdf/issues/3576

Currently this are the errors I have to solve (https://services.fnfe-mpe.org):
Vera PDF only reports: [Specification: ISO 19005-3:2012, Clause: 6.2.11.4.1, Test number: 1](https://github.com/veraPDF/veraPDF-validation-profiles/wiki/PDFA-Parts-2-and-3-rules#rule-621141-1)
```

Spec. ISO_19005_3 clause 6.2.11.4 test 1
PDF/A-3 error
The font programs for all fonts used for rendering within a conforming file shall be embedded within that file, as defined in ISO 32000-1:2008, 9.9.
Level: CosDocument
Context:
root/document[0]/pages[0](6 0 obj PDPage)/contentStream[0](7 0 obj PDContentStream)/operators[818]/font[0](Helvetica-Bold)
root/document[0]/pages[0](6 0 obj PDPage)/contentStream[0](7 0 obj PDContentStream)/operators[853]/font[0](Helvetica)
Missing /Names/EmbeddedFiles in PDF Catalog
PDF/A-3 error
Required tag 'DocumentType' missing
XMP error
Missing tag /x:xmpmeta/rdf:RDF/rdf:Description/fx:DocumentType in XMP Metadata (can also be set via an attribute 'DocumentType' of the tag '/x:xmpmeta/rdf:RDF/rdf:Description')
Required tag 'DocumentFileName' missing
XMP error
Missing tag /x:xmpmeta/rdf:RDF/rdf:Description/fx:DocumentFileName in XMP Metadata (can also be set via an attribute 'DocumentFileName' of the tag '/x:xmpmeta/rdf:RDF/rdf:Description')
Required tag 'Version' missing
[...]

PR #3593, opened 2025-02-08
Add support for SetAdditionalXmpRdf

williamdes:
Add support for SetAdditionalXmpRdf

Ref: #3443

### DIFF
--- a/lib/Cpdf.php
+++ b/lib/Cpdf.php
@@ -298,6 +298,11 @@ class Cpdf
     public $encryptionKey = '';
 
     /**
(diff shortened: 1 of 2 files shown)


bsweeney (approved):
I was trying to come up with something a little more flexible, but this should be sufficient for an initial implementation.

bsweeney:
I'm not familiar enough with the Factur-X or Zugferd specifications to validate that this change adds sufficient functionality to fully support those specifications, but I was able to validate a document as ZUGFeRD 2.x MINIMUM.

PR #3640, opened 2025-08-03
Support paging of non-flow elements and table cells

bsweeney:
This update modifies Dompdf's paging logic in the following manner:

- The pageable parent of an element can now be a sub-element of the page so that non-flow elements (floats, absolute positioning) can page without impacting the paging of in-flow elements.
- Paging of table cells is now supported, including the tracking of rowspanned cells so that they are not lost when a table is paged.


suhari467:
Hi, i am use newest dompdf version 3.1.4, but some for reason breaking table cells because to many text not working. 
I manually replace some file on this update (brach: paging-updates). after I replace some update & the last file is FrameReflower/Page.php, but has ended:

Error Call to a member function move() on null

FrameDecorator\AbstractFrameDecorator.php:974
final function move(float $offset_x, float $offset_y, bool $ignore_self = false): void
{
   $this->_positioner->move($this, $offset_x, $offset_y, $ignore_self);
}

any solution of that ?
P.s I am using Laravel 8

Issue #3644, opened 2025-08-18
Deprecation in Php8.5


oleibman:
Php8.5, which is now in Beta so it will be official soon enough, deprecates the use of "non-canonical" types for casts. As far as I can tell, you have only one of these - a cast to `double` rather than `float` in:
- src/Css/AttributeTranslator.php on line 506.

oleibman:
I downloaded and ran our test suite using your PR #3654. I am pleased to report that I saw no deprecation messages for Dompdf during that test. We don't use SVG. I hope that you are able to deal with that problem as effectively as those that I reported.

PR #3654, opened 2025-09-01
Improve PHP 8.5 compatibility

bsweeney:
- only use imagedestroy with PHP < 8
- only use curl_close with PHP < 8
- use http_get_last_response_headers with PHP >= 8.4
- use canonical types for casts

addresses #3644


oleibman:
Great job! I have locked all our active branches to Dompdf 3.1.1, and there are no longer any deprecation notices, for Php8.5, and everything continues to work for 8.5 and earlier Php releases. Let's hope for no more surprises.

PR #3662, opened 2025-09-22
Backport some of Debian patches

williamdes:
This backports some of my changes from Debian: https://udd.debian.org/patches.cgi?src=php-dompdf&version=3.1.0%2Bdfsg-1

I can remove the ones you do not like, this is to try to have less patches to rebase on each version

### DIFF
--- a/src/FontMetrics.php
+++ b/src/FontMetrics.php
@@ -113,7 +113,7 @@ class FontMetrics
      */
     public function loadFontFamilies()
     {
-        $file = $this->options->getRootDir() . "/lib/fonts/installed-fonts.dist.json";
+        $file = $this->options->getFontDir() . "/installed-fonts.dist.json";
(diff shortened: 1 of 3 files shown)


src/FontMetrics.php
```diff
@@ -113,7 +113,7 @@ public function load_font_families()
      */
     public function loadFontFamilies()
     {
-        $file = $this->options->getRootDir() . "/lib/fonts/installed-fonts.dist.json";
+        $file = $this->options->getFontDir() . "/installed-fonts.dist.json";
```
bsweeney, line 116: I'd have to think on this one more. The purpose of this particular line of code was to pull the distributed JSON to ensure Dompdf always has a baseline of supported fonts. The generated JSON file can override that, but there's no guarantee that it will include the ".dist" file.

PR #3673, opened 2025-10-19
Set offset when parsing CSS for data URIs

### LINKED ISSUE #3672: Hanging/Infinite loop bug in Stylesheet::_parse_css
See
https://github.com/dompdf/dompdf/blame/master/src/Css/Stylesheet.php#L1306

This appears to have changed around release 3.1.0 => 3.1.1.

If the character before "data:" is not one of the expected chars, the loop will continue forever.

I'm seeing this happen in a css string that contains elements like `&quot;data:`.

bsweeney:
fixes #3672

### DIFF
--- a/src/Css/Stylesheet.php
+++ b/src/Css/Stylesheet.php
@@ -1303,8 +1303,9 @@ class Stylesheet
 EOL;
 
         // replace data URIs with blob URIs
-        while (($start = strpos($css, "data:")) !== false) {
-            $len = null;
+        $offset = 0;
+        while ($offset < strlen($css) && ($start = strpos($css, "data:", $offset)) !== false) {
+            $len = 0;
             $prev = $css[$start - 1];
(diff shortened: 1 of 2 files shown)


bsweeney:
@Cruiser13 that specific sample rendered OK for me on 3.1.4.

This change addresses an issue parsing data URIs that are part of a stylesheet. Which is maybe where you're encountering an issue? For example, the following does not work in 3.1.4:

```
<style>
    .bg { background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg'); }
</style>
<div class="bg">Test</div>
```

I've already produced a fix in relation to #3706 and plan to push a new release in the near future. You can work around that issue by using double quote instead of single quotes

PR #3696, opened 2025-12-02
Optimise split function

### LINKED ISSUE #3695: Optimise split function
1. Reduce number of calls to split function
2. Prevent redundant calls

nehajyoti04:


bsweeney (changes requested):

williamdes:
*(Posting on behalf of the repo owner, who asked me — Claude, an AI coding assistant — to share these measurements.)*

Re @bsweeney's request for *"sample HTML that shows the improvement"* — I benchmarked this PR while investigating #3738 (super-linear table rendering). Here is a sample that stresses the split path, and the numbers.

```php
function tableHtml(int $rows): string {
    $row = '<tr>' . str_repeat('<td>value</td>', 8) . '</tr>';
    return '<style>'
         . 'table { border-collapse: collapse; }'
         . 'tr    { page-break-inside: avoid; }'   // forces a split decision per row
         . 'td    { border: 1px solid #000; padding: 2px; font-size: 9px; }'
         . '</style><table>' . str_repeat($row, $rows) . '</table>';
}
// render tableHtml(100|200|400|800) and time $dompdf->render()
```

`3.1.5` as-is vs `3.1.5` + this PR, 2 runs each, PHP 8.3:

| rows | 3.1.5 (run1 / run2) | + this PR (run1 / run2) |
|-----:|---------------------|-------------------------|
|  100 | 0.78s / 0.73s | 0.70s / 0.75s |
|  200 | 1.40s / 1.58s | 1.28s / 1.44s |
|  400 | 3.33s / 3.71s | 3.21s / 3.39s |
|  800 | 8.70s / 9.03s | 8.14s / 8.54s |

Net: a consistent **~5–7% constant-factor improvement**, and the re-entrancy / "don't re-split an already-split node" guards look like a genuine correctness win. It does not change the scaling on this case — ms/row still grows with row count — because the batched move still `reset()`s the breaking row plus every following sibling and the remainder is re-reflowed (same O(n) work, just collected first). Fuller analysis of that quadratic is in #3738. So this PR is a worthwhile improvement on its own; it just isn't the full fix for #3738.

Issue #3713, opened 2026-02-02
Text wrapping around floating element stops working after some iterations


Fl0Cri:
I am using Dompdf in a background process (Laravel queue) to batch-generate potentially large amouts of documents. To keep memory usage under control, the documents are generated individually as temp files and then merged with Ghostscript afterwards.

This has been working well until I recently introduced a new document with a `float: right` applied to an image, to make text wrap around it.

The first documents generated are fine, but at some point, Dompdf stops reflowing the text. Any new documents generated after this point have this problem. After restarting the background process, some documents are fine before the issue reappears (~80 in my real-world scenario).

The screenshot below shows the exact moment at which the problem occurs: the first 2 lines wrap around the green square correctly, but the subsequent lines don't. All preceding documents are fine (all lines wrap correctly).

<img width="858" height="496" alt="Image" src="https://github.com/user-attachments/assets/7d4b16d8-98ba-4939-b913-ef85bde58709" />

As every document is generated with a new Dompdf instance, I suspected that a static property somewhere was the cause of this. After some quick research, I found the culprit: https://github.com/dompdf/dompdf/blob/77a7a1347d14eed275097b6e084433203857343c/src/LineBox.php#L157

Removing the `static` keyword (which effectively disables this loop prevention) does resolve the issue, but I guess it may introduce infinite loops with certain inputs. I havn't experienced any so far, that being said I am not trying to render webpages or messy user-supplied html.

If a proper fix would be too complicated to implement, maybe moving the static variable to a static class property which could be reset by external code may allow to work around the issue. (it may be reset by DomPDF itself when starting to render)

Here is an example code to demonstrate the issue:
```php
$template = <<<HTML
<html>
    <head>
        <style>
            .green-square {
                float: right;
                width: 5cm;
                margin-left: 8mm;
                margin-bottom: 8mm;
            }
        </style>
    </head>

    <body>
[...]

Issue #620, opened 2013-04-16
Static LineBox offset processing counter breaks layout when it reaches zero


jamieburchell:
Steps to reproduce the issue:

Create a PDF containing around 30 blocks of paragraph text, each with a single image floated right. After several pages the text stops flowing around the image and instead goes behind it.

At first I thought there was a specific limit of floated images before this happens, but I feel it might also be connected with the number of pages as after some more testing the number changes.

I have reproduced the issue on dompdf trunk and dompdf 0.6.0b3 on Windows and Linux.

Am happy to provide a sample HTML/PDF is necessary.

Thanks!

davidsickmiller:
This can be "fixed" by increasing this variable in dompdf/include/line_box.cls.php's get_float_offsets():

```php
static $anti_infinite_loop = 500; // FIXME smelly hack
```

martijngastkemper:
The limit mentioned by @davidsickmiller has been increased in later versions. But there is still a limit. I've "fixed" it by replacing the usage of `float` with `display: inline-block`. I've full control over the HTML being used to create the PDF file so that made this solution possible without breaking stuff.

PR #3714, opened 2026-02-04
Workaround for float reflows limit

Fl0Cri:
As discussed in #3713, related to #620.

Before someone can come up with a real fix (or a proof that this loop-prevention is not needed anymore), this PR resets the static reflows counter on every render.

By the way, users experiencing problems when rendering very large documents can override this limit externally to any absurd value that may suit their needs, for example:
```php
\Dompdf\LineBox::$max_float_reflows = PHP_INT_MAX;
```

### DIFF
--- a/src/Dompdf.php
+++ b/src/Dompdf.php
@@ -772,6 +772,8 @@ class Dompdf
 
         $canvas = $this->canvas;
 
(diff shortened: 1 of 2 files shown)


bsweeney (approved):

bsweeney:
nicely refactored, I appreciate that you made the property public static so users could modify during use and not have to modify the source

PR #3729, opened 2026-05-02
feat: enable compressed response encoding for cURL requests

EgorGruzdev:
## Summary

Enable compressed response encoding for cURL requests by setting `CURLOPT_ENCODING` to an empty string `''`, which sends an `Accept-Encoding` header with all supported encodings (gzip, deflate, br). This improves download performance for remote resources.

## Changes

- Added `curl_setopt($curl, CURLOPT_ENCODING, '')` in `Helpers::fileGetContent` to advertise support for compressed responses

## Files Changed

- `src/Helpers.php` — Modified (1 insertion)

## Testing

Needs testing — verify that remote resource fetching still works correctly with compressed responses enabled.

## Related Issues

None

### DIFF
--- a/src/Helpers.php
+++ b/src/Helpers.php
@@ -1086,6 +1086,7 @@ class Helpers
 
                 curl_setopt($curl, CURLOPT_RETURNTRANSFER, true);
                 curl_setopt($curl, CURLOPT_HEADER, true);
+                curl_setopt($curl, CURLOPT_ENCODING, '');
                 if ($offset > 0) {
                     curl_setopt($curl, CURLOPT_RESUME_FROM, $offset);
                 }


EgorGruzdev:
## Motivation

Dompdf's cURL implementation in `Helpers::getFileContent` did not set `CURLOPT_ENCODING`, meaning no `Accept-Encoding` header was sent with outgoing requests. Per the HTTP specification, servers should not compress responses in this case — and most origin servers behave correctly.

However, some proxies, CDNs, and security services may return compressed responses regardless of whether the client requested compression — both from cache and when proxying requests to the origin. This is non-standard behavior, but it does occur in practice (e.g. DDoS-Guard).

Without `CURLOPT_ENCODING`, cURL does not automatically decompress the response body. As a result, dompdf receives raw compressed binary data instead of CSS/HTML, which causes parse errors.

## Fix

Setting `CURLOPT_ENCODING` to `''` instructs cURL to automatically decompress the response body whenever the server returns a `Content-Encoding` header — regardless of the behavior of any CDN or proxy sitting in front of the origin server.

PR #3730, opened 2026-05-04
Adding enum `PageOrientation`

ThomasLandauer:
What about adding those `const`s instead of using strings everywhere?

A dedicated file `enum Orientation` would be even better, but a BC break ;-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ $dompdf = new Dompdf();
 $dompdf->loadHtml('hello world');
 
 // (Optional) Setup the paper size and orientation
-$dompdf->setPaper('A4', 'landscape');
+$dompdf->setPaper('A4', PageOrientation::Landscape);
 
 // Render the HTML as PDF
 $dompdf->render();
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
         "ext-gd": "*",
         "ext-json": "*",
         "ext-zip": "*",
+        "phpstan/phpstan": "*",
         "phpunit/phpunit": "^7.5 || ^8 || ^9 || ^10 || ^11",
         "squizlabs/php_codesniffer": "^3.5",
         "mockery/mockery": "^1.3",
--- a/src/Adapter/CPDF.php
+++ b/src/Adapter/CPDF.php
@@ -14,6 +14,7 @@ use Dompdf\Exception;
 use Dompdf\FontMetrics;
 use Dompdf\Helpers;
 use Dompdf\Image\Cache;
+use Dompdf\PageOrientation;
 use FontLib\Exception\FontNotFoundException;
 
 /**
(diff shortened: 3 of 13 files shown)


bsweeney:
I generally do prefer constants or enums over "magic strings" so this seems reasonable.

ThomasLandauer:
OK, so pick which you like :-)
1. `const`
2. `enum PageOrientation` (=BC break). One day, this might then even become a native PHP feature, like the new [`SortDirection` enum](https://wiki.php.net/rfc/sort_direction_enum) ;-)
3. Something in between, like:
    ```php
    public function setPaper($size, PageOrientation|string $orientation = PageOrientation::Portrait): self
    ```

I'll then search for more occurrences and replace them all.

bsweeney:
Oh sure, make me decide 😅. I prefer enums, generally, but to avoid breaking changes constants seems to be a good choice. If you want to do the work for the in-between option that'll position us to easily break with the past when we think we're ready.

ThomasLandauer:
I just wanted to go for `PageOrientation|string`, but - oh my gosh - you're still supporting PHP 7.1, don't you? https://github.com/dompdf/dompdf/blob/master/composer.json#L27
Enums were added in PHP 8.1

This PR is probably not worth bumping the required version?!
So let's stick with the `const`s for now?

bsweeney:
Oh jeez ... sorry about that. I was thinking about bumping the version for the 4.0.0 if I don't overlook it.

ThomasLandauer:
And bump it to 8.1 right away? I mean, this enum isn't urgent - so if you'd create a v4 branch, I could commit it there.

bsweeney:
For sure I'm thinking PHP 8.1 would be the minimum. It's already EOL and is a good breakpoint from a functional standpoint.

ThomasLandauer:
OK, then I would say, let's go for the `enum PageOrientation` right away. Will you create a branch for v4 or how shall we proceed?

bsweeney:
I'll go ahead and create a branch for 8.1 and any work specific to that can branch off of it.

ThomasLandauer:
OK, so you'll tell me when the branch is ready?
https://github.com/dompdf/dompdf/pull/3703 could probably also benefit.

bsweeney:
Done and I updated the target branch.

There are quite a few bits of logic that would benefit from the change.

ThomasLandauer:
OK, here we go :-)

* I went for the enum (i.e. BC break) now, for 2 reasons:
  * Better *one* change now for everybody, instead of educating people to move towards a `const` now, only to have another change towards an enum in a few months/years.
  * It shouldn't be too hard for people to upgrade, since the default value didn't change (so many won't even notice it).
* It's a pure enum, as per https://wiki.php.net/rfc/sort_direction_enum
* I named it `PageOrientation`, since this seems to be the more common term than "paper orientation"
* I kept the existing strings in CSS related files, since ultimately, in CSS, they are strings.
* I also added PHPStan in `require-dev` - might be useful for other contributors as well. But I don't need it anymore, so if you want, I can remove it.

In principle, the same could be done for paper sizes, but the array at https://github.com/dompdf/dompdf/blob/master/src/Adapter/CPDF.php#L42 looks a bit scary ;-)

PR #3734, opened 2026-06-02
Update Block.php

mujuonly:
Null guard — skip layout if containing block is unavailable


bsweeney:
Do you have sample HTML where the containing block would be unavailable?

PR #3750, opened 2026-09-02
Fix encrypted embedded file creation date

### LINKED ISSUE #3747: Undefined variable when encrypting an attached file
In https://github.com/dompdf/dompdf/blob/b14267808b811db092f53830f81f4706f4917c79/lib/Cpdf.php#L3106

The variable `$creation` doesn't exist, it's actually `$created`.

asllanmaciel:
Fixes the undefined `$creation` variable used while encrypting embedded-file metadata. The creation timestamp is stored in `$created` and later emitted as `/CreationDate`, so the encryption step now updates that same variable.

Adds a regression test that reproduces the warning on the previous `master` head and verifies that rendering an encrypted embedded file no longer emits `Undefined variable $creation`.

Verification:

- RED reproduced on `b14267808b811db092f53830f81f4706f4917c79`
- focused regression: 1 test / 2 assertions — PASS after fix
- full PHPUnit suite: 1,147 tests / 2,893 assertions — PASS
- PHP lint — PASS
- PHPCS — PASS
- `git diff --check` — PASS

Fixes #3747.


bsweeney (approved):

asllanmaciel:
Thanks again for the review and approval. Just checking whether anything else is needed from me before merge; happy to adjust if needed.

bsweeney:
Nothing more needed. Thanks for the patch. I'll start merging things for the 3.1.7 release in the near future.

